PR #33, opened 2015-04-27
Settings reimplemented

fabiansinz:
* moved DataJointError to __init__
* moved logger to settings
* renamed core into utils
* implemented settings.Config
* instantiate datajoint.config in __init__
* wrote tests for settings

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@
 __*__
 .idea
 datajoint.egg-info/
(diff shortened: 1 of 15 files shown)


eywalker:
New implementation looks amazing. Validators currently do not support nested keys, so we have to either make sure of we modify to make validators supported nested config keys or simply not allow for nested keys.

PR #62, opened 2015-05-10
changed `RelatinalOperand.__iter__` and made `settings.Config` a singleton

fabiansinz:
Change of `RelatinalOperand.__iter__`:

* `RelatinalOperand.__iter__ ` now returns an iterator that yields the single tuples of the table as tuples
* the old behaviour of returning the primary keys can be achieved by `mytable.project()`
* `RelatinalOperand.fetch` now uses `__iter__`

Settings:

* I thought it makes sense that `settings.Config` is a singleton since there should be one set of global settings.


dimitri-yatsenko:
The regular fetch returns numpy records. Why should the iterator yield tuples instead.  Let's make it return numpy arrays with typed attributes.

fabiansinz:
I would have preferred it otherwise, but I don't make the main design decisions. Therefore, I changed it. 

Here is why I would have preferred it otherwise: 

I don't like record arrays. While they are awkward to construct and deal with, I think they provide very little benefit. I thought it would have been the cleaner solution for `__iter__` to return tuples. In particular, since the creating of a single `np.void` is done via

```
np.array([tuple(unpack(field) if up else field for up, field in zip(do_unpack, q))],
                           dtype=self.heading.as_dtype)[0]
```

I couldn't find a way to just create a single line of the above array. This means that the iteration

```
for animal_id, v_trace, comment in Mice():
   ...
```

now constructs a record array from a tuple, takes its first element, returns an np.void, only to be converted into a tuple again. This does not seem efficient/clean to me.

Does `fetch` get any easier? No, since constructing record arrays from existing np.voids is not straightforward. What used to be

```
np.array(list(self.__iter__(offset, limit, order_by, descending)), dtype=self.heading.as_dtype)
```

now became

```
np.atleast_1d(rfn.stack_arrays(list(self.__iter__(offset, limit, order_by, descending)), usemask=False))
```

What do we gain? We can use

```
for row in Mice():
   x = row.animal_id # or row['animal_id']
```

To me that looks just like emulating Matlab while Python can offer cleaner constructs for that. If we want the 'row.animal_id' behaviour, we could have just used `namedtuples`.

eywalker:
Looking at the current implementation, I agree with @fabiansinz that returning record array seems to be making the code rather dirty and convoluted, yet it is unclear what we are gaining from this...

dimitri-yatsenko:
The result of the iteration needs to preserve the attribute names.  So at
the very least it's a dict.

fabiansinz:
What do we need that for? For restrictions?

dimitri-yatsenko:
exactly

PR #91, opened 2015-05-20
bug fix and limit make_tuples attempts.

dimitri-yatsenko:


dimitri-yatsenko:
Is it just me or is this solution overly complex? Lots of things happening behind the scenes across multiple locations in the code?  

I am all for using python's clever features as long as they make things clearer, more efficient and explicit.

I recommend removing the `in_transaction` decorator.  Its function is to replace the line:
```python
if self.conn.in_transaction:
    raise TransactionError("this function cannot be called from within a transaction
```
in two places in the code with something much more elaborate and distributed.  It makes the code harder to read and understand.  It also adds an extra dependency, complicates the call stack, and makes debugging more difficult.  It takes more time to understand and use than the original constructs. 

I am not sure about the `Transaction` class.  It adds some 60 lines of code with no added functionality or clarity.

fabiansinz:
Since this pull request has do be done by hand, I will close it, pull the changes into my local branch, merge everything, and make a new pull request.

PR #94, opened 2015-05-21
Removed context manager for connection transaction and decorator for transaction

fabiansinz:


datajoint/autopopulate.py
```diff
@@ -45,35 +45,48 @@ def populate(self, restriction=None, suppress_errors=False, reserve_jobs=False):
...
+                            logger.info('Transaction error in {0:s}.'.format(tr_err.culprit))
+                    else:
+                        raise DataJointError(
+                            '%s._make_tuples failed after multiple attempts, giving up' % self.__class__)
+                except Exception as error:
+                    self.conn.cancel_transaction()
+                    if not suppress_errors:
+                        raise
+                    else:
+                        print(error)
```
dimitri-yatsenko: This should be a `logger.error()` statement so that user has control over display.

datajoint/autopopulate.py
```diff
@@ -45,35 +45,48 @@ def populate(self, restriction=None, suppress_errors=False, reserve_jobs=False):
...
+                        try:
+                            self._make_tuples(dict(key))
+                            break
+                        except TransactionError as tr_err:
+                            self.conn.cancel_transaction()
+                            tr_err.resolve()
+                            self.conn.start_transaction()
+                            logger.info('Transaction error in {0:s}.'.format(tr_err.culprit))
+                    else:
+                        raise DataJointError(
```
dimitri-yatsenko: The attempts are not always multiple, so I would just say `%s._make_tuples transaction failed.`

PR #104, opened 2015-06-15
Started to reimplement tests

fabiansinz:


tests/__init__.py
```diff
@@ -33,7 +33,18 @@ def setup():
...
+    cur = BASE_CONN.cursor()
+    # cancel any unfinished transactions
+    cur.execute("ROLLBACK")
+    # start a transaction now
+    cur.execute("START TRANSACTION WITH CONSISTENT SNAPSHOT")
+    cur.execute("SHOW DATABASES LIKE '{}\_%'".format(PREFIX))
+    dbs = [x[0] for x in cur.fetchall()]
+    cur.execute('SET FOREIGN_KEY_CHECKS=0') # unset foreign key check while deleting
+    for db in dbs:
+        cur.execute('DROP DATABASE `{}`'.format(db))
```
dimitri-yatsenko: DROP DATABASE causes and implicit commit.  So the transaction is of no use.
https://dev.mysql.com/doc/refman/5.6/en/implicit-commit.html

Issue #28, opened 2015-04-24
Increase test coverage >90%


fabiansinz:
At the moment we are at 47%, which is clearly unacceptable.

dimitri-yatsenko:
It's a worthy goal to strive for but let's keep the issues relevant to downstream users.

fabiansinz:
I think it is important to have to well tested. At least leave the issue open.

eywalker:
With whole bunch of recent changes both in the main code and in the tests, our coverage is now down to 63%. We should aim to increase this to at least ~70% prior to our first release.

fabiansinz:
I agree. Just as an additional reminder, please use a short docstring for each test function as this will show up in travis and makes it a lot easier to infer what the function is testing.

dimitri-yatsenko:
Attained 90% coverage in fa3e26edf.  We will now only accept pull requests that do not decrease coverage.

PR #107, opened 2015-06-20
I think we are probably good to merge into the master branch

fabiansinz:


fabiansinz:
I just realized that the compression of the code massively increased the coverage. Currently, we are at 88%. Keep writing test! #28 we can do it!

PR #124, opened 2015-07-16
getitem, declare, license

fabiansinz:
- implemented a __getitem__ method in RelationalOperand that takes: ints, strings, dj.key, and (mixed) slices of ints and strings. dj.key returns the key. 

- implemented a test for the new `__getitem__`

- inserted "if not exists" in the declaration of a new database to avoid conflicts if database creation happens in a distributed call

- changed the license back from MIT to LGPL

### DIFF
--- a/datajoint/relational_operand.py
+++ b/datajoint/relational_operand.py
@@ -7,7 +7,9 @@ import abc
 import re
 from collections import OrderedDict
 from copy import copy
-from datajoint import DataJointError, config
+from . import config
+from . import key as PRIMARY_KEY
+from . import DataJointError
 import logging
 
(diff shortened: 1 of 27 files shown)


datajoint/relational_operand.py
```diff
@@ -310,6 +311,41 @@ def make_condition(arg):
 
         return ' WHERE ' + ' AND '.join(condition_string)
 
+    def __getitem__(self, item): # TODO: implement dj.key and primary key return
+
+        attr_keys = list(self.heading.attributes.keys())
```
dimitri-yatsenko: `attr_keys = self.heading.names`

PR #134, opened 2015-07-23
Fetch object

fabiansinz:
- the behavior of `relation.fetch()` and `relation.fetch1()` should be reproduced
- `relation.keys()` moved to `relation.fetch.keys()`
- `relation.values()` was removed and is now `relation.fetch.__iter__()`

It probably needs more thorough testing, but I put it up so we can start using it.


eywalker:
Implementation looks awesome! Coveralls is not working as they are under maintenance. We'll have to make sure that we document that

1. `relation.fetch` returns a **new** fetch object, but
2. method calls like `order_by` or `limit_by` changes the fetch object **in place**

Reviewed all code and merging it.

PR #135, opened 2015-07-24
Extended order_by and bugfixes

fabiansinz:
- `relation.fetch.order_by` allows for individual sorting arguments now. I removed the kwarg `descending`. The new syntax is `relation.fetch.order_by('name DESC', 'trial', 'stimulus DESC')`. If the sorting argument is omitted, ASC is set by default. 

- I removed the `relation.fetch.apply` function. It is not needed at the moment and simplified the syntax substantially. Additionally, it might cause confusion, since something like `relation.fetch.apply(myfunc).order_by('stimulus')['name']` suggests that the function is applied in the beginning, but it would have been applied at the end. 

- I renamed `FetchQuery` to `Fetch` and `Fetch1Query` to `Fetch1`

- I renamed `limit_by` to `limit_to`. Dimitri actually prefers `limit`, but I think `limit_to` is more consistent with `order_by`. 

- Wrote quite a few more tests.

### DIFF
--- a/datajoint/fetch.py
+++ b/datajoint/fetch.py
@@ -1,8 +1,11 @@
 from collections import OrderedDict
+import itertools
 from .blob import unpack
 import numpy as np
 from datajoint import DataJointError
 from . import key as PRIMARY_KEY
+from collections import abc
+
 
 def prepare_attributes(relation, item):
     if isinstance(item, str) or item is PRIMARY_KEY:
@@ -20,43 +23,35 @@ def prepare_attributes(relation, item):
         raise DataJointError("Index must be a slice, a tuple, a list, a string.")
     return item, attributes
 
-class FetchQuery:
 
+class Fetch:
     def __init__(self, relation):
         """
 
         """
+
         self.behavior = dict(
-            offset=0, limit=None, order_by=None, descending=False, as_dict=False, map=None
+            offset=0, limit=None, order_by=None, as_dict=False
         )
         self._relation = relation
-
 
     def from_to(self, fro, to):
         self.behavior['offset'] = fro
         self.behavior['limit'] = to - fro
         return self
 
-    def order_by(self, order_by):
-        self.behavior['order_by'] = order_by
+    def order_by(self, *args):
+
+        if len(args) > 0:
+            self.behavior['order_by'] = self.behavior['order_by'] if self.behavior['order_by'] is not None else []
+            self.behavior['order_by'].extend(args)
         return self
 
     def as_dict(self):
         self.behavior['as_dict'] = True
 
-    def ascending(self):
-        self.behavior['descending'] = False
-        return self
 
-    def descending(self):
-        self.behavior['descending'] = True
-        return self
-
-    def apply(self, f):
-        self.behavior['map'] = f
-        return self
-
-    def limit_by(self, limit):
+    def limit_to(self, limit):
         self.behavior['limit'] = limit
         return self
 
@@ -78,9 +73,7 @@ class FetchQuery:
         """
         behavior = dict(self.behavior, **kwargs)
 
-        cur = self._relation.cursor(offset=behavior['offset'], limit=behavior['limit'],
-                                    order_by=behavior['order_by'], descending=behavior['descending'],
-                                    as_dict=behavior['as_dict'])
+        cur = self._relation.cursor(**behavior)
 
         heading = self._relation.heading
         if behavior['as_dict']:
@@ -92,22 +85,15 @@ class FetchQuery:
             for blob_name in heading.blobs:
                 ret[blob_name] = list(map(unpack, ret[blob_name]))
 
-        if behavior['map'] is not None:
-            f = behavior['map']
-            for i in range(len(ret)):
-                ret[i] = f(ret[i])
-
         return ret
 
     def __iter__(self):
         """
         Iterator that returns the contents of the database.
         """
-        behavior = self.behavior
+        behavior = dict(self.behavior)
 
-        cur = self._relation.cursor(offset=behavior['offset'], limit=behavior['limit'],
-                                    order_by=behavior['order_by'], descending=behavior['descending'],
-                                    as_dict=behavior['as_dict'])
+        cur = self._relation.cursor(**behavior)
 
         heading = self._relation.heading
         do_unpack = tuple(h in heading.blobs for h in heading.names)
@@ -126,10 +112,10 @@ class FetchQuery:
         """
         Iterator that returns primary keys.
         """
+        b = dict(self.behavior, **kwargs)
         if 'as_dict' not in kwargs:
-            kwargs['as_dict'] = True
-        yield from self._relation.project().fetch.set_behavior(**kwargs)
-
+            b['as_dict'] = True
+        yield from self._relation.project().fetch.set_behavior(**b)
 
     def __getitem__(self, item):
         """
@@ -146,7 +132,7 @@ class FetchQuery:
         single_output = isinstance(item, str) or item is PRIMARY_KEY or isinstance(item, int)
         item, attributes = prepare_attributes(self._relation, item)
 
-        result = self._relation.project(*attributes).fetch()
+        result = self._relation.project(*attributes).fetch(**self.behavior)
         return_values = [
             np.ndarray(result.shape,
                        np.dtype({name: result.dtype.fields[name] for name in self._relation.primary_key}),
@@ -158,8 +144,7 @@ class FetchQuery:
         return return_values[0] if single_output else return_values
 
 
-class Fetch1Query:
-
+class Fetch1:
     def __init__(self, relation):
         self._relation = relation
 
--- a/datajoint/fetch.py
+++ b/datajoint/fetch.py
@@ -1,5 +1,7 @@
 from collections import OrderedDict
+from functools import wraps
 import itertools
+import re
 from .blob import unpack
 import numpy as np
 from datajoint import DataJointError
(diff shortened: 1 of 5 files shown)


dimitri-yatsenko:
note that order_by can operate on expressions, not just attribute names.  For example, if I wish to sort so that name='Fabian' comes before all others I would need to say, according to the proposed syntax:
```python
relation.fetch.order_by('name="Fabian"=DESC', 'trial')
```
This does not seem like good syntax. Plus the word DESC is hardly PEP8-friendly.

How about tuples `('expression', True)` for descending. 
```python
relation.fetch.order_by(('name', True), 'trial', ('stimulus', True))
```

datajoint/fetch.py
```diff
@@ -20,46 +25,57 @@ def prepare_attributes(relation, item):
         raise DataJointError("Index must be a slice, a tuple, a list, a string.")
     return item, attributes
 
-class FetchQuery:
+def copy_first(f):
```
dimitri-yatsenko: What's the reason for `copy_first`?

tests/schema.py
```diff
@@ -39,6 +39,26 @@ class Subject(dj.Manual):
...
 
+@schema
+class Language(dj.Lookup):
+
+    definition = """
+    # languages spoken by some of the developers
+
+    entry_id    : int
+    ---
+    name        : varchar(40) # name of the developer
```
dimitri-yatsenko: Sorry, this table is not in 3rd normal form.   I know it's just a test example but we might as well show good examples.

datajoint/fetch.py
```diff
@@ -20,46 +25,57 @@ def prepare_attributes(relation, item):
...
+            self._relation = relation._relation
+        else:
+            self.behavior = dict(
+                offset=0, limit=None, order_by=None, as_dict=False
+            )
+            self._relation = relation
 
 
+    @copy_first
     def from_to(self, fro, to):
```
dimitri-yatsenko: is the pair of methods (`from_to`, `limit_to`) preferred to the pair (`limit', 'offset') that is already familiar to SQL programmers?
```python
relation.fetch.offset(300).limit(100)()
relation.fetch.from_to(300,400)()
```

datajoint/fetch.py
```diff
@@ -20,46 +25,57 @@ def prepare_attributes(relation, item):
...
-        self.behavior['descending'] = True
-        return self
-
-    def apply(self, f):
-        self.behavior['map'] = f
-        return self
 
-    def limit_by(self, limit):
+    @copy_first
+    def limit_to(self, limit):
```
dimitri-yatsenko: How about `limit` and `order` instead of `order_by` and `limit_to`?  It seems just as clear but shorter.

datajoint/fetch.py
```diff
@@ -20,46 +25,57 @@ def prepare_attributes(relation, item):
...
         self.behavior['limit'] = limit
```
dimitri-yatsenko: what's the reason for grouping the settings in `behavior` rather than making them properties of `Fetch`?

eywalker: I believe this is to make copying of `Fetch` object easier

datajoint/fetch.py
eywalker: We thought that it would make more sense for each operation on the `fetch` object to return a new `fetch` object rather than modifying the object in place. Since `fetch` object in itself is very cheap, this shouldn't be a real over head. `copy_first` is a decorator that modifies the decorated method to work on and modify the copy of the object, supporting this idea.

dimitri-yatsenko: why not use `copy.copy`?

datajoint/fetch.py
```diff
@@ -20,46 +25,57 @@ def prepare_attributes(relation, item):
...
+        else:
+            self.behavior = dict(
+                offset=0, limit=None, order_by=None, as_dict=False
+            )
+            self._relation = relation
 
 
+    @copy_first
     def from_to(self, fro, to):
         self.behavior['offset'] = fro
```
dimitri-yatsenko: so `from_to(1,2)` will yield 1 tuple?  That's not the usual understanding of what 100 to 100 means.  I think we should stick with `offset` and `limit` in line with SQL's usage.

fabiansinz: That's in line with the index convention of python and I think it makes a lot of sense. If you want to split an interval a,b at c you can say from_to(a,c) and from_to(c,b) instead of from_to(a,c) and from_to(c+1,b).

datajoint/fetch.py
fabiansinz: And it's easier to insert them into the cursor as kwargs. It distinguishes the fetch behavior from other properties of the object, like `_relation`.

datajoint/fetch.py
fabiansinz: If you use `order_by` and `limit_to` a fetch statement almost reads like a sentence: `rel.fetch.order_by('stimulus').limit_to(10)`. I like that. It's clear.

datajoint/fetch.py
fabiansinz: We could introduce another method called  `offset_by`

tests/schema.py
fabiansinz: I can change it later.

datajoint/fetch.py
fabiansinz: `copy.copy` does not copy the dict `behavior`. The function `deepcopy` would copy the relation as well. That's why I introduced a copy constructor and used a decorator to tell the function, that I'd like `self` copied. Having the elements of `behavior` as object attributes wouldn't solve the problem, since some of the values of `behavior` are lists.

datajoint/fetch.py
dimitri-yatsenko: Python index convention does not use the language 'from' and 'to', which implies including the first and last values.  I really think we should revert to SQL's `offset` and `limit`.  We don't need to reinvent the wheel.  These are not used too often.

fabiansinz: I am fine with offset_by  and limit_to.


Am 25. Juli 2015 00:28:19 schrieb Dimitri Yatsenko <notifications@github.com>:

datajoint/fetch.py
dimitri-yatsenko: But why do we need to copy the object in the first place?

datajoint/fetch.py
dimitri-yatsenko: `offset` and `limit` -- cleaner

datajoint/fetch.py
```diff
@@ -20,46 +25,57 @@ def prepare_attributes(relation, item):
...
         return self
 
-    def order_by(self, order_by):
-        self.behavior['order_by'] = order_by
+    @copy_first
+    def order_by(self, *args):
+        if len(args) > 0:
+            self.behavior['order_by'] = self.behavior['order_by'] if self.behavior['order_by'] is not None else []
+            namepat = re.compile(r"\s*(?P<name>\w+).*")
+            for a in args: # remove duplicates
```
dimitri-yatsenko: why do we need to remove duplicates?

datajoint/fetch.py
dimitri-yatsenko: ok, i have looked into other modules and, yes, things like these usually copy the object.  I am merging this PR.

datajoint/fetch.py
fabiansinz: Users could do something like this:
```
dog = relation.fetch.order_by('stimulus ASC', 'user_id DESC', 'trial ASC').limit_to(1000)

data1 = dog.order_by('user_id DESC')()
data2 = dog.order_by('trial ASC')()
```

`dog` would be the default, and only one sorting parameter is changed at a time. To have that-in my opinion nice-feature we need (i) copies of the object and (ii) the ability to overwrite existing settings. According to @eywalker this is consistent with SQLAlchemy ( @eywalker correct me if I am wrong).

datajoint/fetch.py
fabiansinz: Either `order`, `limit`, and `offset` or `order_by`, `limit_to`, `offset_by`. Since we are not MySQL there is no need to adopt inconsistent naming conventions. I like the second option better because statements read clearer (like `populated_from`). We could make a poll among existing datajoint-python users which statement they prefer.

eywalker: FYI, SQL alchemy also uses the method name `order_by`, although they use the method name `limit` (not `limit_to`). Following Python convention, I like it when the method calls reads like a sentence. I do really like `order_by` as it strongly suggests that we need to pass in arguments to make it work, and that `order_by()` without argument doesn't really make sense. I somehow feel less strong about `limit` by I could see same argument goes for `limit_to`.

datajoint/fetch.py
eywalker: Each method call creating separate `fetch` object is consistent with how SQLAlchemy handles their `Query` objects. I actually just checked SQLAlchemy and the query it generates, and it looks like it does not go about removing duplicates - it simply appends all statements into SQL. As for SQL, it's completely fine executing queries like `ORDER_BY name DESC, language ASC, name ASC`. As it sorts from last to first, this will cause the newer `sort_by('name ASC')` call to be effecitvely ignored.

One thing that SQLAlchemy supports that may be nice is to call `order_by(None)`. This returns a new query object with all `order_by` reset. Call to `order_by` without an argument is simply ignored in SQLAlchemy

datajoint/fetch.py
eywalker: There is no standalone `OFFSET` command in SQL - it always has to occur with `LIMIT` statement. I suppose we can define `offset_by(N)` to translate into `LIMIT X OFF N`, where X is some large number by default to ensure all rows from *offset* to the end are included in the query.

datajoint/fetch.py
fabiansinz: If SQL sorts from last to first, this means that the first attribute is the major sort key and the others are subordinate keys. The question is, what do we expect from a call `rel.fetch.order_by('attr1').order_by('attr2')`? Is attr1 the major sort key or the subordinate? 

- If attr1 is subordinate, we can just insert new order keys in front of `behavior['order_by']` and don't even need to remove duplicates. Removing them would probably be a bit more efficient on MySQL side, since I don't expect that MySQL checks whether the keys are double. 

- If attr1 is supposed to be the major sort key, then we need to agree on whether a newly added attribute with the same name should cancel out the old one or not. I think I am actually slightly in favour of not removing it and implementing `rel.fetch.order_by(None)` to delete the current order_by chain.

eywalker: I would think that it would actually make sense to make later keys the subordinate ones, in line with how MySQL handles `ORDER BY`. In this case, newly added attribute with same name already existing should *not* cancel out the older one. Rather, I agree that we should just implment `order_by(None)` to support resetting, so the user can construct ordering as desired.

datajoint/fetch.py
fabiansinz: What I don't like is that `limit` and `offset` sound like properties.

datajoint/fetch.py
dimitri-yatsenko: Let's not get carried away and overcomplicate things.  Let's do this:  `order_by` takes the full ordering list and overrides any previous `order_by` settings:
```
a = relation.fetch.order_by('field1 desc', 'field2')()
```

datajoint/fetch.py
dimitri-yatsenko: I do not think that making code read like a sentence is an advantage.  Expressiveness and brevity are better than sounding like English sentences.

fabiansinz: It's clear that we have different opinions on this. Let's have a vote and stick with what the majority prefers. I suggest that the two options are 1. `limit`, `order`, `offset` and 2. `limit_to`, `order_by`, `offset_by`. I am happy to setup a doodle and send it around.

datajoint/fetch.py
fabiansinz: That's the reason why I put in the `from_to`, because it sets both. In the current version you cannot set `offset` without setting `limit` as well.

PR #142, opened 2015-08-14
optimized relational algebra

dimitri-yatsenko:
- optimized relational algebra operators to generate faster, simpler queries.
- removed in-place restriction
- bugfix in cascading delete
- added tests for relational operators
- added datajoint.kill()


dimitri-yatsenko:
how can I see code not covered by tests?

eywalker:
You can look at the report generated by the coveralls.
https://coveralls.io/github/datajoint/datajoint-python

datajoint/connection.py
```diff
@@ -53,14 +53,13 @@ class Connection:
...
     def __init__(self, host, user, passwd, init_fun=None):
-        self.erm = ERM(self)
         if ':' in host:
             host, port = host.split(':')
             port = int(port)
         else:
             port = config['database.port']
         self.conn_info = dict(host=host, port=port, user=user, passwd=passwd)
-        self._conn = pymysql.connect(init_command=init_fun, **self.conn_info)
+        self._conn = connector.connect(**self.conn_info)
```
fabiansinz: Why is `init_fun` ignored now?

dimitri-yatsenko: oh, this was because I tried another connector instead of `pymysql` and
only `pymysql` has the `init_command` option.  I will undo this. I will
redesign this part if we switch to other connectors.

PR #161, opened 2015-09-13
Optimized SQL queries

dimitri-yatsenko:
Reduced the number of subqueries made in complex expressions.


eywalker:
Checked through the code and looks good - will merge upon completion of coveralls execution.

eywalker:
For some reason coveralls is stuck - looks like a glitch on their system. Will merge as travis test is passing.

Issue #15, opened 2014-08-28
cascading delete


dimitri-yatsenko:
see `dj.Relvar/del`  on the MATLAB side.

eywalker:
First pass version of cascading delete is now implemented. As far as I could test, it works appropriately. However, it is crucial that we add more tests for complex relational operation. In particular, we would need to add test cases for `delete` and combination of `delete`, `restrict`, and `project`.

I'll try adding test cases covering cascading as well as normal deletes hopefully within today.

eywalker:
There is a known bug in the cascading delete where restricted deletion on tables with multiple dependency to the source table may be incorrect. Fix will be implemented soon...

dimitri-yatsenko:
mostly solved in e0aadc03. 

There is one non-critical case when delete fails to cascade correctly, but this case has never happened in real schemas.  This case only concerns multiple dependence paths through non-primary dependencies.

eywalker:
Yes, but the solution is known and thus there is no reason not to implement
the fix :)

Issue #158, opened 2015-09-09
restrictions by a list


dimitri-yatsenko:
Recently we introduced a discrepancy of what it means to restrict by a list of objects.  

Initially, restricting by a list combined the conditions using OR.  Recently we overrode this feature and combined conditions using AND. 

I believe, that restricting by a list or an array of conditions should be an OR operation.  This is how the MATLAB version handles it also.  With this convention, a collection of restrictions can be thought of as another relation and semijoins and restrictions become equivalent:

For example, the following two restrictions should produce the same result:
```python
A() & B()
A() & B().fetch()
```

At the same time, we need to be able to combine a list of restrictions using AND in some cases, internally. This is necessary when apply the list of restrictions from one relation to another relation, for example.

dimitri-yatsenko:
This also means that
```python
A() & []
```
will return an empty relation.
Currently it returns `A()`.

dimitri-yatsenko:
Actually, I am not sure what `A() & []` should mean. I think my previous comment was wrong.  It should return `A()` since every condition in `[]` is met.

fabiansinz:
If it is OR, then it should return `A()`.

dimitri-yatsenko:
What about `A() - []` then?  It seems that it should be ```A()``` too.  I sense a contradiction.

But this would be inconsistent with `semijoin`:  `rel & empty_relation` is empty if `empty_relation` has common attributes with `rel`.

dimitri-yatsenko:
The definition of restriction `rel & conditions` is: "All tuples in `rel` that meet any of the conditions in `conditions`".

With this definition, `rel & []` is empty and `rel - []` is `rel`.

fabiansinz:
I don't think that `A() - []` should be `A()`. By the above logic it should be empty. However

- I agree that `A() - []` and `A() & []` should yield opposite results. 
- I think that `A() & []` being `A()` is intuitive, but `A() - []` being empty is not. 
- As I pair I think `A() & []` being empty and `A() - []` being `A()` is more intuitive than the opposite combination. 

Therefore, I agree with you.

dimitri-yatsenko:
Conclusions
---
After off-line discussion with @fabiansinz:

`rel & [cond1, ..., condn]` is defined as "a relation comprising all tuples from `rel` that match *any* of the conditions in the list."

Then the following will always result in an empty relation:
```python
rel & []
rel & ()
rel & set()
```

Then the negative restriction will return the original relation:
```python
[...]

PR #162, opened 2015-09-14
Fixed issue #15

dimitri-yatsenko:
Converging dependencies are now handled correctly in deletes.
Also, `rel & [cond1, cond2, ...]` now works for any type of conditions, including relations and strings. The conditioned are ORed in this case.

### DIFF
--- a/datajoint/relation.py
+++ b/datajoint/relation.py
@@ -1,5 +1,5 @@
 from collections.abc import Mapping
-from collections import OrderedDict
+from collections import OrderedDict, defaultdict
 import numpy as np
 import logging
 import abc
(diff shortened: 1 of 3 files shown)


datajoint/relation.py
```diff
@@ -257,24 +257,33 @@ def delete(self):
...
-        for r in relations:
-            restrict_by_me.update(r.references)
-        relations = OrderedDict((r.full_table_name, r) for r in relations)
 
+        # construct a list (OrderedDict) of relations to delete
+        relations = OrderedDict((r.full_table_name, r) for r in self.descendants)
+
+        # construct restrictions for each relation
+        restrict_by_me = set()
+        restrictions = defaultdict(lambda: list())
```
fabiansinz: That should be `defaultdict(list)`

datajoint/relation.py
```diff
@@ -257,24 +257,33 @@ def delete(self):
...
-                    *([r.project()] if name in restrict_by_me else r.restrictions))
-
+                if name in restrict_by_me:
+                    restrictions[dep].append(r)
+                else:
+                    restrictions[dep].extend(restrictions[name])
+
+        # apply restrictions
+        for name, r in relations.items():
+            if restrictions[name]:
```
fabiansinz: Wouldn't `name in restrictions` be the idiom?

datajoint/relation.py
dimitri-yatsenko: you are right.  Will correct.

datajoint/relation.py
dimitri-yatsenko: No, because even in a name is in restrictions, it may contain an empty list.  Empty lists should not be applied and should be treated as no restriction.  According to issue #158, restricting by an empty list results in an empty relation rather than in an unrestricted relation.

dimitri-yatsenko: I will add a comment to clarify.

PR #166, opened 2015-10-14
Fetch1 checks object length, attribute ordering bug removed

fabiansinz:


datajoint/fetch.py
```diff
@@ -279,7 +279,10 @@ def __getitem__(self, item):
         single_output = isinstance(item, str) or item is PRIMARY_KEY or isinstance(item, int)
         item, attributes = prepare_attributes(self._relation, item)
 
-        result = self._relation.project(*attributes).fetch()
+        result = self._relation.project(*attributes)
+        assert len(result) == 1, 'Fetch1 should only return one tuple'
```
dimitri-yatsenko: len() is a separate query and can be an expensive operator.  Sometimes len() takes as long as the fetch query.  It will be more efficient to fetch and to check that the result only has one element.

dimitri-yatsenko: Rather than `assert`, use `raise` to provide a more manageable error object.

eywalker:
Changes looks good.

PR #167, opened 2015-10-27
add _repr_html_ for ipython notebook output

fabiansinz:
ipython notebook uses `_repr_html_` to output objects. For example

```
from schema import myRelation
myRelation()
```

will now output a nice html version of the relation. Just as in pandas.


FloFra:
Super Idee, vielen Dank.

fabiansinz:
Kein Problem.

Issue #172, opened 2015-11-27
FUNCTION datetime.date does not exist


tstadler:
I get the following error message when I call populate() for a computed table where the ancestor table has a primary key of type date:

InternalError: (1630, "FUNCTION datetime.date does not exist. Check the 'Function Name Parsing and Resolution' section in the Reference Manual")

I could replicate it with a minimal example and have attached the schema and how I run it. I use an IPython Notebook for testing out schemas.

I could remove the error by giving exp_date the type varchar(20)

Example:
```python
# The minimal example that creates the error
import datajoint as dj

schema = dj.schema('sandbox',locals())

@schema
class Basic(dj.Manual):
    definition="""
    exp_date    :date   # primary key with type date
    ---
    path        :varchar(20)    # variable used in Dependent
    """

@schema
class Dependent(dj.Computed):
    definition="""
    -> Basic
    ---
    new_path    :varchar(21)    # variable computed
    """

    def _make_tuples(self,key):
        p = (Basic() & key).fetch1['path']

        new_path = p + 'new'

        self.insert1(dict(key,new_path=new_path))
```

Run it:
```python
# And how I run the file in an IPython Notebook
import datajoint as dj
import os

c = dj.conn()

%run sandbox.py

from sandbox import Basic
basic = Basic()
comp = Dependent()

basic.insert1({'exp_date':'2015-11-24','path':'example/path'})

comp.populate()

# For me it yields the following error:
# InternalError: (1630, "FUNCTION datetime.date does not exist. Check the 'Function Name Parsing and Resolution' section in the Reference Manual")
```

tstadler:
Sry there was a typo in the schema!
[sandbox.txt](https://github.com/datajoint/datajoint-python/files/45928/sandbox.txt)

fabiansinz:
I updated your comment so that we can see the code.

fabiansinz:
I can reproduce the error and identified the cause. We'll discuss how to properly solve it and fix it.

tstadler:
Great! Thanks

Theresa Stadler
Niethammerstrasse 3
72076 Tübingen
Theresa.Stadler@student.uni-tuebingen.de
Am 27.11.2015 6:34 nachm. schrieb "Fabian Sinz" <notifications@github.com>:

fabiansinz:
Should be fixed by pull request #173 once it is accepted.

PR #173, opened 2015-11-27
Fix datetime bug #172

fabiansinz:


eywalker:
Nice - thanks for including tests @fabiansinz!

Issue #178, opened 2015-12-16
Fix the display of part relations in the ERD


dimitri-yatsenko:
Currently, part relations are missing their master's name.

For example, if `probe.Simulation.Detection` is a part relation of `probe.Simulation`, the current display is 
```
probe.Simulation-------->probe.Detection
```
but it should be
```
probe.Simulation-------->probe.Simulation.Detection
```

PR #180, opened 2015-12-20
Development

dimitri-yatsenko:
Fixed issue #178


datajoint/schema.py
```diff
@@ -87,15 +87,18 @@ def process_relation_class(relation_class, context):
...
+        def is_part(x):
+            return inspect.isclass(x) and issubclass(x, Part)
 
-        for var, part in inspect.getmembers(cls, is_part):
-            parts.append(part)
-            part._master = cls
-            # TODO: look into local namespace for the subclasses
-            process_relation_class(part, context=dict(self.context, **{cls.__name__: cls}))
+        parts = list()
+        for part in dir(cls):
```
eywalker: Why add more lines of code unncessarily? Since we are using inspect, we should just use inspect.getmembers rather than manually traversing dir and applying if

dimitri-yatsenko: Please correct me if I am wrong, but I think `inspect.getmembers` actually evaluates every member.  So if there is a computed property, e.g. `_populated_from`, it will actually trigger its computation.  In contrast, the proposed implementation looks at the names first and only evaluates members if they start with a capital letter.  So it avoids extra computations.

datajoint/user_relations.py
```diff
@@ -2,59 +2,83 @@
...
 
-import abc
 from .base_relation import BaseRelation
 from .autopopulate import AutoPopulate
 from .utils import from_camel_case
 from . import DataJointError
 
 
-class Part(BaseRelation, metaclass=abc.ABCMeta):
+class classproperty:
```
eywalker: Why are we reintroducing `classproperty`? You have ditched this idea in the past since the concept of `classproperty` is not a native one to Python.

datajoint/schema.py
eywalker: I see your concern. As long as you are invoking `inspect.getmembers` on a class and that there are no members that implement `__get__`, no additional computation is performed. However, I see that you have reintroduced `classproperty` concept, and if such thing is present, indeed `inspect.getmembers` will evaluate the *class* property.

dimitri-yatsenko: I thought `property` implemented `__get__`.  No?

dimitri-yatsenko: The default `AutoPopulate._populated_from` is a property that loads dependencies, looks up parents, and returns the join of the parents.

datajoint/user_relations.py
dimitri-yatsenko: Yes, but I think it is just last time the case for using them was not as compelling.  Let's review that again.  Frankly, I did not understand classproperties very well last time and it seemed like then we found an equally good solution without them.

In this case, introducing class properties substantially simplified the code in `erd` for looking up the `master` class.

datajoint/schema.py
eywalker: The `property` and thus `_populated_from` doesn't run the function unless it's run on an instance. That's why I was saying that as long as you invoke `inspect.getmembers` on a class (which is what we'd do anyway), the content of the `_populated_from` will not be evaluated.

datajoint/user_relations.py
eywalker: Last time I brought up the case for stuff like `table_name` as I felt many of these values do not depend on the instance but still needs to be computed. We avoided using `classproperty` ultimately and ended up converting all of such cases into instance level property, requiring us to instantiate the class to access some of these properties. This probably means that we can rewrite the `schema` to take advantage of class property.

Anyway, I'm just giving you a bit of hard time, and I agree that use of `classproperty` is nice (after all I pushed for it last time). I just felt compelled to comment on this :P

dimitri-yatsenko: What I had not realized in my ignorance was that to downstream users, class properties can be used the same way as instance properties.  At the time, I was under the impression that class properties always had to be called on the class and could not be called on the instance.  I also did not realize that when subclassing, each subclass gets its own class property and not just one shared across all subclasses.  These features, combined, make it acceptable to use classproperties.  Sorry I did not know this.  I thought classproperty would introduce all sorts of complications but after checking everything, it seems safe.

dimitri-yatsenko: Yes, now I realize that my arguments agains classmethods  and classproperties were invalid.  Thankfully, changing things to classmethods now does not break any user code.

dimitri-yatsenko: Why doesn't Python3 have a standard `classproperty`?

eywalker: Yeah I have no idea why there is no `classproperty`... Anyway, I'm glad to reintroduce `classproperty` for our use. Given that it's not standard, I still think it'd make sense to keep its use only within DataJoint internal, and not directly at user interfaces.

dimitri-yatsenko: perhaps it is because there might be a confusion with properties of the metaclass.  I would like to find a discussion explaining this.

datajoint/schema.py
dimitri-yatsenko: Then I don't feel too strongly about it.  It seems that `inspect` is designed for inspecting objects during debugging with no regard for performance.  `inspect`.`getmembers` seems to be doing a lot more work than necessary.

PR #181, opened 2015-12-28
Fixed issue #178

dimitri-yatsenko:


eywalker:
What are we trying to do here? If we want `development` branch to be updated with the `master` branch, we should do this from within this repository, not as a pull request from a fork.

dimitri-yatsenko:
@eyewalker, ok. Let's merge `master` into `development` before accepting this pull request.

fabiansinz:
Wouldn't the following solution be cleaner:
* delete the old development branch since it is ancient and has been merged into master already
* rebase your development branch to head of master
* create a new development branch on central repository from that development branch

As a side note: I don't understand why the changes in your development branch should be "development". What is the reason that you don't want to merge the `classproperty` changes into the master?

dimitri-yatsenko:
okay, I have just updated `development`.

I kept it these changes in `development` for now  in case we want to make further changes related to changing some properties to classproperties and methods to classmethods and then merge them all together when it settles.

datajoint/schema.py
```diff
@@ -87,15 +87,18 @@ def process_relation_class(relation_class, context):
...
+            return inspect.isclass(x) and issubclass(x, Part)
 
-        for var, part in inspect.getmembers(cls, is_part):
-            parts.append(part)
-            part._master = cls
-            # TODO: look into local namespace for the subclasses
-            process_relation_class(part, context=dict(self.context, **{cls.__name__: cls}))
+        parts = list()
+        for part in dir(cls):
+            if part[0].isupper():
```
fabiansinz: What if somebody uses a lower case name for the Part table? Do we explicitly check that when creating the part table? What if we drop that check at some point? Then we need to delete this check as well. I don't know whether the performance gain (or is there another reason?) is worth the interdependency in the code.

dimitri-yatsenko: Yes, we assume that class names are in CamelCase notation, starting with a capital letter and containing only alphanumerical characters.  No underscores.  This is assume for a consistent translation into mysql names.  The translation is performed by `datajoint.utils.from_camel_case`.

eywalker: Although I do find the check a bit weird (exclusion of `dir` names based on whether the name is capitalized), as far as the naming convention for the table goes, I believe we have been consistent in requiring that the class name be in camel case, starting with a capital letter. It would only make sense for the Part table to follow the same convention :)

dimitri-yatsenko: In fact, we should only look at names that are matched the regular expression in  `datajoint.utils.from_camel_case` rather than names that start with a capital letter.

fabiansinz: Do you plan to implement the regexp checking? I think it makes sense if we are not planning to abandon the naming convention in the future (which - after reading your comments - I guess we don't).

dimitri-yatsenko: good idea.  Let's apply the regex check.

eywalker: hey @dimitri-yatsenko, would you mind quickly implementing the regex based check of the Part relation field before I accept the pull request? Otherwise, I'll create a new issue to remind ourselves to do this later.

dimitri-yatsenko: ok

eywalker:
I'm going to accept this branch as is and merge into the `development` branch - we'll need to add stuff like regexp based check of table names.

PR #183, opened 2016-01-07
add max_allowed_packet and bugfix in insert1 for tuples with null/nan values

fabiansinz:


datajoint/base_relation.py
```diff
@@ -260,7 +261,7 @@ def make_attribute(name, value):
                 sql = 'INSERT IGNORE'
             else:
                 sql = 'INSERT'
-            attributes = (a for a in attributes if a[0])   # omit dropped attributes
+            attributes = (a for a in attributes if a[0] is not None)  # omit dropped attributes
```
eywalker: Looks like the older statement caused some valid fields to be dropped as well? Would you mind adding a test case specifically for this to ensure we don't regress in the future? :)

fabiansinz: This statement is not the problem. I just added `is not None` because it is more explicit. However, I'll add a test case for the bug above.

fabiansinz:
Added additional test.

fabiansinz:
Wrong commit message. I meant to say logic for loading configs.

eywalker:
Changes look good - I will use the new config features to generate the Docker containers.

PR #187, opened 2016-01-27
tables are not declare if definition is None or Ellipsis

fabiansinz:
So far, if a schema is defined in matlab, the corresponding class in python does not define a definition (for consistency). The syntax is 

```python
class MyMatlabTable(dj.Tier):
   definition = None
```

```python
class MyMatlabTable(dj.Tier):
   definition = ...
```

The problem is that this will raise an error if the matlab table has not yet been defined in the database. This pull request solves that by checking whether `definition=None` or `definition=...` in `BaseRelation.declare`. If that is the case, then the table is not declared and a warning is raised. 

I refrained from overriding the constructor because the user might have defined a constructor and I didn't want to mess with that.

### DIFF
--- a/datajoint/base_relation.py
+++ b/datajoint/base_relation.py
@@ -5,14 +5,13 @@ import logging
 import abc
 import binascii
 
-from . import config
+from . import config, NoDefinitionError
 from . import DataJointError
 from .declare import declare
 from .relational_operand import RelationalOperand
 from .blob import pack
 from .utils import user_choice
 from .heading import Heading
-from warnings import warn
 
 logger = logging.getLogger(__name__)
 
@@ -76,9 +75,8 @@ class BaseRelation(RelationalOperand, metaclass=abc.ABCMeta):
                     declare(self.full_table_name, self.definition, self._context))
             else:
                 name = self.__class__.__name__
-                warn("""Table for %s is not declared and definition is not defined.
-                        This table is probably defined but not yet declared in Matlab.""" % (name, ))
-                logger.info("Table for %s is not declared and definition is not defined." % (name, ))
+                raise NoDefinitionError("%s.definition is not defined and table is not in the database." % (name,))
+
 
     @property
     def from_clause(self):
--- a/datajoint/base_relation.py
+++ b/datajoint/base_relation.py
@@ -12,6 +12,7 @@ from .relational_operand import RelationalOperand
 from .blob import pack
 from .utils import user_choice
 from .heading import Heading
+from warnings import warn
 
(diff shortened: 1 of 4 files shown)


datajoint/base_relation.py
```diff
@@ -70,8 +71,14 @@ def declare(self):
...
         """
         if not self.is_declared:
-            self.connection.query(
-                declare(self.full_table_name, self.definition, self._context))
+            if self.definition is not None and self.definition is not Ellipsis:
+                self.connection.query(
+                    declare(self.full_table_name, self.definition, self._context))
+            else:
+                name = self.__class__.__name__
+                warn("""Table for %s is not declared and definition is not defined.
```
eywalker: Why not throw an error? I think it will be more helpful if it can thrown an error for debug purpose. If any code uses the table depending on its functionality, warning will lead to a hard to debug code.

datajoint/base_relation.py
```diff
@@ -70,8 +71,14 @@ def declare(self):
...
+                        This table is probably defined but not yet declared in Matlab.""" % (name, ))
```
eywalker: I'm also not sure if we should mention `Matlab` per se. You could imagine having blank definition if you expected someone else's Python code to generate the table. Let's just say that the definition is empty but the table is not defined.

eywalker:
And yes! I like the idea of not injecting constructor after the fact better too.

datajoint/base_relation.py
fabiansinz: Ok, I'll remove the reference to Maltab.

eywalker: Cool - Python users need not be tortured by the term Matlab during coding :)

datajoint/base_relation.py
eywalker: If we go with throwing an error, then the `schema` can specifically catch and ignore the error (and may be throw a warning at that point).

fabiansinz: Because that is what I want to avoid. Consider the following situation. 
* somebody writes a matlab schema and I'd like to write some functions for that tables in python
* therefore I define the corresponding classes in python
* now somebody drops some of the tables from matlab because (s)he wants to repopulate it
* so far I cannot even import my schema anymore until the tables are redefined in the database
* I am fine if my code fails if I want to instantiate an undeclared table, but I do want to be able to import my other objects to still use them until the dropped table is redfined in the database.

eywalker: Right, I understand that, but that's why just catch the specific error in the schema. I think it still makes sense to prevent actual usage of the class bound to undeclared table.

PR #188, opened 2016-01-31
automatically populate namespace with existing tables on creation of  `dj.schema`

fabiansinz:
The just instantiating the `schema` decorator populates the local namespace with class objects for all tables available in the database. So far `definition` is set to `...`.


datajoint/schema.py
```diff
@@ -1,15 +1,59 @@
...
+            class_name, class_obj = SchemaFactory.table2class[database, table_name]
+        else:
+            if table_name.rfind('__') > 1:
+                master_name, master_class = self(database, table_name[:table_name.rfind('__')])
+                class_name = class_name[len(master_name):]
+                class_obj = type(class_name, (Part,), dict(definition=...))
+                setattr(master_class, class_name, class_obj)
+                class_name, class_obj = master_name, master_class
+
+            elif table_name.startswith('__'):
```
dimitri-yatsenko: The table name prefixes should be looked up in the corresponding classes.

datajoint/schema.py
```diff
@@ -1,15 +1,59 @@
...
+                class_obj = type(class_name, (Computed,), dict(definition=..., _make_tuples=_make_tuples))
+            elif table_name.startswith('_'):
+                class_obj = type(class_name, (Imported,), dict(definition=..., _make_tuples=_make_tuples))
+            elif table_name.startswith('#'):
+                class_obj = type(class_name, (Lookup,), dict(definition=...))
+            else:
+                class_obj = type(class_name, (Manual,), dict(definition=...))
```
dimitri-yatsenko: rather than only looking at the prefixes, we should check the entire name by a regular expression.  For example, the current definition would interpret the job table `~jobs` as a manual table.

datajoint/schema.py
```diff
@@ -1,15 +1,59 @@
...
+    def __call__(self, database, table_name):
+        class_name = to_camel_case(table_name)
+
+        def _make_tuples(other, key):
+            raise NotImplementedError("This is an automatically created class. _make_tuples is not implemented.")
+
+        if (database, table_name) in SchemaFactory.table2class:
+            class_name, class_obj = SchemaFactory.table2class[database, table_name]
+        else:
+            if table_name.rfind('__') > 1:
```
dimitri-yatsenko: let's use a regular expression to split the name.

datajoint/schema.py
```diff
@@ -1,15 +1,59 @@
...
+from .user_relations import Part, Computed, Imported, Manual, Lookup
 import inspect
+
 logger = logging.getLogger(__name__)
 from warnings import warn
 
 
+class SchemaFactory:
+    """
+    Creates the appropriate python schema class objects from tables in a database.
```
dimitri-yatsenko: Would "Creates UserRelation classes for tables in the database" be clearer?

We don't need to say that these are in python.  Also, "schema class objects" is probably more confusing since these are not schema classes but classes for relations.

datajoint/base_relation.py
```diff
@@ -157,6 +157,39 @@ def _repr_helper(self):
...
+    #     return tdef
+
+    @property
+    def table_comment(self):
+        comment = ''
+        if self.is_declared:
+            td = self.connection.query(
+                'SHOW CREATE TABLE `{database}`.`{table_name}`'.format(
+                    database=self.database, table_name=self.table_name)).fetchone()[1]
+            idx = td.rfind('COMMENT=')
```
dimitri-yatsenko: Use a regular expression or pyparsing.

datajoint/base_relation.py
```diff
@@ -157,6 +157,39 @@ def _repr_helper(self):
...
+    #         exit()
+    #         #--------------------------
+    #
+    #         keys = ""
+    #         attributes = ""
+    #
+    #     return tdef
+
+    @property
+    def table_comment(self):
```
dimitri-yatsenko: This will soon be replaced with a single function that parses out the entire `CREATE TABLE` statement, not just the comment.

datajoint/schema.py
```diff
@@ -1,15 +1,59 @@
...
 from .base_relation import BaseRelation
-from .user_relations import Part
+from .user_relations import Part, Computed, Imported, Manual, Lookup
 import inspect
+
 logger = logging.getLogger(__name__)
 from warnings import warn
 
 
+class SchemaFactory:
```
dimitri-yatsenko: SchemaFactory might not be the most descriptive name.  This is effectively a function.  The only reason it's a class is because it has a state.  So we can use an action word.  I think I would call it something like `create_relation_class_from_database`.

datajoint/schema.py
fabiansinz: I'll change that to make it clearer.

datajoint/schema.py
fabiansinz: I agree that there can be better names. But I think the name `create_relation_class_from_database` is too long and I am hesitant to name a class with lower case underscore style name. To the outside this suggests that it is a function and does therefore not have a state (which is not entirely true for python functions but expected anyway). I also like the name `Factory` because it is essentially a factory design pattern.  I'll think about something clearer.

datajoint/base_relation.py
fabiansinz: If it's replaced anyway, I'll remove it entirely.

datajoint/schema.py
fabiansinz: Yes, good point.

datajoint/schema.py
fabiansinz: I can do that, but just out of curiosity: What's the advantage?

datajoint/schema.py
fabiansinz: Good point with `~jobs`. Since you came up with the naming convention, can you supply the regular expressions (pseudocode is fine)?

datajoint/schema.py
dimitri-yatsenko: yes, each `UserRelation` should return its table name regular expression.  This feature will have other uses.

datajoint/schema.py
dimitri-yatsenko: regular expressions allow more precise matching.  The naming convention involves more than just the prefixes.  Tables that don't follow the conventions should be ignored.

datajoint/base_relation.py
dimitri-yatsenko: We already parse `SHOW CREATE TABLE` in `dependencies.py` to get the foreign keys.  We also parse individual attributes in `heading.py` but there we do it from  `SHOW FULL COLUMNS` instead.  The only other thing we may need to get from the definitions are the indexes.  Now I think it's okay to keep things as you implemented them here.

datajoint/schema.py
dimitri-yatsenko: Why not make `SchemaFactory` a method of `Schema`?

datajoint/schema.py
```diff
@@ -41,6 +85,20 @@ def __init__(self, database, context, connection=None):
                 raise DataJointError("Database named `{database}` was not defined, and"
                                      " an attempt to create has failed. Check"
                                      " permissions.".format(database=database))
+        else:
+            class_factory = SchemaFactory()
+            classes = []
```
dimitri-yatsenko: What is the function of the variable `classes` here?  It does not seem to be used.

dimitri-yatsenko:
Is there a good reason why this pull request is not onto the `development` branch?  Let's make all changes in `development` before merging to `master`?  `master` will contain the stable version for users.

datajoint/user_relations.py
```diff
@@ -2,59 +2,82 @@
...
+        return self.f(owner)
+
+
+class UserRelation(BaseRelation):
     """
-    Inherit from this class if the table's values are details of an entry in another relation
-    and if this table is populated by this relation. For example, the entries inheriting from
-    dj.Part could be single entries of a matrix, while the parent table refers to the entire matrix.
-    Part relations are implemented as classes inside classes.
+    A subclass of UserRelation defines is a dedicated class interfacing a base relation.
```
dimitri-yatsenko: grammar

fabiansinz: I agree. That came from the development branch.

tests/schema_empty.py
```diff
@@ -0,0 +1,10 @@
+"""
+Sample scheme with realistic tables for testing
```
dimitri-yatsenko: "schema"

fabiansinz: That is a copy-paste duplicate from `schema.py`. Fixed it there as well.

datajoint/user_relations.py
```diff
@@ -64,29 +87,71 @@ def _prepare(self):
...
+
+class Part(BaseRelation):
+    """
+    Inherit from this class if the table's values are details of an entry in another relation
+    and if this table is populated by this relation. For example, the entries inheriting from
+    dj.Part could be single entries of a matrix, while the parent table refers to the entire matrix.
+    Part relations are implemented as classes inside classes.
+    """
+
+    _regexp = r'(' + '|'.join([c._regexp for c in [Manual, Imported, Computed, Lookup]]) + r'){1,1}' \
```
eywalker: We should add a tag `?P<master>` so we can easily extract this part later.

fabiansinz: Good idea. I thought about that, but then forgot. I added it.

PR #193, opened 2016-03-08
pymysql works with larger blobs

fabiansinz:
* since version 0.7.2 which I put in the `setup.py` requirements
* dropped the hack for large packet's in `connection.py`

### DIFF
--- a/datajoint/connection.py
+++ b/datajoint/connection.py
@@ -58,8 +58,7 @@ class Connection:
             port = int(port)
         else:
             port = config['database.port']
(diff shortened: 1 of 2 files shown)


dimitri-yatsenko:
awesome!

tstadler:
I updated my local clone of your repo from which I installed datajoint by

`python setup.py install`

but I still get the following error:

`InternalError: (1153, "Got a packet bigger than 'max_allowed_packet' bytes")`

I checked with our IT that they set the `max_allowed_packet` high enough so I have no idea what's wrong...

PR #194, opened 2016-03-17
base_relation.insert now executes in one query rather than a sequence of insert1 calls

dimitri-yatsenko:
To reduce the effect of network latency, multiple insert queries are combined into one.


datajoint/base_relation.py
```diff
@@ -171,106 +171,153 @@ def full_table_name(self):
...
+            :return: a dict with fields 'names', 'placeholders', 'values'
             """
-            for field in fields:
-                if field not in heading:
-                    raise KeyError(u'{0:s} is not in the attribute list'.format(field))
 
-        def make_attribute(name, value):
+            def make_placeholder(name, value):
+                """
+                For a given attribute `name` with `value, return its processed value or value placeholder
```
fabiansinz: "`value`" instead of "`value"

datajoint/base_relation.py
```diff
@@ -171,106 +171,153 @@ def full_table_name(self):
...
+                    """
+                    Validates that all items in `fields` are valid attributes in the heading
+                    :param fields: field names of a tuple
+                    """
+                    if field_list is None:
+                        for field in fields:
+                            if field not in heading:
+                                raise KeyError(u'{0:s} is not in the attribute list'.format(field))
+                    else:
+                        if len(fields) < len(field_list):
```
fabiansinz: Will that work if a table has default values?

fabiansinz:
Can you quickly explain your `field_list` magic? I don' really understand what is happening there and I am not sure whether it will work if the table has attributes with default values.

datajoint/base_relation.py
dimitri-yatsenko: When inserting a list of tuples, all tuples must have the same columns.  That's what is required by the SQL syntax.  If the insert relies on default values, then all tuples must rely on default values, not just some.

dimitri-yatsenko:
@fabiansinz , `field_list` serves to ensure that all rows submitted to `insert` have the same attributes.  This is required by SQL syntax.  I will add a comment to make this clear.

fabiansinz:
I get that. What I didn't see is where exactly the field list is set for the different cases. But I think I do now. Just wanted to make sure that you thought about the case of default values.

PR #201, opened 2016-03-30
Docker related enhancements

eywalker:
* Updated `Dockerfile` to creating DataJoint container by using locally available code rather than the indirect reference to GitHub content
* Added `docker-compose.yml` so that you can run `docker-compose up` to launch a Jupyter notebook server along with a MySQL server to have a complete working environment for DataJoint. Will be ideal for just testing/playing with DataJoint.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,8 +1,10 @@
 FROM eywalker/pydev
 
-MAINTAINER Fabian Sinz <sinz@bcm.edu>
+MAINTAINER Edgar Y. Walker <edgar.walker@gmail.com>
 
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -1,8 +1,10 @@
 FROM eywalker/pydev
 
-MAINTAINER Fabian Sinz <sinz@bcm.edu>
```
fabiansinz: Oops, how did that end up here.

eywalker: Sorry I overwrote it with a file I already had - didn't mean it to make a name claiming thing :(

Issue #202, opened 2016-03-30
rename `project()` to `proj()`


dimitri-yatsenko:
This may generate a heated discussion but let's have it now.

The topic is the relational projection operator.  The verb form is _project_ (stress on the last syllable).

In matlab, we implemented it as `Relvar.pro()`.  In Python, we implemented it as `RelationalOperand.project()`.

I propose to rename it to `proj` in both languages while keeping the old methods as aliases for backward compatibility.

This will make the documentation and the tutorials more consistent.

The word `project` looks too much like the noun _project_ (stress on the first syllable), which may be confusing.  I like `proj` because it does not look like the noun _project_, which users may be primed to interpret as a software project.

Since we will keep the old methods for backward compatibility, this will not affect the current code based but the new tutorial will consistently use `proj`.

eywalker:
Accepted - but you owe me one for this.

fabiansinz:
Accepted - I want one "skip feature request discussion" joker for free.

dimitri-yatsenko:
Great.  But no, we will review and question all new features.

dimitri-yatsenko:
Fixed in d29ebe1

PR #203, opened 2016-04-06
Fixed restrictions by empty AndLists and OrLists.

dimitri-yatsenko:
Restrictions by AndLists and OrLists now work correctly
- `rel & []`  is equivalent to `rel & dj.OrList()`
- `rel & []` produces  the empty relation
- `rel - []` has no effect
- `rel & dj.AndList((cond1, cond2))` is equivalent to `rel & cond1 & cond2`
- `rel & dj.AndList()` has no effect
- `rel - dj.AndList()` produces the empty relation

Also, `rel.restriction` is now of type `dj.AndList`.

Also, `RelationalOperand.project` has been renamed to  `RelationalOperand.proj`  (issue #202) with `project` aliased to `proj` for backward compatibility.


eywalker:
Looks good :)

Issue #206, opened 2016-05-14
Allow renaming the referencing attribute in a foreign key


dimitri-yatsenko:
From the start, DataJoint allowed foreign key constraints only between identically named attributes.  This convention simplified the definitions and the relational algebra. Most data definition problems can be adequately addressed under this limitation.  

However, some problems would become easier after renaming the referencing attribute.   Therefore, I suggest the following definition syntax for defining foreign key constraints
```
[attribute [=default-value]]  ->  Table
```
If the referenced table `schema.Table` has exactly one primary key attribute that is not already in the current table, then the referencing attribute will be named `attribute`. 

For example, imagine we have a table `Vertex` listing the vertices of graphs.  Then the table `Edge` may reference `Vertex` twice as  `from` and `to`.
```python
class Vertex(dj.Manual):
    definition = """  # vertex in a graph
     -> Graph
     vertex_id   : int  
     ---
     value : float  
     """

class Edge(dj.Manual):
    definition = """  # directed edge connecting two vertices
     -> Graph
     from -> Vertex 
     to -> Vertex
     ---
     weight : float   # edge weight
     """
```

The relational algebra remains unchanged.  For example, to compute the mean of the two connected vertex values for each edge, the expression will be
```python
src = Vertex().proj('node_id -> from', 'value -> v1')
dest = Vertex().proj('node_id -> to','value -> v2')
result = (src*Edge()*dst).proj('(v1+v2)/2 -> edge_value')
```

Proper renaming of attributes before joins and restrictions will become necessary.  Therefore, this approach will be used rarely when it results in significant improvements compared to the original DataJoint approach such as the example with graphs above.

dimitri-yatsenko:
We may also allow renaming multiple attributes if the referenced table has multiple new attributes:

```
class Room(dj.Manual): 
    definition = """    # rooms in buildings
    building :   char(8)   # building name
    room : char(6)   # room name or number
    """

class Animal(dj.Manual):
    definition = """ # lab animal
    animal_id:   int    # animal id 
    --- 
    housing_building, housing_room  -> Room 
    """
```

The order of the new attributes must correspond to the order of the new attributes in the referenced table.

dimitri-yatsenko:
[...]

Issue #112, opened 2015-06-30
Implement the special relation `dj.U` 


dimitri-yatsenko:
`dj.U(*attributes)` represents the relation comprising all possible values of its attributes.  It's initialized with the list of its attributes.  Its primary key comprises all its attributes.  `dj.U()` cannot be fetched from directly and is not represented in the database but it is useful in may relational expressions.

Let the relation `trials` represent some stimulus trials and `'contrast'` and `'brightness'` are two of its attributes.  Then 
```python
dj.U('contrast') & trials   # contains all unique values of contrast found in trials.
dj.U('brightness', 'contrast') & trials   # contains all unique combinations of contrast/brightness found in trials.
dj.U('contrast').aggregate(trials, 'count(*)->n')   # contains the number of trials for each unique contrast
```

Prior to `dj.U`, the same functionality could be achieved by creating a dedicated lookup table containing all values of contrast (for the first example) or combinations of brightness and contrast (for the second example).  `dj.U` allows avoiding creating extra tables that are required only for the convenience of relational expressions.

Here is a more complex example.  If there are two relations containing trials: `trials1` and `trials2`.  
```python
(dj.U('contrast') & trials2) - trials1  # contains all contrasts from trials2 that are not found in trials1
```

The special case with the empty attribute list `dj.U()`  represents the relation with one tuple that has no attributes.  As such, it matches any tuple.  It can be thought as the identity relation with respect to join and restrict:  `rel*dj.U() == rel` and `rel & dj.U() == rel`.  It becomes useful for the aggregation of an entire relation:
```python
dj.U().aggregate(trials, 'count(distinct contrast)')    # count the number of distinct contrasts in trials
```

dimitri-yatsenko:
I am thinking to call this relation `dj.U` (for "universal") rather than `dj.All`.

eywalker:
@dimitri-yatsenko has already convinced me that `dj.U` is a good choice.

fabiansinz:
Agreed. it's a good name

dimitri-yatsenko:
implemented in 4debbd88f

Issue #82, opened 2015-05-18
ERD plotting support


eywalker:
The logic for creating and manipulating graphs that represent the ERD is pretty much there and can be found in the `erd` branch. I will keep this branch as the feature implementation branch. There are a few issues that I'm running into as I try implementing plotting support:

1. `pygraphviz` library that was once used in our project for plotting ERD does not support Python 3 natively **yet**. Explicitly using their beta branch with `pip install pygraphviz==1.3rc2` allows us to use it in Python 3, but I'm not sure if we would like to introduce a dependency on beta-version of the package, especially as this is a very specific dependency.
2. `pydot` is an alternative package that may be used, but again there is no official Python 3 support. There appears to be a port `pydot3k`

The above two were the preferred method to work with `graphviz` graph plotting engine, that yields very nicely arranged graphs. Alternatively, I could try working with `graphviz` directly, but to my knowledge, the library can only render files like `.pdf` files at the end and it takes some effort to display this into `matplotlib`.

dimitri-yatsenko:
Fixed in PR #208

PR #208, opened 2016-05-24
Redesigned Dependencies, completed ERD, implemented relation U. 

dimitri-yatsenko:
- foreign keys can now use aliased attributes (See #206)
- ERD is now fully functional (See #7, #82 )
- `dj.ERD` is now a standalone object for plotting dependencies
- `dj.ERD` now has a custom layout independent on `pygraphviz`
- simplified Fetch and removed slice indexing
- `connection.dependencies` is now a `networkx.DiGraph`
- removed text-based ERD
- Schema now has method `spawn_missing_classes` to explicitly bring missing classes into context; it should be invoked after all the other classes have been declared, e.g. at the end of the module.
- The `_prepare` callback has been removed.  If a user relation has a `contents` property, it will be inserted only once at the time of the table declaration. 
- All RelationalOperand classes have a copy constructor
- Implemented `dj.U` (previously proposed as `dj.All` in issue #112)
- `AutoPopulate._populated_from` has been renamed into `poprel`
- removed the `demos` directory
- added IPython notebook tutorial pages


datajoint/__init__.py
```diff
@@ -6,18 +6,25 @@
 
 DataJoint is free software under the LGPL License. In addition, we request
 that any use of DataJoint leading to a publication be acknowledged in the publication.
+
+Citations:
```
fabiansinz: Minor Comment: Maybe that should be renamed in 'Please cite us:'. Otherwise it sounds as if these are citations of datajoint.

datajoint/__init__.py
```diff
@@ -6,18 +6,25 @@
...
+    http://biorxiv.org/content/early/2015/11/14/031658
+    http://dx.doi.org/10.1101/031658
 """
 
 import logging
 import os
 
 __author__ = "Dimitri Yatsenko, Edgar Walker, and Fabian Sinz at Baylor College of Medicine"
-__version__ = "0.2"
+__version__ = "0.2.1"
```
fabiansinz: Shouldn't that be 1.0? I thought you killed of all issues for milestone 1.0.

datajoint/autopopulate.py
```diff
@@ -16,28 +17,27 @@ class AutoPopulate(metaclass=abc.ABCMeta):
     """
     AutoPopulate is a mixin class that adds the method populate() to a Relation class.
     Auto-populated relations must inherit from both Relation and AutoPopulate,
-    must define the property populated_from, and must define the callback method _make_tuples.
```
fabiansinz: Why are we changing this to matlab style? I liked the more verbose version `populated_from`.

datajoint/autopopulate.py
```diff
@@ -118,12 +122,12 @@ def populate(self, *restrictions, suppress_errors=False, reserve_jobs=False, ord
                             jobs.complete(self.target.table_name, key)
         return error_list
 
-    def progress(self, restriction=None, display=True):
+    def progress(self, *restrictions, display=True):
```
fabiansinz: Why not make `**kwrestrictions` as well?

datajoint/erd.py
```diff
@@ -1,473 +1,272 @@
...
+        graph = self._make_graph(prefix_module)
+        if pos is None:
+            pos = (layout if layout else self._layout)(graph, **layout_options)
+        import matplotlib.pyplot as plt
+
+        edge_list = graph.edges(data=True)
+        edge_styles = ['solid' if e[2]['primary'] else 'dashed' for e in edge_list]
+        nx.draw_networkx_edges(graph, pos=pos, edgelist=edge_list, style=edge_styles, alpha=0.2)
+
+        label_props = { # http://matplotlib.org/examples/color/named_colors.html
```
fabiansinz: That could go up on module level so because it

- does not need to be defined every time
- can be modified by the user there

datajoint/fetch.py
```diff
@@ -1,71 +1,42 @@
...
 
-class Fetch(Iterable, Callable):
+class Fetch(FetchBase, Callable, Iterable):
     """
     A fetch object that handles retrieving elements from the database table.
 
     :param relation: relation the fetch object retrieves data from.
     """
 
     def __init__(self, relation):
```
fabiansinz: Why do we get rid of the copy constructor and copying the fetch object before it is modified?

datajoint/fetch.py
```diff
@@ -199,15 +153,11 @@ def __getitem__(self, item):
         :return: tuple with an entry for each element of item
 
         Examples:
-
         >>> a, b = relation['a', 'b']
         >>> a, b, key = relation['a', 'b', datajoint.key]
-        >>> results = relation['a':'z']    # return attributes a-z as a tuple
```
fabiansinz: Why is that being deleted? Did you delete the slicing functionality? If so, why?

datajoint/__init__.py
dimitri-yatsenko: I just incremented the version.  When we have a real 1.0 candidate, we can call it 0.9.1 or something.

datajoint/autopopulate.py
dimitri-yatsenko: example?  We do not have a concept of kw restrictions in the restrict operator.

datajoint/fetch.py
dimitri-yatsenko: yes, slicing and number indexing of attributes is an invitation to write bad code, relying on the order and number of attributes in a relation rather their names.  This feature adds the testing and documentation load while its functionality could be straightforwardly accomplished by other, more explicit means.

datajoint/autopopulate.py
fabiansinz: Yes we do every time we restrict by a dictionary.

datajoint/fetch.py
fabiansinz: For a given relation the order or attributes is fixed. Why is that bad code?

datajoint/erd.py
dimitri-yatsenko: we compute the font size. See `font_scale`

datajoint/fetch.py
dimitri-yatsenko: Attributes are defined by name, not by order.  It's one of the basic premises of the relational data model.  The order of the attributes does not matter.  Adding a new attribute to a relation should not break any code.  Besides, this feature can be accomplished by something like
```python
attrs = rel.heading.names
rel.fetch[*attrs[3:12]]
```

datajoint/autopopulate.py
dimitri-yatsenko: Yes, restriction can be a dict, which is not the same as **kwargs.

fabiansinz: I think this syntax would be useful: `rel.progress(animal_id=5000)`

datajoint/autopopulate.py
dimitri-yatsenko: When I write tutorials and record presentations, I have to use `poprel` as a noun and it's much easier to discuss than `populated_from`.  It's a new term that we introduce in datajoint, which has specific uses. It has been used in MATLAB from the early versions, so changing it there is not practical.  Naming it the same way in both languages helps.

fabiansinz: I breaks all computed tables that have been written so far.

datajoint/fetch.py
dimitri-yatsenko: I agree that positional insert contradicts the above principle.  I was hesitant to add it.  But the coding and execution efficiency improvement is much greater for the positional insert than for fetch slicing.

Slicing of the form `["string":"string":3]` is unfamiliar and therefore not expressive to new users.

datajoint/autopopulate.py
dimitri-yatsenko: it's interesting but not used anywhere else yet.  Yes, we could add this syntax.  The `progress` should not use any other arguments such as `display` or may be prefix it as `_display` to avoid name collisions.

datajoint/autopopulate.py
dimitri-yatsenko: Yes, we can either fix them or provide `populated_from` alias for  backward compatibility.

datajoint/autopopulate.py
dimitri-yatsenko: I am hesitant to do that since `progress` has the same signature as `populate`, which takes a lot more named arguments.

datajoint/autopopulate.py
dimitri-yatsenko: most computed tables rely on the default `poprel` and are not affected.  Very few tables should modify it.

datajoint/autopopulate.py
dimitri-yatsenko: I agree. It looks elegant enough that we should implement it for both `progress` and `populate`.  Name collisions should be obvious to users and they will avoid them.  Does this look confusing?
```python
ScanInfo().populate(animal_id=5014, reserve_jobs=True, suppress_errors=True)
```

datajoint/autopopulate.py
fabiansinz: A good portion of my computed and imported tables rely on this feature for more efficient computation. Let's not make things complicated by providing several ways of doing the same thing. Calling it `populated_from` was a team decision, it is descriptive, and the ease with which something can be used in discussions should not be a primary reason to choose property names. If most computed tables rely on the default value, why not change it in Matlab?

datajoint/autopopulate.py
fabiansinz: I'd use the keyword arguments in a different order for clarity, but it would be fine with me. I think we should either allow `*restrictions` and `**kwrestrictions` or neither and have a simple `restrictions` argument.

datajoint/autopopulate.py
dimitri-yatsenko: MATLAB does not yet have a default `popRel` implemented, so every computed class does have one.

The `poprel` feature needs to be documented fairly well.  Try using 'populated_from` in a sentence.  It is easy to define a `poprel` in the first paragraph and then use it throughout.  From the name alone, no one will understand what `populated_from` actually is, so the name is not that descriptive.  Might as well give it a name and describe it in the tutorial.

datajoint/autopopulate.py
dimitri-yatsenko: okay, let use a simple `restrictions` argument.  It will be cleaner.

datajoint/heading.py
```diff
@@ -184,19 +191,19 @@ def init_from_database(self, conn, database, table_name):
                     attr['dtype'] = numeric_types[(t, is_unsigned)]
         self.attributes = OrderedDict([(q['name'], Attribute(**q)) for q in attributes])
 
-    def proj(self, *attribute_list, **renamed_attributes):
+    def proj(self, attribute_list, renamed_attributes, include_primary_key):
```
fabiansinz: Can we factor that into a function `_proj(attribute_list, renamed_attributes, include_primary_key)` and a function `proj(self, *attribute_list, **renamed_attributes)` that simply calls `_proj`? I really liked the old syntax of `project`. I can see why we want to have the other one, but I'd still like to keep the old syntax because

- it is significantly more effort to refactor old code (instead of simply renaming `project` into `proj`)
- it is a lot cleaner to call `rel.proj('myattr',new_attr='func(old_attr)')` instead of `rel.proj(['myattr'],dict(new_attr='func(old_attr)'})`

eywalker:
Looks good :smile:

Issue #214, opened 2016-06-12
Deprecate the "expression -> new_attribute" form of projection arguments


dimitri-yatsenko:
Currently, we can rename attributes and new attributes using two notations
```python
result = rel.proj(cos = 'cos(ori)')
```
and 
```python
result = rel.proj('cos(ori)->cos')
```
The latter notation comes from the MATLAB version of DataJoint and I don't see it used often.  The code and the documentation will become simpler if we remove the second form from python. 

Any objections?

fabiansinz:
Since there is a substantial difference in the syntax of DataJoint Matlab and Python already, I don't see why we need to keep this particular one. So, no objections.

dimitri-yatsenko:
okay, the upcoming release will deprecate this feature.  All *args in projection and aggregation must be simple attribute names.

dimitri-yatsenko:
one problem with `**kwargs` is that they come unordered.  The use has no control of the order of the attributes in the resulting projection.

fabiansinz:
Based on the discussion we had on the `__getitem__` that does not matter.

dimitri-yatsenko:
exactly. One more reason not to support numeric indexing of attributes.

dimitri-yatsenko:
I have also quietly dropped the % notation for projections.  We used to be able to do this:
```python
rel % ('one', 'sum(two)->n')
```
as the equivalent of 
```python
rel.proj('one', n='sum(two)')
```
Only the latter remains now.

fabiansinz:
I think nobody used that anyway.

dimitri-yatsenko:
fixed in 4debbd88f

Issue #177, opened 2015-12-10
Add __repr__ to class AndList


dimitri-yatsenko:
We added the `AndList` to handle lists of restrictions in `Relation` but it lacks a `__repr__` method, so our restricted relations now look ugly.

dimitri-yatsenko:
I am about to close it as a won't fix.  

Developing a meaningful `__repr__` for relations and their restrictions is a difficult problem that's hard to test and is therefore prone to bugs.   Developers have better tools for inspecting objects than the `repr`.

dimitri-yatsenko:
addressed in 4debbd88f

Issue #163, opened 2015-10-02
add __repr__ and __str__ to datajoint.schema


dimitri-yatsenko:
These may display the list of tables contained in the schema and perhaps their dependencies

dimitri-yatsenko:
fixed in 4debbd88f

Issue #136, opened 2015-07-25
`__str__` and `__repr__` for RelationalOperand


eywalker:
Currently the `__repr__` for `RelationalOperand` prints out the heading as well as the first few rows of the query. Although this is a very useful representation for a quick inspection of the table (or result of relational operations) content, it makes debugging quite annoying as quick inspection of the `RelationOperand` both executes the query and prints out often times massive table. As succinctly stated in Python's documentation on [`__repr__`](https://docs.python.org/3/reference/datamodel.html#object.__repr__), the `__repr__` is meant to be used during debugging, providing *unambiguous* information about the object (this concept is also well described [here](http://satran.in/2012/03/14/python-repr-str.html)).

Given this, I propose two alternatives:

1. Move the current verbose string representation found in `__repr__` into `__str__`, and supply more succinct but useful-for-debugging string representation in `__repr__` (e.g. relational operation that yielded the current object). In fact, if `__repr__` can contain the full description of how it was produced, that'd be useful. If a user would want to see what the relation's content looks like, it can be viewed via `print(relation)`

2. Since I understand that being able to just type in the name of relation at the interactive Python to get content preview is quite nice, we could add a config flag that controls the behavior of the `__repr__` . We would ship DataJoint with this config flag default to something like `preview`, causing `__repr__` to print out the preview of the table content as it does already. When debugging, the config flag can be set to `debug` with more succinct `__repr__` behavior.

Although I think option 1  follows the mantra of Python language, packages like `pandas` use option 2, defaulting to verbose `__repr__` useful during interactive programming.

fabiansinz:
I think option 1) would be the clean way, but I think we probably want option 2). You could check the debug level in `dj.config` and change the behavior of `__repr__` if it is set to `DEBUG`.

dimitri-yatsenko:
How about moving the table preview into the fetch object?

eywalker:
We'll have it such that `__repr__` on `RelationalOperand` only outputs information about how it was constructed. To view the content preview, user can simply look at the `rel.fetch` - the `__repr__` for `Fetch` object will provide with the preview of the content. This is consistent with the idea that it's only `Fetch` object that will execute the query.

dimitri-yatsenko:
With all the new features, constructing a useful representation for debugging is very difficult and requires additional computations.  I don't think it is reasonable to produce a very informative `__repr__`.  Developers have much better tools for inspecting the objects than `__repr__`.

[...]

PR #215, opened 2016-06-15
Optimized relational operators

dimitri-yatsenko:
- `dj.U` is no longer a `RelationalOperand` -- it only supports a limited number of operations, which produce relational operands. 
- `GroupBy` now uses a `HAVING` clause for restrictions, avoiding an extra subquery. 
- Added `__repr__` to `Schema`. 
- Closing issues #163, #136, #214, #177, #112 -- these should all be fixed.

### DIFF
--- a/datajoint/heading.py
+++ b/datajoint/heading.py
@@ -3,33 +3,39 @@ from . import DataJointError
 from collections import namedtuple, OrderedDict
 import re
 
+default_attribute_properties = dict(    # these default values are set in computed attributes
+    name=None, type='expression', in_key=False, nullable=False, default=None, comment='calculated attribute',
+    autoincrement=False, numeric=None, string=None, is_blob=False, sql_expression=None, dtype=object)
 
-class Attribute(namedtuple('Attribute',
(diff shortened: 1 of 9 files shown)


datajoint/heading.py
```diff
@@ -191,70 +198,46 @@ def init_from_database(self, conn, database, table_name):
                     attr['dtype'] = numeric_types[(t, is_unsigned)]
         self.attributes = OrderedDict([(q['name'], Attribute(**q)) for q in attributes])
 
-    def proj(self, attribute_list, renamed_attributes, include_primary_key):
+    def project(self, attribute_list, named_attributes, force_primary_key=None):
```
fabiansinz: You put project back? I thought it is named `proj` now.

dimitri-yatsenko: No.  This is a method of `Heading` -- It is a private method that should not be used by external users.  I renamed it because it made it easier to search.  `RelationalOperand` still only has `proj`.

fabiansinz: Sorry. Always run into this error. I got confused because it shows a deleted line `proj` above it.

PR #216, opened 2016-06-15
Reconnect and bugfixes

fabiansinz:
- automatic reconnect upon a `pymysql.err.OperationalError`; can switch that off in `config`
- fix bug in `contents` insert if `contents` is an iterator
- if table has autoincrement, the contents is ignored with a warning because it would insert everytime otherwise.


datajoint/connection.py
```diff
@@ -104,9 +116,19 @@ def query(self, query, args=(), as_dict=False):
...
-
+        try:
+            logger.debug("Executing SQL:" + query[0:300])
+            cur.execute(query, args) # TODO insert reconnect
+        except err.OperationalError as e:
+            if config['database.reconnect']:
+                print(e)
+                warnings.warn('''Mysql server has gone away.
+                    Reconnected to the server. Data from transactions might be lost and referential constraints may
+                    be violated. You can switch of this behavior by setting the 'database.reconnect' to False.
```
dimitri-yatsenko: "switch off"

fabiansinz: fixed it ... classic

dimitri-yatsenko:
tests are not passing.  Please fix.

PR #221, opened 2016-06-24
re-execute query after reconnection

fabiansinz:


dimitri-yatsenko:
Let's increment the minor version with every bugfix.

datajoint/connection.py
```diff
@@ -122,6 +122,8 @@ def query(self, query, args=(), as_dict=False):
                     be violated. You can switch off this behavior by setting the 'database.reconnect' to False.
                     ''')
                 self.connect()
+                logger.debug("Re-Executing SQL:" + query[0:300])
```
dimitri-yatsenko: I think the default setting should be to disable reconnects.  This is a compromise we have sometimes made but automatic reconnects may have deleterious effects.

PR #239, opened 2016-08-01
add update and tests

fabiansinz:


datajoint/base_relation.py
```diff
@@ -359,6 +359,49 @@ def show_definition(self):
...
+
+            Example
+
+            >>> (v2p.Mice() & key).update('mouse_dob',   '2011-01-01')
+            >>> (v2p.Mice() & key).update( 'lens')   # set the value to NULL
+
+        """
+        assert len(self) == 1, 'Update is only allowed on one tuple at a time'
+        assert attrname in self.heading, 'invalid attribute name'
+        assert attrname not in self.heading.primary_key, 'cannot update a key value. '
```
dimitri-yatsenko: Use `raise DataJointError` instead of `assert` 

Formally, `assert` is used for debugging and may be turned off in production.

Issue #244, opened 2016-09-13
Serializing non `ndarray` objects


eywalker:
Currently BLOB field on Python side only accepts inserting `ndarrays` of simple data types. We should provide broader serialization support on the Python side.

This, however, brings up an issue of how _compatible_ we should keep Python serialization with DataJoint matlab. Namely, should we allow Python to serialize objects that cannot be deserialized in Matlab side? Equivalently, should we allow Python to serialize via serialization scheme other than `mYm`?

aecker:
I would vote against serialization of arbitrary data types, since it kind of defeats the purpose of a relational database and almost surely causes all sorts of trouble and complexities that we'd rather not deal with. What would be a use case where serialization of complex data types would be required or convenient? 

I think serializing 'basic' data types is sufficient. 

Since this came up in the context of the discussion around storing PKs as structs in the jobs table: I don't see a simple alternative to the scheme using a struct, so maybe serialization of dicts with simple data types is a good compromise?

fabiansinz:
We could store them as json and deserialize them in structs and dicts on the respective ends. Newer versions of MySQL support json. https://dev.mysql.com/doc/refman/5.7/en/json.html

dimitri-yatsenko:
We should take advantage of JSON support since MySQL and other RDBMS will provide optimized searches by data in JSON fields.

We need to add support for JSON on both sides: MATLAB and Python.  We will do so with backward compatibility.  Currently, BLOB data is prefixed with "mYm" indicating mym serialization.  We will use prefix "json" for future structured blobs.  Let's phase out mym serialization.

dimitri-yatsenko:
Although normalization is almost always a good thing, many people prefer denormalized or "schemaless" designs.  Supporting the JSON datatype provides support for both philosophies and searches by and within the JSON fields will be quite flexible too.

dimitri-yatsenko:
I am okay with no longer supporting mym structures and storing in JSON from now on.  We will add JSON support in both languages right away.

fabiansinz:
Before dropping mym, let's make sure we cover all cases. We should definitely be backwards-compatible as well.

dimitri-yatsenko:
I don't know if we should implement mym serialization in Python.  Deserialization already works.  I would rather add JSON support to both.  This should be done in the format that allows MySQL JSON functionality, i.e. no compression of the json structure.

fabiansinz:
[...]

Issue #243, opened 2016-09-13
Insert target table primary key in job table


eywalker:
Currently the primary key of the target table that is getting populated via `reserve=True` is _not_ inserted into the `jobs` table, making it difficult to identify on which primary key on the target table that the job is working on, or has encountered an error when populating.

The primary key of the target table should be inserted at the time of the job reservation.

eywalker:
Insertion of target table primary key will be implemented as soon as issue #244 is resolved.

dimitri-yatsenko:
This has been fixed for some time now.  Need to add a test and close.

dimitri-yatsenko:
This was fixed in #316.  Closing.

PR #248, opened 2016-09-17
Add BLOB support for dict of simple types

eywalker:
A dictionary with valid serializable field values (e.g. `np.ndarray` and scalars) can now be serialized and inserted into BLOB. This helps partly address issue #244 and in particular will allow us to solve the issue #243. 

When fetching back BLOB of structure array and if the structure array is a singleton, you can do the following to unpack as dict:

```python
dj.blob.unpack(blob, as_dict=True)
```

This is used to facilitate the testing procedure.

### DIFF
--- a/datajoint/blob.py
+++ b/datajoint/blob.py
@@ -3,7 +3,7 @@ Provides serialization methods for numpy.ndarrays that ensure compatibility with
 """
 
 import zlib
-from collections import OrderedDict
+from collections import OrderedDict, Mapping, Iterable
 import numpy as np
 from . import DataJointError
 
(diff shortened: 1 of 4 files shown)


datajoint/blob.py
```diff
@@ -201,41 +212,95 @@ def __str__(self):
...
+
+def pack_obj(obj):
+    blob = b''
+    if isinstance(obj, np.ndarray):
+        blob += pack_array(obj)
+    elif isinstance(obj, Mapping):  # TODO: check if this is a good inheritance check for dict etc.
+        blob += pack_dict(obj)
+    elif isinstance(obj, str):
+        blob += pack_array(np.array(obj, dtype=np.dtype('c')))
+    elif isinstance(obj, Iterable):
```
fabiansinz: If `obj` is an iterator, then `np.array` will not exhaust it. Replace with `list(obj)`.

eywalker: Working right on

eywalker: I've made the requested change and added tests specific to this point.

Issue #246, opened 2016-09-15
`dj.U()` with no argument doesn't work as expected


eywalker:
Given a sample table such as the one below:

```python
@schema
class Test(dj.Computed):
    definition="""
    key   :   int     # key
    ---
    value   :   int     # value
    """
    contents = [(k, k) for k in range(10)]
```

The following query using `dj.U()` with no argument

```python
dj.U().aggregate(Test(), vmax='max(value)')
```

is expected to return the maximum value of the column `value` across all tuples in `Test`. However, the query results in the following error instead:

```python
InternalError: (1054, "Unknown column '' in 'group statement'")
```

Inspection of SQL gives:

```sql
SELECT max(value) as `vmax` FROM `dj_test`.`__test` GROUP  BY ``
```

dimitri-yatsenko:
yes, the `GROUP BY` clause should be omitted when the grouping attributes are empty.

eywalker:
Cool, thanks!

dimitri-yatsenko:
fixed in #251.

eywalker:
PR #251 merged and verified to fix this issue.

Issue #247, opened 2016-09-15
Easier `argmax`


eywalker:
Currently finding a tuple with a column value equal to the maximum value of that column (`argmax`) is not easy at all. Given an example table below:

``` python
@schema
class Test(dj.Computed):
    definition="""
    key   :   int     # key
    ---
    value   :   int     # value
    """
    contents = [(k, k) for k in range(10)]
```

A suggested solution is to do

``` python
rel = Test()
rel * dj.U().aggregate(rel, vmax='max(value)') & 'value=vmax'
```

Aside from the fact that the above query currently doesn't work (refer to issue #246), the query is simply not intuitive even for an experienced user for an operation as simple as `argmax`. Of course you can do something like:

``` python
vmax = Test().fetch['value'].max()
Test() & ('value = %r' % vmax)
```

but it would be nice to be able to do this entirely in DataJoint.

dimitri-yatsenko:
The following will work even if it may defy some intuition.
```python
rel & dict(value=rel.proj('vmax=max(value)').fetch['vmax'])
```

Argmax is not untuititive in the relational model in general.  How about we define as a separate max method?
```python
rel & dict(value = rel.max['value'])     # restrict to value = max(value)   
```
This will require two transactions and complicates algebraic closure.  

We could have a separate restriction object `argmax` that would work like this:
```python
rel & dj.argmax('value')   # restrict to value=max(value)
```

dimitri-yatsenko:
Or even 
```python
rel.maximize('value')    #  restrict to value=max(value)    
```

eywalker:
I agree that whatever the solution we come up with, we should not break it into two separate queries. I'd have to think more as to what would be a good solution form.

dimitri-yatsenko:
the previous two solutions run in one query but are not general enough.

dimitri-yatsenko:
We can solve this by introducing two operators:
```python
rel.maximum('value')     # restricts to value=max(value)
``` 
and 
```python
rel.max(vmax='value')   
```
which is a relation comprising one tuple with one attribute `vmax`, which is the max value of `value`) 

For example, 
```python
Test()*Lookup().max(vmax='value') & 'param=vmax'   # all tests in which param = max value from Lookup
```

dimitri-yatsenko:
As designed now, after the bugfix, the syntax would be
```python
dj.U().aggregate(rel, vmax='max(value)') * rel & 'value=vmax'
```

eywalker:
[...]

PR #251, opened 2016-09-27
Fixed issue #246 and #247

dimitri-yatsenko:


datajoint/relational_operand.py
```diff
@@ -229,6 +229,12 @@ def aggregate(self, group, *attributes, keep_all_rows=False, **named_attributes)
         return GroupBy.create(self, group, keep_all_rows=keep_all_rows,
                               attributes=attributes, named_attributes=named_attributes)
 
+    def aggr(self, *args, **kwargs):
```
eywalker: It looks like you have two aggr method definitions, here and at line 743?

dimitri-yatsenko: They belong to different classes.  The one below is for `dj.U` which does not subclass from `dj.RelationalOperand`.

eywalker: This aggr calls itself, falling into infinite loop.

dimitri-yatsenko: thanks for catching this..  fixed.

datajoint/schema.py
```diff
@@ -123,15 +122,16 @@ def process_relation_class(self, relation_class, context, assert_declared=False)
...
-            contents = list(instance.contents)
-            if len(contents) > len(instance):
-                if instance.heading.has_autoincrement:
-                    warnings.warn(
-                        'Contents has changed but cannot be inserted because {table} has autoincrement.'.format(
-                            table=instance.__class__.__name__))
-                else:
-                    instance.insert(contents, skip_duplicates=True)
+        if instance.is_declared:
+            if hasattr(instance, 'contents'):
```
eywalker: Should we just combine these two if statement into one?

eywalker: Should we just combine these two if statement into one?

eywalker: Should we just combine these two if statement into one?

tests/test_relation_u.py
```diff
@@ -50,7 +50,13 @@ def test_invalid_join():
         rel = dj.U('language') * dict(language="English")
 
     def test_aggregations(self):
```
eywalker: I think we should have tests for `aggregate` and `aggr` methods on non `dj.U` as well.

datajoint/schema.py
dimitri-yatsenko: fixed

tests/test_relation_u.py
dimitri-yatsenko: meaning?

eywalker: In `test_relational_operand.py`, under `TestRelational.test_aggregate()`, could you add one or more tests that use `aggr` in place of `aggregate`? It would be easiest for you to add this as the PR is still open.

dimitri-yatsenko: just did it

datajoint/schema.py
dimitri-yatsenko: fixed

dimitri-yatsenko: fixed

tests/test_relation_u.py
eywalker: beautiful, thanks.

dimitri-yatsenko: so what's the process for merging now after review?  Who is supposed to merge?

PR #267, opened 2016-11-29
improved table appearance in jupyter

dimitri-yatsenko:


dimitri-yatsenko:
Old look:
old look

New look:
new look

The primary key is black while non-primary attributes are in white.

fabiansinz:
First!

eywalker:
I agree this is an improvement over the existing appearance. We should perhaps consider making the CSS style part of configurable or at least make it easily overridable.

Issue #268, opened 2016-12-06
Missing INSERT privileges for log table results in error


aecker:
Due to the new log table users with only SELECT privileges on a schema cannot access this schema anymore. Instead, they get an error since entering the log entry fails.

A possible workaround is to give the user access to only the log table via

    GRANT INSERT ON `schema`.`~logs` TO 'user'

However, this is cumbersome, since it needs to be done for every user and schema, and tools like e.g. MySQL Workbench do not show (or let you edit) per-table access rights, requiring the admin to do this via the command line.

The simplest solution would be to catch the error and skip logging in this case. Since the user has only SELECT privileges, there isn't anything (s)he can do that needs to be logged anyway (assuming that the user does not have additional per-table privileges declared for other tables in this schema).

Alternatively, we could store the logs in a separate schema, simplifying the access privileges.

fabiansinz:
I am in favor of skipping logging and throwing a warning.

dimitri-yatsenko:
okay, I think we should just use the `logging` module to log the warning.  
```python
logging.warn('Insufficient privileges to make a schema log entry in `database`.')
```

aecker:
That works of course, but the problem remains that the admins have to manage table-level access privileges for each user and schema or the users keep getting annoying warnings. Doesn't strike me as a very clean solution. What's the downside of a log schema?

fabiansinz:
Might be easy to forget to set it up and give all users insert rights on it. The warning solution always works. However, I am not strongly opposed to the logging schema idea.

dimitri-yatsenko:
You may disable `logging.warn` display

dimitri-yatsenko:
I am not sure about creating a separate schema for logging.  It seems nice to keep all related information in the same schema.  I am open to the idea.

fabiansinz:
I thought so as well initially. But a schema for all the logging might actually be more comfortable when you need to track things across several schemata. It'd be all in one place.

dimitri-yatsenko:
would the job reservations also be in that special schema?

dimitri-yatsenko:
It seems a bit insecure.  Users can view their logs.  They should not see other people's logs.  So we are still not avoiding the problem of table-level access control.  

I vote for keeping the logs and the jobs and all the other auxiliary tables in the same schema that they describe.

dimitri-yatsenko:
I think logging.warn is not displayed by default anyway.

dimitri-yatsenko:
[...]

Issue #270, opened 2016-12-06
Bug in argmax with `dj.U`


fabiansinz:
Consider the following schema (added in #269).
```python
@schema
class ArgmaxTest(dj.Lookup):
    definition = """
    primary_key     : int
    ---
    secondary_key   : char(2)
    val             : float
    """

    n = 10

    @property
    def contents(self):
        n = self.n
        yield from zip(range(n ** 2),
                       itertools.chain(*itertools.repeat(
                                         tuple(map(chr, range(100, 100 + n))), n)),
                       np.random.rand(n ** 2))
```

If I want to perform an argmax for each secondary key, the syntax should be 
```python
(dj.U('val') * rel) & dj.U('secondary_key').aggr(rel, val='min(val)')
```
However, the only syntax that works is
```python
(dj.U('val') * rel) * dj.U('secondary_key').aggr(rel, val='min(val)')
```

Shouldn't they both give the same result?

dimitri-yatsenko:
yes, seems right.

eywalker:
This should be fixed with #271 and appropriate tests were added. @fabiansinz please verify the fix and I'll close this issue.

Issue #257, opened 2016-10-05
Views


dimitri-yatsenko:
I propose to add the `dj.View` class.  Essentially it wraps around a query and makes it look like a base relation or its own table.  But it does more than that.  For example, it can optimize the query by adding indexes on the queried tables.

For example, one of the most common queries for us is to get trials that correspond to a scan.  The query looks something like this:
```python
vis.Trial() * preprocess.Sync() & 'trial_idx between first_trial and last_trial'
```
We could now define the class
```python
@schema
class Trial(dj.View):
    definition = vis.Trial() * preprocess.Sync() & 'trial_idx between first_trial and last_trial'
```
and then use it in queries as
```python
(Trial() & key).fetch['flip_times']
```
Underneath, `dj.View` uses SQL's `CREATE VIEW` command to accommodate these queries.

dimitri-yatsenko:
Views must be decorated with `@schema` since views are created within databases just like tables.

Creating  a view requires the `CREATE VIEW` privilege which we do not always provide.

dimitri-yatsenko:
Views will not show in the ERD.

dimitri-yatsenko:
implemented in #258 as `dj._View` for now.  As we identify all the use cases, we will rename to `dj.View`.

dimitri-yatsenko:
This feature was removed in #271 as practical use of views had numerous issues and did not provide sufficient performance increase over derived relations.

PR #271, opened 2016-12-06
Fixed issues #268 and #270, removed support of Views.

dimitri-yatsenko:
* debugged aggregation (issue #270)
* datajoint can now be imported without a connection attempt (it was a bug)
* dropped support of views (reversed #257)
* fixed issue #268 (handling of insufficient CREATE and INSERT privileges)

Views caused a few problems with creation and dropping but did not provide tangible benefits.


eywalker:
Looks like Coveralls is hanged. Everything looks good, merging.

PR #272, opened 2016-12-06
minor fix for previous commit

dimitri-yatsenko:
causes SELECT * to be used rather tha SELECT attr1, ..., attrN when possible.


eywalker:
Good catch.

PR #274, opened 2016-12-20
Report how many keys need to be populated

eywalker:
Unlike in Matlab, `populate` does not report how many keys still have to be populated at the beginning of the `populate` invocation. This information is a useful quick indicator.

### DIFF
--- a/datajoint/autopopulate.py
+++ b/datajoint/autopopulate.py
@@ -81,6 +81,7 @@ class AutoPopulate:
 
         jobs = self.connection.jobs[self.target.database] if reserve_jobs else None
         todo -= self.target.proj()
+        n_to_populate = len(todo)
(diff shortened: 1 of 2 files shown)


datajoint/autopopulate.py
```diff
@@ -89,6 +89,7 @@ def populate(self, *restrictions, suppress_errors=False, reserve_jobs=False, ord
             keys = list(keys)
             random.shuffle(keys)
 
+        logger.info('Found %d keys to populate' % len(todo))
```
dimitri-yatsenko: `len(todo)` is an extra query, which may take some time.  Would you extract the length from `keys` instead?

eywalker: Changes made.

PR #286, opened 2017-02-06
Improved handling of config settings.

dimitri-yatsenko:
* `dj.config` now has methods `save_local` and `save_global` for saving the settings explicitly.
* Removed support for vagrant and for readthedocs.
* Updated CHANGELOG.md


datajoint/admin.py
```diff
@@ -1,15 +1,15 @@
...
-    if 'yes' == user_choice('Do you wish to change the password?'):
-        connection = conn() if connection is None else connection
-        connection.query("SET PASSWORD = PASSWORD('%s')" % new_password)
-        print('Done!')
-    else:
-        print('Password left unchanged')
+def set_password(new_password=None, connection=None):   # pragma: no cover
+    connection = conn() if connection is None else connection
+    if new_password is None:
+        new_password = getpass()
```
eywalker: A prompt should be added here.

datajoint/settings.py
```diff
@@ -93,6 +93,18 @@ def __iter__(self):
...
+        """
+        saves the settings in the local config file  
+        """
+        self.save(LOCALCONFIG)
+
+    def save_global(self):
+        """
+        saves the settings in the global config file  
+        """
+        self.save(os.path.expanduser('~/' + GLOBALCONFIG))
```
eywalker: To be safe, I recommend that we use `path.join` from `os.path` package when combining paths.

datajoint/admin.py
dimitri-yatsenko: `getpass` shows a prompt.

datajoint/settings.py
dimitri-yatsenko: Does `expanduser` work for Windows?  Sorry i don't have a windows system.

datajoint/admin.py
eywalker: Ah ok cool, didn't know that it had default prompt.

datajoint/settings.py
eywalker: Yes, it will default to `HOMEPATH`, so at least it is consistently defined.

PR #295, opened 2017-03-03
Treat empty string value in numeric field as a nan

mspacek:
Otherwise, empty string values, such as those read from empty cells in a spreadsheet, have to be converted manually ahead of time to None or np.nan, which is a real pain.

### DIFF
--- a/datajoint/base_relation.py
+++ b/datajoint/base_relation.py
@@ -177,7 +177,7 @@ class BaseRelation(RelationalOperand):
                     value = pack(value)
                     placeholder = '%s'
                 elif heading[name].numeric:
-                    if value in [None, ''] or np.isnan(np.float(value)): # nans are turned into NULLs
+                    if value is None or value == '' or np.isnan(np.float(value)): # nans are turned into NULLs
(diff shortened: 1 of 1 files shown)


datajoint/base_relation.py
```diff
@@ -177,7 +177,7 @@ def make_placeholder(name, value):
                     value = pack(value)
                     placeholder = '%s'
                 elif heading[name].numeric:
-                    if value is None or np.isnan(np.float(value)):  # nans are turned into NULLs
+                    if value in [None, ''] or np.isnan(np.float(value)): # nans are turned into NULLs
```
dimitri-yatsenko: I agree with the change.  However, I feel that the condition should rather be
```python
value is None or value == '' or np.isnan(np.float(value))
```
and not rely on `__contains__` to work correctly with `None`.  It is probably more efficient too.

dimitri-yatsenko: actually, I am not sure.  It's just a feeling I had.  Something about always using `is None` rather than hashing `None`.

mspacek:
OK. How's this?

PR #296, opened 2017-03-06
Add limit and width kwargs to preview()

mspacek:
This allows the user to display more or less data, as desired, without having to change the defaults in the `dj_local_conf.json` file.


fabiansinz:
Alternatively you could use: 
```python
with dj.config(display__limit=50):
   # do things
```

eywalker:
For the preview usage, I agree with @mspacek that it doesn't hurt to expand with this feature. For other temporary config changes, use the `with dj.config` as @fabiansinz suggested.

PR #297, opened 2017-03-06
add `create_virtual_module`

fabiansinz:
This allows the user to create modules with datajoint classes on the fly. This is particularly useful if one python package wants to use data from another one without the need to install all the dependencies.

### DIFF
--- a/datajoint/__init__.py
+++ b/datajoint/__init__.py
@@ -81,7 +81,7 @@ from .admin import set_password, kill
 def create_virtual_module(modulename, dbname):
     """
     Creates a python module with the given name from a database name in mysql with datajoint tables.
-    Automatically creates the classed of the appropriate tier in the module.
+    Automatically creates the classes of the appropriate tier in the module.
 
     :param modulename: desired name of the module
     :param dbname:     name of the database in mysql
--- a/datajoint/__init__.py
+++ b/datajoint/__init__.py
@@ -14,6 +14,7 @@ Please cite:
 
 import logging
 import os
+from types import ModuleType
 from .version import __version__
 
 __author__ = "Dimitri Yatsenko, Edgar Y. Walker, and Fabian Sinz at Baylor College of Medicine"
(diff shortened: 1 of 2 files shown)


datajoint/__init__.py
```diff
@@ -75,3 +76,18 @@ class DataJointError(Exception):
...
+    Creates a python module with the given name from a database name in mysql with datajoint tables.
+    Automatically creates the classed of the appropriate tier in the module.
+
+    :param modulename: desired name of the module
+    :param dbname:     name of the database in mysql
+    :return: the python module
+    """
+    mod = ModuleType(modulename)
+    s = schema(dbname, mod.__dict__)
+    s.spawn_missing_classes()
```
eywalker: Perhaps add `schema` as a member item for module object, allowing the canonical `module.schema` access to the schema object.

eywalker: Also, would you mind moving this to `datajoint.utils` and importing that in `__init__` instead? `__init__` is starting to get really crowded :)

datajoint/__init__.py
```diff
@@ -75,3 +76,19 @@ class DataJointError(Exception):
 from .schema import Schema as schema
 from .erd import ERD
 from .admin import set_password, kill
+
+
+def create_virtual_module(modulename, dbname):
+    """
+    Creates a python module with the given name from a database name in mysql with datajoint tables.
+    Automatically creates the classed of the appropriate tier in the module.
```
eywalker: `classed` -> `classes`

datajoint/__init__.py
```diff
@@ -75,3 +76,19 @@ class DataJointError(Exception):
 from .schema import Schema as schema
 from .erd import ERD
 from .admin import set_password, kill
+
+
+def create_virtual_module(modulename, dbname):
+    """
+    Creates a python module with the given name from a database name in mysql with datajoint tables.
```
eywalker: `python` -> `Python`

Issue #303, opened 2017-03-17
Add `self` keyword to table definition


eywalker:
When you want to define a model reflected into a table (i.e. result of fitting model becomes an entry in a table) but want to apply it flexibly to various data sources with common structure but differing primary attributes, it would be very convenient if you can define a *template* table class for the Model that can work flexibly with different data sources. For this, something like below would be nice:

```python
class Model(dj.Manual):
    """
    Abstract base class defining basic logic for all models
    """
    data = None # should point to data source table
    
    @property
    def definition(self):
        return """
        -> Configs
        -> self.data
        ---
        score: float  # trained model score
        """
    
    def train(self, config, data):
        """
        Code to train a model based on configs and data.
        """
        ...
    
    def _make_tuples(self, key):
        data = (self.data() & key).fetch1()
        config = (Configs() & key).fetch1()
        key['score'] = self.train(config, data)
        self.insert1(key)
        
@schema
class Data1(dj.Manual):
    definition = """
    data_id: int          # data ID
    ---
    spikes: longblob      # spike data
    """
    
@schema
class Data2(dj.Manual):
    definition = """
    source_id: int        # source ID
    region: int           # region ID
    ---
    spikes: longblob      # spike data
    """

@schema
class ModelData1(Model):
    data = Data1
    
@schema
class ModelData2(Model):
    data = Data2
```

Key here is that in the table definition for the superclass `Model`, you can use keyword `self` to refer to the *current* class - that is the subclass that is getting decorated with the `schema`. This way, the `data` class member defined at the subclass can be used both for the table definition generation as well as in the `_make_tuples` method smoothly.

eywalker:
The ability to refer to `master` as a special keyword in the table definition for `dj.Part` table is now supported with PR #304. However, the inheritance pattern discussed above needs some more work, so I'll keep this issue open and use as a discussion board to sketch out this feature more in details.

eywalker:
You can use both `self` and `master` in the table definition to refer to the particular object instance (`self`) and the outer class object from the `Part` table (`master`), respectively.

Issue #302, opened 2017-03-17
Use keyword `master` in Part table's definition to refer to the Master table


eywalker:
Currently, you have to explicitly refer to the Master table's name in the Part table's definition as in 

```python
class A(dj.Computed):
    definition="""
    id: int    # unique id
    """
    
    class B(dj.Part):
        definition="""
        -> A
        ---
        value: float  # associated value
        """
```

What's the master for `B` is obvious from the context and having to explicitly state `-> A` feels redundant. I propose to add support for a keyword like `master` such that the above definition would look like this instead:

```python
class A(dj.Computed):
    definition="""
    id: int    # unique id
    """
    
    class B(dj.Part):
        definition="""
        -> master
        ---
        value: float  # associated value
        """
```

As I will detail in a separate issue, this also supports defining a mix-in class with Part table already defined (e.g. you can easily support a design in which additional part tables can be **added** by inheriting from a mix-in class).

dimitri-yatsenko:
I think this should be unified with #303 and used as `self.master`.

eywalker:
I much prefer `master` as I think `master` is simpler and it's clear what it means from the context. If we say `self.master`, then we have to add `master` attribute to the `Part` class instead of the `_master`.

dimitri-yatsenko:
then we should probably omit `self` altogether.  `-> data`  is sufficient.

dimitri-yatsenko:
what's the latest on this issue? Can we close this?

eywalker:
This has already been implemented and working well. Use of `self` is critical to distinguish the property of a class vs names defined in the local context. Closing.

PR #304, opened 2017-03-17
Support usage of special keywords in table definition

eywalker:
Address issue #302 and #303 
* `Part` tables can now use keyword `master` to refer to the `Master` table, and thus need not know the explicit name of the `Master` table
* You can now easily add new `Part` tables by inheriting from a mix-in class that defines the `Part` table thanks to the indirection of the `Master` table name
    * to support this, `schema` decorator looks for all available members for the class (including the inherited ones) to search for `Part` table
* support keyword `self` to refer to the table class that is being decorated. This allows for the definition to use the class attributes pointing to another table (refer to issue #303 for more details)
* added `force` keyword argument to `schema.drop` to allow for dropping of tables without being prompted. This is a quicker alternative to setting `safe_mode` config to `False` which as a side effect also affects policy on `delete` on Tables.
* added a test that uses all of the above aspects

### DIFF
--- a/datajoint/schema.py
+++ b/datajoint/schema.py
@@ -95,13 +95,14 @@ class Schema:
             self.process_relation_class(part_class, context=self.context, assert_declared=True)
             setattr(master_class, class_name, part_class)
 
-    def drop(self):
+    def drop(self, force=False):
         """
         Drop the associated database if it exists
         """
         if not self.exists:
             logger.info("Database named `{database}` does not exist. Doing nothing.".format(database=self.database))
         elif (not config['safemode'] or
+              force or
               user_choice("Proceed to delete entire schema `%s`?" % self.database, default='no') == 'yes'):
             logger.info("Dropping `{database}`.".format(database=self.database))
             try:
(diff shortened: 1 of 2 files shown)


datajoint/schema.py
```diff
@@ -151,15 +152,26 @@ def __call__(self, cls):
...
-        self.process_relation_class(cls, context=self.context)
+        ext = {
+            cls.__name__: cls,
+            'self': cls
+        }
+        self.process_relation_class(cls, context=dict(self.context, **ext))
+
         # Process part relations
-        for part in cls._ordered_class_members:
+        for part in dir(cls):
```
dimitri-yatsenko: the reason we used `_ordered_class_members` is that they are ordered the same way as in the definition.  Thus parts that are defined in a specific order are processed in the same order.  This is important because parts may refer to one another.  Using `dir` will break that.

eywalker: Very good point, I have now implemented an extended version of `dir` that respects order of declarations.

eywalker:
I have now implemented `ordered_dir` that lists all attributes, including the inherited ones but preserving the order of declarations where it matters. I have also included a test case that attempts to test for the need for order of declaration when using interdependent `Part` tables.

Issue #318, opened 2017-06-09
Simplify fetch


fabiansinz:
Currently, the fetch object is not elegant and potentially confusing to users. I propose the following changes to simplify it. 

**Stateless fetch**
`fetch` becomes completely stateless. This will simplify the implementation and make the process more transparent. All states become keyword arguments in `fetch`. For example `rel.fetch.order_by('attribute').as_dict()` becomes `rel.fetch(as_dict=True, order_by='attribute')`. All previous states become forced keyword arguments. 

**__getitem__ and iterators**
Fetch keeps `__getitem__`, `keys`, and `__iter__`. The disadvantage that comes with the stateless implementation is that the ways the elements are fetched cannot easily be modified for most of these functions. The following changes would solve this

- `keys` does accept arguments like `__call__` (i.e. `rel.keys(order_by='attr1')`)

- `__call__` does also accept unnamed arguments. In that case it behaves like `__getitem__` but with possible modifications. For example `a1, a2 = rel.fetch['attr1','attr2']` is equivalent to `a1, a2 = rel.fetch('attr1','attr2')`. However, `__call__` can be modified such as `a1, a2 = rel.fetch('attr1','attr2', order_by='attr1')`.

- `__iter__` stays, but cannot be modified. In addition, we provide the iterator `traverse` which accepts the same modifying keyword arguments as `__call__`. Additionally, `traverse` can get unnamed arguments. For example `for a1, a2 in rel.fetch.traverse('attr1','attr2', limit=10):`.

dimitri-yatsenko:
Yes, agreed.

dimitri-yatsenko:
If fetch is going to be stateless, there is little reason to keep it as a class.  It can become a method of `RelationalOperand`.  `traverse` then will be used rather than `fetch.traverse`.

fabiansinz:
I think we should keep it for two reasons:

- `__getitem__` needs an object and we might want to use getitem in the relation at some point (e.g. projections).

- it keeps all elements related to fetching separated from the relational algebra.

eywalker:
Yes let's not remove `fetch` object all of a sudden - we have introduced it precisely for the above two reasons that @fabiansinz pointed out. Making it stateless is a distinct consideration.

[...]

PR #319, opened 2017-06-10
Update `fetch` and `fetch1` syntax

eywalker:
Implemented changes discussed in #318, namely:

* Updated `fetch` and `fetch1` `__call__` syntax to take in unnamed arguments as attribute names 
* Deprecated the following properties/methods: `order_by`, `limit`, `offset`, `as_dict`, `squeeze`, `__getitem__`(`rel.fetch[]` and `rel.fetch1[]` syntax). Invoking these methods will now throw a deprecation warning
* Properly handle warning messages. Namely, remove capturing of `warnings` by `logging` as this is not something a library should do!
* Warnings issued by call to underlying `pymysql` are appropriately silenced
* Added tests for new `fetch` syntax while keeping tests for previous syntax until the old syntax is completely removed in a future release.

With this major feature addition (i.e. ability to fetch individual attributes with `__call__`), the version number has been incremented from `0.6.1` to `0.7.0` in accordance to semver.


datajoint/fetch.py
```diff
@@ -37,9 +38,15 @@ def _initialize_behavior(self):
...
+        DEPRECATED
+
         Changes the state of the fetch object to squeeze the returned values as much as possible.
         :return: a copy of the fetch object
         """
+
+        warnings.warn('Use of `squeeze` on `fetch` object is deprecated. Please use `squeeze=True` keyword arguments '
+                      'in the call to `fetch`/`keys` instead', stacklevel=2)
+
         ret = self.copy()
```
dimitri-yatsenko: Perhaps we should deprecate `FetchBase.copy` too.  It seems to be needed only for the old stateful Fetch.

datajoint/fetch.py
```diff
@@ -37,9 +38,15 @@ def _initialize_behavior(self):
     @property
     def squeeze(self):
         """
+        DEPRECATED
```
dimitri-yatsenko: Perhaps we could remove `FetchBase` and `Fetch1` and make `Fetch` do everything.  `Fetch` would then have the argument `exactly_one`.  The `fetch1` method could simply call `fetch(*args, exactly_one=True, **kwargs)`.  This could obviate the inheritance and the state handling.

datajoint/fetch.py
eywalker: Yes this will be deprecated once statefullness is completely retired.

datajoint/fetch.py
```diff
@@ -233,28 +285,44 @@ class Fetch1(FetchBase, Callable):
     :param relation: relation the fetch object fetches data from
     """
 
-    def __call__(self, **kwargs):
+    def __call__(self, *attrs, **kwargs):
```
eywalker: Yeah that would actually be cleaner - probably in next iteration.

datajoint/fetch.py
```diff
@@ -233,28 +285,44 @@ class Fetch1(FetchBase, Callable):
...
 
+        if len(attrs) == 0:  # fetch all attributes
+            cur = self._relation.cursor(as_dict=True)
+            ret = cur.fetchone()
+            if not ret or cur.fetchone():
+                raise DataJointError('fetch1 should only be used for relations with exactly one tuple')
+            ret = OrderedDict((name, unpack_(ret[name]) if heading[name].is_blob else ret[name])
+                               for name in heading.names)
+        else:
+            single_output = len(attrs) == 1
```
eywalker: Good point.

datajoint/fetch.py
eywalker: Let's think about it but let's not all of the sudden go about changing everything.

datajoint/fetch.py
dimitri-yatsenko: I am just saying that it should be marked DEPRECATED like the other deprecated things.

datajoint/fetch.py
```diff
@@ -93,71 +107,104 @@ def order_by(self, *args):
...
         :param as_dict: returns a list of dictionaries instead of a record array
         :return: the contents of the relation in the form of a structured numpy.array
         """
+        # if 'order_by' passed in a string, make into list
+        if isinstance(kwargs.get('order_by'), str):
+            kwargs['order_by'] = [kwargs['order_by']]
+
         sql_behavior = update_dict(self.sql_behavior, kwargs)
         ext_behavior = update_dict(self.ext_behavior, kwargs)
+        total_behavior = dict(sql_behavior)
```
eywalker: Yes - just note that we are still going to keep backward compatible behavior for a few rounds and hence I'm not drastically changing the internals yet.

datajoint/fetch.py
eywalker: This usage is only internal hence I didn't put explicit deprecate. I guess I could add that now.

PR #322, opened 2017-06-12
Fix bug in password change handling logic

eywalker:
* When passwords don't match, it still ended up updating the password.


datajoint/admin.py
```diff
@@ -11,6 +11,7 @@ def set_password(new_password=None, connection=None):   # pragma: no cover
         confirm_password = getpass('Confirm password: ')
         if new_password != confirm_password:
             print('Failed to confirm the password! Aborting password change.')
+            return
```
dimitri-yatsenko: oops.  sorry for not catching this.

dimitri-yatsenko: programming style-wise, an else clause is more appropriate here than `return` to keep the code more structural.

eywalker: No problem...sorry for that stupid bug to begin with.

PR #329, opened 2017-06-29
introduce limit in populate, fix bug in fetch

fabiansinz:
1. Sometimes it is useful to specify the maximal number of keys that should be populated. For instance, when running memory intensive jobs on a cluster so only one of two populations can run at a time, it would be nice to take turns. Setting `limit=1` allows one key to be populated instead of all keys to be exhausted before the other relation can be populated. 

2. Fixed the `to_dicts` function. Before, it returned numpy datatypes in the dictionary which was different from `rel.fetch.keys()` and could not be inserted back into a table. Now, the dictionaries have native python datatypes.

### DIFF
--- a/datajoint/autopopulate.py
+++ b/datajoint/autopopulate.py
@@ -5,6 +5,7 @@ import random
 from pymysql import OperationalError
 from .relational_operand import RelationalOperand, AndList
 from . import DataJointError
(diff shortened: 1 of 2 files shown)


fabiansinz:
Can somebody merge it please. I need the new datajoint image for population.

Issue #241, opened 2016-08-09
DataJoint schemas should continue to work when CREATE TABLE fails.


dimitri-yatsenko:
Currently, if a user has code defining new tables (or old tables that are no longer used), datajoint will attempt to create the missing tables.  When the user has no CREATE privileges, this operation throws an error and the user cannot use the schema. 

We need two new features:
- `dj.Schema` should offer a switch to suppress table creation.  
- If the user has insufficient privileges to create a table, the module should not fail but should throw a warning.

dimitri-yatsenko:
Example
```python
schema = dj.Schema('schema', locals(), create_tables=False)
```
`create_tables` is True by default.

dimitri-yatsenko:
a temporary solution is to always grant CREATE to collaborators but not INSERT, UPDATE, DROP or DELETE.

dimitri-yatsenko:
In our discussion yesterday, we decided that adding the extra option to `dj.Schema()` is unnecessary.  Rather, insufficient privileges should be handled gracefully with meaningful messages.  If a table fails to get created, the message will appear that and attempts to instantiate relation objects will fail with a meaningful message.  As a reminder, previously, the class declaration would fail aborting the import of entire modules.

dimitri-yatsenko:
Here is a test example that fails:
```python
# information from https://genome.ucsc.edu/goldenPath/help/mysql.html
dj.config['database.host'] = 'genome-mysql.soe.ucsc.edu'
dj.config['database.password'] = 'password'
dj.config['database.user'] = 'genomep'
mm10 = dj.create_virtual_module('mm10', 'mm10')
```

dimitri-yatsenko:
This was fixed a long time ago but we need to add a test.

PR #335, opened 2017-07-26
Fixed #334 and #241

### LINKED ISSUE #334: Handling of `Decimal` data type
Currently when you fetch from a table that contains `Decimal` datatype and then used the retrieved data dictionary to restrict the table back, this results in an error. For example,

```python
@schema
class A(dj.Lookup):
    definition = """
    value     : decimal(4,3)  # response latency in seconds.
    """

    contents = [[0.25]]
```

and if you do the following

```python
>>> r = A().fetch1(dj.key)
>>> A() & r
ProgrammingError: (1064, "You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'Decimal('0.250')) LIMIT 7' at line 1")
```
and the underlying SQL query is 

```mysql
SELECT `value` FROM `schema`.`#a` WHERE (`value`=Decimal('0.250')) LIMIT 7
```

Issue seems to be related to the fact that when fetching back `Decimal` datatype from SQL table, it is retrieved as `decimal.Decimal` which when used in query, renders as `Decimal('0.25')` instead of the desired `'0.25'`.

dimitri-yatsenko:
Fixed
* #334 -- handling of the Decimal datatype in restrictions
* #241 -- handling of insufficient privileges.  This was already fixed.  I just added tests.

### DIFF
--- a/tests/test_privileges.py
+++ b/tests/test_privileges.py
@@ -1,0 +1,8 @@
+from nose.tools import assert_false, assert_true, raises
+import datajoint as dj
+
+
+@raises(dj.DataJointError)
+def test_no_create_schema():
+    schema = dj.schema('forbidden_schema', locals())
+
--- a/tests/test_privileges.py
+++ b/tests/test_privileges.py
@@ -1,8 +1,58 @@
 from nose.tools import assert_false, assert_true, raises
 import datajoint as dj
+from os import environ
+from . import schema
+
+namespace = locals()
 
 
-@raises(dj.DataJointError)
-def test_no_create_schema():
-    schema = dj.schema('forbidden_schema', locals())
+class TestUnprivileged:
+
+    def __init__(self):
+        self.connection = None
+        self.previous_connection = None
+
+    def setUp(self):
+        # a connection with only SELECT privileges.  This user must be defined through mysql with only SELECT
+        # privileges to djtest schemas
+        view_connection = dict(
+            host=environ.get('DJ_TEST_HOST', 'localhost'),
+            user='djview',
+            password='djview')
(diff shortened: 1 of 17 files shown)


tests/test_schema.py
```diff
@@ -70,8 +68,8 @@ def test_unauthorized_database():
 
 
 def test_drop_database():
-    schema = dj.schema(PREFIX + '_drop_test', locals(), connection=dj.conn(**CONN_INFO))
+    schema = dj.schema(PREFIX + '_drop_test', locals(), connection=dj.conn(**CONN_INFO, reset=True))
```
eywalker: This statement leads to an error in 3.4 or lower due to placement of a keyword argument after dictionary expansion. Replace with `dj.conn(reset=True, **CONN_INFO)`

tests/test_privileges.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def setUp(self):
+        # a connection with only SELECT privileges.  This user must be defined through mysql with only SELECT
+        # privileges to djtest schemas
+        view_connection = dict(
+            host=environ.get('DJ_TEST_HOST', 'localhost'),
+            user='djview',
+            password='djview')
+        self.previous_connection = dj.conn.connection if hasattr(dj.conn, 'connection') else None
+        self.connection = dj.conn(**view_connection, reset=True)
```
eywalker: Same as below - won't work in Python 3.4 or below. Change with `dj.conn(reset=True, **view_connection)`.

eywalker:
Thanks for putting in this fix.

PR #356, opened 2017-08-15
fix for 'ignore_extra_fields'/insert select (datajoint-python#355)

ixcat:
Discussed with @dimitri-yatsenko -

Rather than getting deep into the rest of the query building mechanism, short-circuit this case and build a custom query from the fields in the destination table.

Table alias required by query is of the form ___<rel1>_IFR_<rel2> in order to reduce the likelihood of namespace collisions.

### DIFF
--- a/datajoint/base_relation.py
+++ b/datajoint/base_relation.py
@@ -142,8 +142,19 @@ class BaseRelation(RelationalOperand):
         """
 
         if isinstance(rows, RelationalOperand):
-            # INSERT FROM SELECT
-            query = 'INSERT{ignore} INTO {table} ({fields}) {select}'.format(
+            # INSERT FROM SELECT - build alternate field-narrowing query (only) when needed
+            if ignore_extra_fields and False in [name in self.heading.names for name in rows.heading.names]:
+                alias = '___' + self.table_name + '_IFS_' + rows.table_name
+                afields = [str('`'+alias+'`.`')+ x + '`' for x in self.heading.names]
+                query = 'INSERT{ignore} INTO {table} ({fields}) SELECT {afields} FROM ({select}) as `{alias}`'.format(
+                alias=alias,
+                ignore=" IGNORE" if ignore_errors or skip_duplicates else "",
+                table=self.full_table_name,
+                fields='`'+'`,`'.join(self.heading.names)+'`',
+                afields=','.join(afields),
+                select=rows.make_sql())
+            else:
+                query = 'INSERT{ignore} INTO {table} ({fields}) {select}'.format(
                 ignore=" IGNORE" if ignore_errors or skip_duplicates else "",
                 table=self.full_table_name,
                 fields='`'+'`,`'.join(rows.heading.names)+'`',
--- a/datajoint/base_relation.py
+++ b/datajoint/base_relation.py
@@ -143,7 +143,7 @@ class BaseRelation(RelationalOperand):
 
         if isinstance(rows, RelationalOperand):
             # INSERT FROM SELECT - build alternate field-narrowing query (only) when needed
-            if ignore_extra_fields and False in [name in self.heading.names for name in rows.heading.names]:
+            if ignore_extra_fields and not all(name in self.heading.names for name in rows.heading.names):
                 query = 'INSERT{ignore} INTO {table} ({fields}) SELECT {fields} FROM ({select}) as `__alias`'.format(
                 ignore=" IGNORE" if ignore_errors or skip_duplicates else "",
                 table=self.full_table_name,
--- a/datajoint/base_relation.py
+++ b/datajoint/base_relation.py
@@ -144,14 +144,10 @@ class BaseRelation(RelationalOperand):
         if isinstance(rows, RelationalOperand):
             # INSERT FROM SELECT - build alternate field-narrowing query (only) when needed
             if ignore_extra_fields and False in [name in self.heading.names for name in rows.heading.names]:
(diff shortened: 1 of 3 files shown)


tests/schema.py
```diff
@@ -22,6 +22,18 @@ class Test(dj.Lookup):
 
 
 @schema
+class TestExtra(dj.Manual):
+    ''' clone of Test but with an extra field '''
+    definition = Test.definition + "\nextra : int # extra int\n"
+
+
+@schema
+class TestNoExtra(dj.extra):
```
dimitri-yatsenko: what's `dj.extra`?

Nose tests did not pass because of this.

datajoint/base_relation.py
```diff
@@ -142,8 +142,19 @@ def insert(self, rows, replace=False, ignore_errors=False, skip_duplicates=False
         """
 
         if isinstance(rows, RelationalOperand):
-            # INSERT FROM SELECT
-            query = 'INSERT{ignore} INTO {table} ({fields}) {select}'.format(
+            # INSERT FROM SELECT - build alternate field-narrowing query (only) when needed
+            if ignore_extra_fields and False in [name in self.heading.names for name in rows.heading.names]:
+                alias = '___' + self.table_name + '_IFS_' + rows.table_name
+                afields = [str('`'+alias+'`.`')+ x + '`' for x in self.heading.names]
```
dimitri-yatsenko: It is unnecessary to prefix the fields with the table name.  All the attributes are from the outer query and are unique.

datajoint/base_relation.py
```diff
@@ -142,8 +142,19 @@ def insert(self, rows, replace=False, ignore_errors=False, skip_duplicates=False
         """
 
         if isinstance(rows, RelationalOperand):
-            # INSERT FROM SELECT
-            query = 'INSERT{ignore} INTO {table} ({fields}) {select}'.format(
+            # INSERT FROM SELECT - build alternate field-narrowing query (only) when needed
+            if ignore_extra_fields and False in [name in self.heading.names for name in rows.heading.names]:
+                alias = '___' + self.table_name + '_IFS_' + rows.table_name
```
dimitri-yatsenko: It is unnecessary to include the table name in the alias name.

tests/test_relation.py
```diff
@@ -98,6 +101,25 @@ def test_insert_select(self):
             'real_id', 'date_of_birth', 'subject_notes', subject_id='subject_id+1000', species='"human"'))
         assert_equal(len(self.subject), 2*original_length)
 
+    @raises(InternalError)
```
dimitri-yatsenko: perhaps this `InternalError` should be reraised as a DataJointError.

ixcat: did notice & agree this might be better - will need to xref against 'normal' queries without ignore_extra_fields as well since it doesn't look like this exception is caught in that case. Indeed other mysql exceptions are also thrown elsewhere so perhaps this should be a separate issue?

eywalker: Let's catch the `InernalError` `insert` method and re-raise as `DataJointError` - in general, we don't want to unnecessarily expose underlying library to the end user.

eywalker: Let's actually treat this as a separate issue since as @ixcat pointed out, re-raising as `DataJointError` is not properly done in many places and needs to be fixed.

datajoint/base_relation.py
ixcat: ah - true, good point. will apply 'fields' here directly instead.

datajoint/base_relation.py
ixcat: ok. Proposing '___alias' - corrected tests pass with this locally.

eywalker: Be aware, that `table_name` is *not* a property of a `RelationalOperand`. This line would only work if the `rows` happen to be a `BaseRelation` which is not a requirement for `select from` syntax.

tests/test_relation.py
```diff
@@ -98,6 +101,25 @@ def test_insert_select(self):
...
+    def test_insert_select_ignore_extra_fields0(self):
```
eywalker: Include a test case where the relation you are inserting from is not a simple table but rather a result of a query.

datajoint/base_relation.py
ixcat: thanks. updated code pending review has been corrected to simply use '___alias'

datajoint/base_relation.py
```diff
@@ -142,8 +142,15 @@ def insert(self, rows, replace=False, ignore_errors=False, skip_duplicates=False
         """
 
         if isinstance(rows, RelationalOperand):
-            # INSERT FROM SELECT
-            query = 'INSERT{ignore} INTO {table} ({fields}) {select}'.format(
+            # INSERT FROM SELECT - build alternate field-narrowing query (only) when needed
+            if ignore_extra_fields and False in [name in self.heading.names for name in rows.heading.names]:
```
dimitri-yatsenko: `... and not all(name in self.heading.names for name in rows.heading.names)`

ixcat: ah yes .. this is better. will incorporate.

tests/test_relation.py
ixcat: added in 36d2ba9

datajoint/base_relation.py
ixcat: added in 13385e4

PR #359, opened 2017-08-21
Fix for specific MySQL exception discussed in #35[567].

ixcat:
Recast the mysql InternalError to DataJointError in the specific case of extra columns given in insert-from-select queries.

### DIFF
--- a/datajoint/base_relation.py
+++ b/datajoint/base_relation.py
@@ -160,7 +160,7 @@ class BaseRelation(RelationalOperand):
             except pymysql.err.InternalError as err:
                 if err.args[0] == server_error_codes['unknown column']:
                     # args[1] -> Unknown column 'extra' in 'field list'
-                    raise DataJointError('%s : set ignore_extra_fields?' % err.args[1])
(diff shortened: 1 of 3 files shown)


datajoint/base_relation.py
```diff
@@ -155,7 +155,14 @@ def insert(self, rows, replace=False, ignore_errors=False, skip_duplicates=False
                 table=self.full_table_name,
                 fields='`'+'`,`'.join(rows.heading.names)+'`',
                 select=rows.make_sql())
-            self.connection.query(query)
+            try:
+                self.connection.query(query)
+            except pymysql.err.InternalError as err:
+                if err.args[0] == server_error_codes['unknown column']:
+                    # args[1] -> Unknown column 'extra' in 'field list'
+                    raise DataJointError('%s : set ignore_extra_fields?' % err.args[1])
```
dimitri-yatsenko: I would expand the suggestion into a full sentence, e.g. "To ignore extra fields, you may set the ignore_extra_fields argument in insert."

Issue #361, opened 2017-09-27
dj.ERD(schema).draw() throws networkx error


Visdoom:
Hey guys, I recently get errors when calling `dj.ERD(schema).draw()` 
Error Log:
```---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
/usr/local/lib/python3.5/dist-packages/networkx/classes/function.py in set_node_attributes(G, values, name)
    640         try:  # `values` is a dict
--> 641             for n, v in values.items():
    642                 try:

AttributeError: 'str' object has no attribute 'items'

During handling of the above exception, another exception occurred:

TypeError                                 Traceback (most recent call last)
<ipython-input-3-5dee5872e286> in <module>()
      1 fig = plt.figure(figsize=(30,30))
----> 2 dj.ERD(schema).draw()

/usr/local/lib/python3.5/dist-packages/datajoint/erd.py in draw(self)
    284 
    285         def draw(self):
--> 286             plt.imshow(self.make_image())
    287             plt.gca().axis('off')
    288             plt.show()

/usr/local/lib/python3.5/dist-packages/datajoint/erd.py in make_image(self)
    278 
    279         def make_image(self):
--> 280             return plt.imread(self.make_png())
    281 
    282         def _repr_svg_(self):

/usr/local/lib/python3.5/dist-packages/datajoint/erd.py in make_png(self)
    275 
    276         def make_png(self):
--> 277             return io.BytesIO(self.make_dot().create_png())
    278 
    279         def make_image(self):

/usr/local/lib/python3.5/dist-packages/datajoint/erd.py in make_dot(self)
    218             import networkx as nx
    219 
--> 220             graph = self._make_graph()
    221             graph.nodes()
    222 

/usr/local/lib/python3.5/dist-packages/datajoint/erd.py in _make_graph(self)
    206             # construct subgraph and rename nodes to class names
    207             graph = nx.DiGraph(self).subgraph(nodes)
--> 208             nx.set_node_attributes(graph, 'node_type', {n: _get_tier(n) for n in graph})
    209             # relabel nodes to class names
    210             mapping = {node: (lookup_class_name(node, self.context) or node) for node in graph.nodes()}

/usr/local/lib/python3.5/dist-packages/networkx/classes/function.py in set_node_attributes(G, values, name)
    646         except AttributeError:  # `values` is a constant
    647             for n in G:
--> 648                 G.nodes[n][name] = values
    649     else:  # `values` must be dict of dict
[...]

PR #364, opened 2017-09-27
requirements.txt: force networkx v1.1x

### LINKED ISSUE #363: force networkx 1.1x install until 2.0 support is implemented
As outlined in #362 , networkx 2.0 has been released and causes regressions.

Until networkx 2.0 support can be implemented, we should specify v1.1x in pip requirements to ensure current functionality.

ixcat:
To fix #363.


eywalker:
Would you also increment the patch version number in the `versions.py` file?

ixcat:
Also will resolve #361.

PR #366, opened 2017-09-29
Fix links in the README.md

eywalker:


ixcat:
ok (trivial https->http for sites w/o cert)

Issue #375, opened 2017-10-18
Define limit in populate as number of keys to compute rathers than keys to check


chbehrens:
At the moment if the function populate is called with the arguments `limit=n` and` reserve_jobs=True`, limit sets the number of keys that is checked against the job table. This means if populate is called on two instances with `populate(reserve_jobs=True, limit=1)`, the first instance selects the first key and populates it while the second instance selects the same key, detects that the job is already reserved and stops without populating an entry. The desired behavior from my point of view would be to select the first free key and populate that entry.
This is important when populating tables on a cluster where job runtimes have to be limited and estimated in advance so that one wants to limit the number of keys populated per job.

ixcat:
Any reason not to have both methods?

I could see controlling wasted number-of-failures as well as controlling number-of-computations:

    Table().populate(reserve_jobs=True, limit=5, max=100)

To compute up to 100 items but quitting if there are more than 5 instances of contention, indirectly indicating that there are 'enough' jobs running. That said this is probably not the best way to measure contention and something more robust would take more changes.

Any insight on how the current implementation came to be? Was there a use case or is the current behavior a bug vs the 'correct' meaning described by the ticket submission?

dimitri-yatsenko:
Currently, the limit argument is simply passed to the key source query.   This is efficient because a limited query may be faster than an unlimited one.  Limiting the number of jobs performed would require checking out more and then counting the ones actually computed.  As @ixcat suggested, perhaps we should have two different arguments.  `limit` should limit the `key_source` query as it does now and another argument should limit the number of jobs performed.  I propose the other argument to be called `max_computations`.  E.g.

eywalker:
I like the idea of keeping `limit` and adding a new argument for the new behavior, also to preserve any backward compatibility. Use of `limit` roughly ensure fixed number of key computed regardless of the number of workers. As @ixcat and @dimitri-yatsenko proposed, let's add new argument like `max_computed` to count up and limit how many keys were computed by each process.

dimitri-yatsenko:
since we use `limit` without describing what is limited, I think it's okay to use `max` rather than `max_computed` or `max_computations`.  The default values for both `limit` and `max` will be `None`.

eywalker:
I feel a bit uneasy about having `limit` and `max` as I feel that that can be confusing as to what does exactly what.

dimitri-yatsenko:
[...]

Issue #367, opened 2017-10-02
add optional progress report to populate


dimitri-yatsenko:
An option to populate to report progress and estimated completion time would be helpful.

Example:
```python
SpikeSort().populate(display_progress=True)
```

eywalker:
Let's perhaps use something like `tqdm` (https://pypi.python.org/pypi/tqdm) to achieve this.

dimitri-yatsenko:
Yes, that was my thinking.  It works great in command line and Jupyter.

Issue #328, opened 2017-06-28
Record the stack trace for errors in job table


ecobost:
Right now the job table has an attribute called `error_stack` but at least in python it is always None/null.

Cheers,

Issue #204, opened 2016-04-21
Support AWS S3 storage of blobs


dimitri-yatsenko:
Since S3 storage is less expensive than RDS storage, let's implement transparent storage of blobs in S3 buckets.  Blobs represent 99% of the overall data volume of a typical datajoint database. 

Let's discuss the implementation of this feature.

fabiansinz:
Why limit to S3? I vote for defining an interface that can possibly hook into many storage backends like simple hard-drives, Dropbox, sqlite ... . We don't necessarily need to implement it right now but we should program it such that it can easily be done. 

My idea is to translate blobs into general file identifiers. The problem there will be to come up with a way that works reliably across systems and programming languages. How that identifier is translated into the exact path depends on the backend which we set in the configuration. The question is: Do we want to implement it for all blobs at once or only for some blobs? In the former case we could use a new data type, in the latter case we can still go with blobs.

dimitri-yatsenko:
Should we add a setting to config to specify the storage of blobs and the path format?  We may set a size threshold above which the blob goes to storage or add a special 'stored_blob' datatype.

dimitri-yatsenko:
Specification:

A new attribute datatype will be added for externally stored data.  The datatype name will be `external`.

The config will specify `external_storage_type`, which will be one of ("file-system", "aws-s3").  This list may be extended in the future.

The config will specify `external_storage_path`, that might correspond to the root path for the file system or the bucket-name/folder-name for S3. 

Finally, config will specify `external_storage_filename` template which can use variables such as $TABLE_NAME, $SCHEMA_NAME, $KEY_VALUE, $KEY_HASH, $DATA_HASH.  For example, `config.external_storage_filename=$SCHEMA_NAME/$TABLE_NAME/$KEY_VALUE-$DATA_HASH.

$KEY_VALUE is the ascii serialized human-readable representation of the key values, in order.  E.g. 381-1-2

`insert` will exit with an error and will fail to insert if the saving of an external field failed.  

If a table has multiple external fields, then `insert` will also ensure that either all or none of the fields are stored.  If none are stored, the tuple is not inserted.

`insert` will throw an error if the external field already exists. Versioning will not be supported.

`delete` will attempt to delete all the external fields after the tuple is deleted.

peabody124:
Why not just have the external datatype be stored as a string with a uri?

fabiansinz:
@peabody124: In the end it will be a string in the database.

[...]

PR #377, opened 2017-10-20
External storage, enhancements, bugfixes

### LINKED ISSUE #264: Unions and Outer Joins
So far DataJoint has not offered the union operator and the outer join operators.  These two operators are closely related.  I propose to implement outer join/union as the `+` operator.

## Syntax
```python
a + b    # full outer join of relations `a` and `b` 
```
Incidentally, outer join is equivalent to union when the two arguments have the same sets of attributes.  

Note that MySQL does not implement full outer joins, so internally it will be implemented as a left join followed by a union with the right argument.

## Rules 
The full outer join follows the same rules as the join operator `a*b`:
1.  The primary key of the expression `a+b` is the union of the primary key attributes of `a` and `b`.
1.   The matching attributes in `a` and `b`, i.e. those sharing the same names in both relations must be of compatible datatypes and must belong to the primary key of either `a` or `b`.  An exception will be raised if a non-key attribute with the same name is present in both relations. 
1.  The order of attributes in the arguments does not matter and is not guaranteed in the result.

## Equivalence to OUTER JOIN
`a+b` is equivalent to the natural outer join operator in relational algebra.  

## Equivalence to SQL's UNION
`a+b` is equivalent to relational union if `a` and `b` have the same sets of attributes. 

## Implementation of LEFT and RIGHT OUTER JOINs
The natural left join (`a NATURAL LEFT JOIN b`) can now be expressed as
```python
a * b + a
```
or, equivalently, as
```python
(a + b) & a
```
or, equivalently, as
```python
a + (b & a) 
```
or, equivalently
```python
(a - b) + b
``` 

It is difficult to make DataJoint recognize all these patterns and to convert them into efficient SQL, so let's introduce the special syntax indicating left or right joins:

```python
a * ~b      # left outer join  (all rows of `a` are kept).
~a * b      # right outer join (all rows of `b` are kept). 
~a * ~b  # outer join (equivalent to `a+b`)
```

The outer flag `~` survives projections and restrictions but not other operators:
```python
~(a & cond) * b    # equivalent to   (~a & cond) * b
~a.proj() * b    # equivalent to   (~a).proj() * b
```

### Example 1:  Simple union
Let relation `a` with primary key (i, k) equal 

| *i 	| *k 	| 
|:-:	|:-:	|
|   1	|  1 	|
|   1	|  2	|
|   2	|  1 	|

Let relation `b` with primary key (k) equal 

| *k 	|  i 	| 
|:-:	|:-:	|
|   1	|   1	|
|   2	|   1	|
|   2	|  3	|

Then `a+b` will be

| *i 	| *k 	| 
|:-:	|:-:	|
|   1	|  1 	|
|   1	|  2	|
|   2	|  1 	|
|   3	|  2 	|

### Example 2: No matching attributes:  outer join = inner join

Let relation `a` with primary key (i) equal 

| *i 	| m | 
|:-:	|:-: |
|   1	|  6  |
|   2	|  6 	|

Let relation `b` with primary key (i, k) equal 

|  *k 	| n | 
|:-:	|:-: |
|   1	|   8 |
|   2	|   8 |

Then `a+b` will be

| *i 	| *k 	| m  | n  |
|:-:	|:-:	|:-:  |:-:  |
|   1	|   1	| 6 | 8 |
|   1	|   2	| 6 | 8 |
|   2	|   1	| 6 | 8 |
|   2 |    2  | 6 | 8 |

### Example 3:  Simple outer join

Let relation `a` with primary key (i) equal 

| *i 	| m | 
|:-:	|:-: |
|   1	|  6  |
|   2	|  6 	|
|   3	|  6 	|

Let relation `b` with primary key (i, k) equal 

| *i 	|  *k 	| n | 
|:-:	|:-:	|:-: |
|   1	|   1	| 8 |
|   1	|   2	| 8 |
|   2	|  3	| 8 |
|   4 |   1  | 8 |

Then `a+b` will be

| *i 	| *k 	| m  | n  |
|:-:	|:-:	|:-:  |:-:  |
|   1	|   1	| 6 | 8 |
|   1	|   2	| 6 | 8 |
|   2	|  3	| 6 | 8 |
|   4 |   1  | NULL | 8 |
|   3  |  NULL | 6 | NULL

Note `a+b` may produce NULLs in the primary key.

dimitri-yatsenko:
Here are the issues currently worked on.  The issues that are fixed or in progress will be included in this pull request.  The queued issues may be included.

* #204 - **fixed** for filesystem, S3 still remains for a future pull request
* #264 - **fixed** 
* #328 - **fixed**
* #367 - **fixed**
* #375 - **fixed**
* #381 - **fixed**
* #387 - **fixed**
* #388 - **fixed**
* #390 - **fixed**
* #391 - **fixed**


eywalker:
Can you make smaller piece-meal pull requests? It is rather difficult to thoroughly review large PR that tries to address multiple issues simultaneously.

ixcat:
Generally speaking, seems reasonable to me - 
I'll be doing more direct testing as part of the S3 feature (#204)

thoughts:
- If we are definitely leaving the external file cleanup as a separate action, it would be good to have a reference implementation of this, if it is even simple.
- I think it would be a good idea to abstract the low-level external storage functions (currently get/put) into a separate class hierarchy that is called from the main ```ExternalTable``` operations. This will help to clearly encapsulate what methods are required to create ```ExternalTable``` 'backends', and also could allow for reuse among backends (e.g. S3 and any other other 'WAN' backends could concievably use operations defined by the plain-file backend to manage the local cache). I'd probably go with an abstract or plain-file class as base, and have a ```{'backendname' : BackendClass}``` dict statically configured as a class property of ```ExternalFile``` at least initially, but that's me. 
- Unless I am mistaken, the blob is currently being written outside of any serialization of operations on the specific external record file; this could conceivably create a race condition in some (rare) corner cases, though I am admittedly not yet 100% up to date on DJ's locking mechanisms and how they might relate to say, 2 processes writing the same external blob when using populate in parallel on different keys which generate the same externally-stored result in a related table.. For posix the *data* should be coherent (though possibly with incorrect timestamps if a slower writer finished after the fast writer overwrote it's data and committed the timestamp of the now-replaced-file to the db), but am also not sure if this is well defined on other systems or if their semantics would match. Quite possibly paranoia, but it might be a good idea to create a temporary '<fname>-lock' and check for the existence of this and either stall or generate an error if it exists, or create some sort of similar locking protocol to track this on the database side.. I think in previous discussions I had assumed external file transactions would be serialized; I'm now not sure this is happening and so am mentioning it here.
- If the database insertion of the external file record into the database fails, we are left with a dangling blob. While this is a rare case and should be caught by the 'external cleanup' mechanism, it might be useful to handle cleanup here and then re-raise the error.

datajoint/external.py
```diff
@@ -0,0 +1,109 @@
...
+        put an object in external store
+        """
+        try:
+            spec = config[store]
+        except KeyError:
+            raise DataJointError('Storage {store} is not configured'.format(store=store))
+
+        # serialize object
+        blob = pack(obj)
+        hash = long_hash(blob) + store[len('external-'):]
```
ixcat: ```store[len('external-'):]``` seems incongruent w/r/t line 88: ```store = 'external' + ('-' if store else '') + store``` which could conceivably allow only ```external```; that said using only 'external' as a specifier is already blocked in  ```declare.py``` so I'm not sure how important this is.. perhaps line 88 should throw an error rather than allowing 'external' only?

dimitri-yatsenko: The idea is to allow either `external` or `external-somethin` where `somethin` cannot exceed 8 characters.  `external` by itself should not be blocked.  If it is, it's my mistake and I will fix it and add a test for it.  Many people will only have one external storage and they will just use `external`.

dimitri-yatsenko: Why is `store[len('external-'):]` incongruent?  You can do `a[n:]` where `n` exceeds the size of `a` no problem.

ixcat: Ah ok - I think on first read I was taking this to mean there was an explicit expectation that the slice would return something to append to the hash, which was then "confirmed" when I saw that declare.py requires the external-<something> convention (via the 'isidentifier' check on the secondary portion) and throws exceptions suggesting this is a requirement.

Not allowing the bare 'external' in some cases (here, declare.py) but allowing it in another (line 88) is why I proposed one was incongruent with respect to the other.

ixcat: if 'external' alone is is allowed, it looks like the store_name parts from declare (line 191+) need to be altered -
```
>>> storestr='external'
>>> storestr.split('-')
['external']
>>> store_name = storestr.split('-')
>>> store_name[1:]
[]
>>> ''.join(store_name[1:])
''
>>> ''.join(store_name[1:]).isidentifier()
False
```

datajoint/relational_operand.py
```diff
@@ -312,14 +354,24 @@ def restrict(self, restriction):
 
         :param restriction: a sequence or an array (treated as OR list), another relation, an SQL condition string, or
             an AndList.
-        """
-        if not restricts_to_same(restriction):
-            assert not self.heading.expressions or isinstance(self, GroupBy), \
-                "Cannot restrict in place a projection with renamed attributes."
-            if isinstance(restriction, AndList):
-                self.restrictions.extend(restriction)
+        :param as_or: if True than the applied restriction becomes or-listed with already existing conditions
```
eywalker: This is documented but not present - either implement it or update the doc.

datajoint/relational_operand.py
```diff
@@ -266,7 +308,7 @@ def __sub__(self, restriction):
         """
         return self & Not(restriction)
 
-    def restrict(self, restriction):
+    def restrict(self, arg):
```
eywalker: I liked it better to call it `restriction` rather than arbitrary `arg` as the parameter name.

tests/test_external.py
```diff
@@ -0,0 +1,26 @@
+import numpy as np
+from numpy.testing import assert_array_equal
+from nose.tools import assert_true, assert_equal
+from datajoint.external import ExternalTable
+
+from . schema_external import schema
+
+
+def test_external_put():
```
eywalker: need to add test case for pure `external` in addition to `external-xxx` formatted configs

tests/test_foreign_keys.py
```diff
@@ -1,5 +1,6 @@
 from nose.tools import assert_equal, assert_false, assert_true
 from datajoint.declare import declare
+from datajoint import AndList, OrList
```
eywalker: why importing `AndList` and `OrList` here? Appears to be unused.

datajoint/autopopulate.py
```diff
@@ -97,15 +114,20 @@ def handler(signum, frame):
...
         if order == "reverse":
             keys.reverse()
         elif order == "random":
             random.shuffle(keys)
 
+        call_count = count()
         logger.info('Found %d keys to populate' % len(keys))
-        for key in keys:
+
+        make = self._make_tuples if hasattr(self, '_make_tuples') else self.make
```
eywalker: We should throw deprecation warning when we find `_make_tuples`

eywalker: Concluded that we are holding off deprecation of `_make_tuples`

PR #394, opened 2017-11-16
Maintain column order in preview(), despite blob columns

mspacek:
I noticed that the preview for some tables does not always maintain the same column order as in the table definition. This only seems to happen for tables with blob columns. It looks like blobs are being inadvertently pushed to the end of the table preview. This simple PR fixes that. I've only tested `preview()`, not `_repr_html_()`. I'm not sure if the `rel.heading.attributes[c].comment` line in `_repr_html_()` should also be changed to `self.heading.attributes[c].comment`.


dimitri-yatsenko:
yes, changing the comment would be great too.

PR #398, opened 2017-11-20
Fix a bug introduced in #394 -- incorrect column order in HTML preview

dimitri-yatsenko:
There was a bug introduced in #394.  The column names were reordered by the body of the displayed table was not.  E.g.
image

In this example, the blob should be in `flip_times`. 

This fixes it.


ixcat:
Looks OK via visual check; minor change -> ack as OK.

PR #399, opened 2017-11-20
sped up the preview by simplifying its internal query

dimitri-yatsenko:


datajoint/relational_operand.py
```diff
@@ -409,19 +408,20 @@ def preview(self, limit=None, width=None):
         has_more = len(tuples) > limit
         tuples = tuples[:limit]
         columns = heading.names
-        widths = {f: min(max([len(f)] + [len(str(e)) for e in tuples[f]]) + 4, width) for f in columns}
+        widths = {f: min(max([len(f)] +
+            [len(str(e)) for e in tuples[f]] if f in tuples.dtype.names else [6]) + 4, width) for f in columns}
```
ixcat: what is the ```[6]``` magic number?

dimitri-yatsenko: len('=BLOB='), sorry

ixcat: can that be made explicit or perhaps commented?

PR #400, opened 2017-11-21
refactor the parsing of foreign keys

dimitri-yatsenko:
Improved messages, fixed a bug that arose in the handling of the following declaration, for example:

```python
@schema
class InputCell(dj.Manual):
    definition = """
    -> Cell
    (input)-> Cell(cell)
    """
```
This used to throw an error

### DIFF
--- a/datajoint/declare.py
+++ b/datajoint/declare.py
@@ -78,26 +78,53 @@ def compile_foreign_key(line, context, attributes, primary_key, attr_sql, foreig
     if not all(r in ref.primary_key for r in result.ref_attrs):
         raise DataJointError('Invalid foreign key attributes in "%s"' % line)
 
-    # match new attributes and referenced attributes and create foreign keys
-    missing_attrs = [attr for attr in ref.primary_key if attr not in attributes] or (
-        ref.primary_key if len(result.new_attrs) == len(ref.primary_key) == 1 else [])
-    new_attrs = result.new_attrs or missing_attrs
-    ref_attrs = result.ref_attrs or missing_attrs
+    try:
+        raise DataJointError('Duplicate attributes "{attr}" in "{line}"'.format(
+            attr=next(attr for attr in result.new_attrs if attr in attributes),
+            line=line))
+    except StopIteration:
+        pass   # the normal outcome
+
+    new_attrs = list(result.new_attrs) or []
+    ref_attrs = list(result.ref_attrs) or []
+
+    if new_attrs and not ref_attrs:
+        # special case, the renamed attribute is implicit
+        if len(new_attrs) != 1:
+            raise DataJointError('Renamed foreign key must be mapped to the primary key in "%s"' % line)
+        if len(ref.primary_key) == 1:
+            # if the primary key has one attribute, allow implicit renaming
+            ref_attrs = ref.primary_key
+        else:
+            # if only one primary key attribute is not yet used, then allow implicit renaming
+            ref_attrs = [attr for attr in ref.primary_key if attr not in attributes]
+            if len(ref_attrs) != 1:
+                raise DataJointError('Could not resovle which primary key attribute should be referenced in "%s"' % line)
+
     if len(new_attrs) != len(ref_attrs):
         raise DataJointError('Mismatched attributes in foreign key "%s"' % line)
 
-    # declare foreign key attributes and foreign keys
-    lookup = dict(zip(ref_attrs, new_attrs))
-    for attr in missing_attrs:
-        new_attr = lookup.get(attr, attr)
+    # expand the primary key of the referenced table
+    lookup = dict(zip(ref_attrs, new_attrs)).get  # from foreign to local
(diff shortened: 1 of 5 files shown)


datajoint/declare.py
```diff
@@ -78,26 +78,53 @@ def compile_foreign_key(line, context, attributes, primary_key, attr_sql, foreig
     if not all(r in ref.primary_key for r in result.ref_attrs):
         raise DataJointError('Invalid foreign key attributes in "%s"' % line)
 
-    # match new attributes and referenced attributes and create foreign keys
```
ixcat: was this comment useful?

datajoint/declare.py
```diff
@@ -78,26 +78,53 @@ def compile_foreign_key(line, context, attributes, primary_key, attr_sql, foreig
...
+
+    if new_attrs and not ref_attrs:
+        # special case, the renamed attribute is implicit
+        if len(new_attrs) != 1:
+            raise DataJointError('Renamed foreign key must be mapped to the primary key in "%s"' % line)
+        if len(ref.primary_key) == 1:
+            # if the primary key has one attribute, allow implicit renaming
+            ref_attrs = ref.primary_key
+        else:
+            # if only one primary key attribute is not yet used, then allow implicit renaming
```
ixcat: Should this case really be allowed?

It seems to encourage having attributes of the same name and also relies on sequence-of-definition which seems somewhat inconsistent with other properties of the DDL - though I perhaps could be misunderstanding the case/logic.

e.g.:

```
# 1st schema
@schema
class Foo(dj.Manual):
    definition = """
    fooid: integer
    barattr: integer
    """

@schema
class BarOne(dj.Manual):
    # is ok? (imports barattr as single attribute)
    definition = """
    fooid: integer
    -> Foo
    """

@schema
class BarTwo(dj.Manual):
    # is error: imports fooid and barattr; duplicate fooid results
    definition = """
    -> Foo
    fooid: integer
    """
```

ixcat: should this be allowed?
Seems to rely on sequence of definition in both source and destination table which is inconsistent with the rest of the DDL. (though I could be misunderstanding something)

dimitri-yatsenko: This comes in handy sometimes and reads naturally.  For example
```
-> Recording
(presynaptic) -> Cell
```
Here the cell belongs to a recording and already includes all the primary attributes of Recording plus the cell id.  This syntax is clear and useful.

datajoint/declare.py
dimitri-yatsenko: adding a few comments...

datajoint/declare.py
ixcat: thanks for the clarification -
my concern was not for legibility but for (perhaps excessive concern of) coherency w/r/t semantics of other syntax-

in other cases, we either import foreign key 'wholesale' (aka *import the whole thing, regardless of foreign attributes*) or import one specifically named attribute (aka *import specifically one thing, which I will name here*) - this case allows a partial import based on the *relative* presence of fields in one or the other, and so could allow non-explicit shifts in the imported references in some rare cases. For example, consider the following:

schema version 0:
```
@schema
class First:
    definition = """
    firstid: integer
    firstattr: integer
    """

@schema
class FirstRef:
    definition = """
    -> First
    firstsomething: integer
    """

@schema
class Second:
    definition = """
    -> First
    secondid: integer
    otherattr: integer
    (firstgizmo) -> FirstRef  # refers to firstsomething
    """
```

schema version 1:

```
@schema
class First:
    definition = """
    firstid: integer
    firstattr: integer
    """

@schema
class FirstRef:
    definition = """
    -> First
    otherattr: integer   # renamed firstsomething
    newattr: varchar(22)
    """

@schema
class Second:
    definition = """
    secondid: integer
    otherattr: integer
    -> FirstRef
    (firstgizmo) -> FirstRef   # now refers to newattr
    """
```

admittedly this would be an extremely rare case resulting from 2 specific things which are discouraged:

1) having attributes of duplicate names
2) modifying primary keys in a schema

but it is allowed in this mechanism and is somewhat different from other cases where attributes are potentially duplicated. We currently don't check for duplicates and don't enforce naming uniqueness, so this could fit under the same sort of 'permissive' rubric which allows 'name shifts' and 'name clashes' when importing the foreign key wholesale - just wanted to make sure it was known/acknowledged/discussed since this to me seems like the only case where we are 'computing' the referenced sub-attributes at definition time.

side note: have probably been thinking too much about strongly typed pure functional languages lately.. this comment could be related.. :D

ixcat: correction:

```
We currently don't check for duplicates and don't enforce naming uniqueness, so this could fit under the same sort of 'permissive' rubric which allows 'name shifts' and 'name clashes' when importing the foreign key wholesale
```

was the case previously (IIRC?), but is addressed in this commit

ixcat: it seems this would be caught in the new duplicate attribute check, so isn't actually a concern 👍

PR #402, opened 2017-11-22
erd.py: fix for #393 

### LINKED ISSUE #393: Evaluating an expression returning the schema module results in missing schema name in ERD plot
```
vmod = dj.create_virtual_module('vmod','schema_name')
vmod
dj.ERD(vmod)
```

results in ```_.Table``` in ERD rather than ```vmod.Table```

ixcat:
handled in erd vs base_relation.lookup_class_name.

fixes #393 

other option would be to modify 'lookup_class_name'; this seems less intrusive
and localized to the problem - other uses of lookup_class_name do not need/use
'module' component and skipping search in '_' there could potentially cause
undefined behavior in some corner cases.

### DIFF
--- a/datajoint/erd.py
+++ b/datajoint/erd.py
@@ -208,7 +208,11 @@ else:
             graph = nx.DiGraph(self).subgraph(nodes)
             nx.set_node_attributes(graph, 'node_type', {n: _get_tier(n) for n in graph})
             # relabel nodes to class names
-            mapping = {node: (lookup_class_name(node, self.context) or node) for node in graph.nodes()}
+            mapping = {node: (node) for node in graph.nodes()}
+            clean_context = dict((k, self.context[k]) for k in self.context
+                                 if '_' not in k)  # hack for ipython '_' var
+            mapping = {node: (lookup_class_name(node, clean_context) or node)
+                       for node in graph.nodes()}
             new_names = [mapping.values()]
             if len(new_names) > len(set(new_names)):
                 raise DataJointError('Some classes have identical names. The ERD cannot be plotted.')
--- a/datajoint/erd.py
+++ b/datajoint/erd.py
@@ -208,7 +208,6 @@ else:
             graph = nx.DiGraph(self).subgraph(nodes)
             nx.set_node_attributes(graph, 'node_type', {n: _get_tier(n) for n in graph})
             # relabel nodes to class names
-            mapping = {node: (node) for node in graph.nodes()}
             clean_context = dict((k, self.context[k]) for k in self.context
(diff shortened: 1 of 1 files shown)


datajoint/erd.py
```diff
@@ -208,7 +208,11 @@ def _make_graph(self):
             graph = nx.DiGraph(self).subgraph(nodes)
             nx.set_node_attributes(graph, 'node_type', {n: _get_tier(n) for n in graph})
             # relabel nodes to class names
-            mapping = {node: (lookup_class_name(node, self.context) or node) for node in graph.nodes()}
+            mapping = {node: (node) for node in graph.nodes()}
+            clean_context = dict((k, self.context[k]) for k in self.context
+                                 if '_' not in k)  # hack for ipython '_' var
```
ixcat: will acknowledge this could be a bit broad of a match even though we prefer *CamelCase* class names..

alternately could e.g.: ```if not re.match('^_*', k)``` or similar...

datajoint/erd.py
```diff
@@ -208,7 +208,11 @@ def _make_graph(self):
             graph = nx.DiGraph(self).subgraph(nodes)
             nx.set_node_attributes(graph, 'node_type', {n: _get_tier(n) for n in graph})
             # relabel nodes to class names
-            mapping = {node: (lookup_class_name(node, self.context) or node) for node in graph.nodes()}
+            mapping = {node: (node) for node in graph.nodes()}
```
dimitri-yatsenko: looks like this value of `mapping` is computed but not used.

ixcat: ah yes. looks like my temp code bled through. Yay Code Review!

datajoint/erd.py
```diff
@@ -208,7 +208,10 @@ def _make_graph(self):
             graph = nx.DiGraph(self).subgraph(nodes)
             nx.set_node_attributes(graph, 'node_type', {n: _get_tier(n) for n in graph})
             # relabel nodes to class names
-            mapping = {node: (lookup_class_name(node, self.context) or node) for node in graph.nodes()}
+            clean_context = dict((k, self.context[k]) for k in self.context
+                                 if '_' not in k)  # hack for ipython '_' var
```
dimitri-yatsenko: shouldn't this be 
`if k != '_'`  or `if not k.beginswith('_')`?

eywalker: Why are we rejecting everything containing `_`? I think this should be `k.beginswith('_')`

eywalker:
This is not the right fix - will be implementing the right fix in the next PR

PR #404, opened 2017-11-28
S3+filesystem caching 

ixcat:
Implementation of an S3 'external' (implements #204) along with a external-blob caching mixin & class definitions for S3 and native to use it.

Refactored the external data handling backend into an OO heirarchy for modularity
as part of the implementation. One nit about the current approach is that the individual 'handler' class instances are created one-per-request but could conceivably be cached or made singletons; I didn't want to put time into implementing this just yet since I wanted to get feedback on the overall redesign approach.

Some of the config directives were tweaked against the 'proposed' versions from the earlier external work; I expect there could be some discussion here on the direction taken. Notably, the "requirement" of calling external things 'external-' persists in that the (global) cache configuration must now be now called 'external-cache'. The other main cfg change is the renaming of user/pass to AWS terms for the S3 class.

Unit test state matches against current 'external' tests and are provided 2x: the first with 'mock' S3 'moto' implementation and the second using environment variables to configure the real S3 - however for some reason these don't like to coexist in the same test run at the moment. (Suspecting 'moto' breaks the in-memory boto S3 library or the config setting mechanism used in the live test results in a stale reference, etc). The test suite as a whole (e.g. run with no arguments) uses the 'moto' mock test and skips the 'real' test if AWS credential environment variables are not set and the real test succeeds when run directly - not sure if this should be a blocker or not, thoughts welcome there - I can see arguments to both sides. Either way further CI tweaking is needed to store/pass credentials to the tests before the CI test can be implemented automatically against the real S3, so I expect some further work w/r/t the unit tests.


dimitri-yatsenko:
Will you please disable the tests that cannot be tested on Travis so that at least the others pass. We can address the testability of S3 functionality as we go.

datajoint/external.py
```diff
@@ -6,6 +6,203 @@
 from .declare import STORE_HASH_LENGTH, HASH_DATA_TYPE
 
 
+class ExternalFileHandler:
```
dimitri-yatsenko: Let's rename this class `ExternalStore`

datajoint/external.py
```diff
@@ -6,6 +6,203 @@
...
+
+    def __init__(self, store, database):
+        self._store = store
+        self._database = database
+        self._spec = config[store]
+
+    @classmethod
+    def get_handler(cls, store, database):
+
+        if store not in config:
```
dimitri-yatsenko: Would catching the `KeyError` be a more efficient approach than explicit membership check?

datajoint/external.py
```diff
@@ -6,6 +6,203 @@
...
+            raise DataJointError(
+                'Store "{store}" is not configured'.format(store=store))
+
+        spec = config[store]
+
+        if 'protocol' not in spec:
```
dimitri-yatsenko: Let's catch the `KeyError` instead of explicit membership checks

datajoint/external.py
```diff
@@ -6,6 +6,203 @@
...
+                'Unknown external storage protocol "{protocol}" for "{store}"'
+                .format(store=store, protocol=protocol))
+
+        return cls._handlers[protocol](store, database)
+
+    def check_required(self, store, storetype, required):
+
+        missing = list(i for i in required if i not in self._spec)
+
+        if len(missing):
```
dimitri-yatsenko: This is equivalent to `if missing:`

datajoint/external.py
```diff
@@ -6,6 +6,203 @@
...
+        ''' returns decoded 'obj' '''
+        raise NotImplementedError(
+            'get_obj method not implemented for', type(self))
+
+    def get(self, hash):
+        ''' returns decoded 'obj' '''
+        return self.get_obj(hash)
+
+
+class RawFileHandler(ExternalFileHandler):
```
dimitri-yatsenko: What's a "raw file"?  I would call this class "FileStore".

datajoint/external.py
```diff
@@ -6,6 +6,203 @@
...
+
+    required = ('location',)
+
+    def __init__(self, store, database):
+        super().__init__(store, database)
+        self.check_required(store, 's3', RawFileHandler.required)
```
dimitri-yatsenko: Why is this storetype='s3'?

datajoint/external.py
```diff
@@ -6,6 +6,203 @@
...
+        protocol = spec['protocol']
+
+        if protocol not in cls._handlers:
+            raise DataJointError(
+                'Unknown external storage protocol "{protocol}" for "{store}"'
+                .format(store=store, protocol=protocol))
+
+        return cls._handlers[protocol](store, database)
+
+    def check_required(self, store, storetype, required):
```
dimitri-yatsenko: `storetype` does not seem to be used for anything.  If you are using specialized classes, then the class itself represents the store type.

datajoint/external.py
```diff
@@ -6,6 +6,203 @@
...
+                raise DataJointError('Lost external blob')
+
+    def get_obj(self, hash):
+        return unpack(self.get_blob(hash))
+
+    def get(self, hash):
+        return self.get_obj(hash)
+
+
+class CacheFileHandler(ExternalFileHandler):
```
dimitri-yatsenko: I don't think cache is its own store type.  Rather, it's a feature in all other store types.

datajoint/external.py
```diff
@@ -6,6 +6,203 @@
...
+    superclass for coherent get/put operations on the 'reference' blobs.
+
+    Cleanup currently not implemented.
+    '''
+
+    def __init__(self, store, database):
+        super().__init__(store, database)  # initialize non-mixin parameters
+
+        # validate mixin cache parameters
+        if 'external-cache' not in config:
```
dimitri-yatsenko: Let's catch the `KeyError` rather explicitly check.

datajoint/external.py
```diff
@@ -6,6 +6,203 @@
...
+            raise DataJointError('External Cache is not configured')
+
+        cache_spec = config['external-cache']
+
+        if 'location' not in cache_spec:
```
dimitri-yatsenko: Let's catch the `KeyError` rather explicitly check

datajoint/external.py
```diff
@@ -6,6 +6,203 @@
...
+        if not os.path.isfile(full_path):
+            try:
+                with open(full_path, 'wb') as f:
+                    f.write(blob)
+            except FileNotFoundError:
+                os.makedirs(folder)
+                with open(full_path, 'wb') as f:
+                    f.write(blob)
+        else:
+            pass  # TODO: track recent file usage to assist _clean_cache
```
dimitri-yatsenko: Incomplete TODO.

Let's not leave TODOs in code unless it's work in progress not ready for a pull request.  If you plan to implement a feature, propose it in issues where we can discuss it.

dimitri-yatsenko:
1.  The caching should not require a separate class.  It can be just an option in the configuration and all the functionality should go into the base class.  This will remove the need for the mixin class and the many cache- variant of every store type.

2.  I think the `config['external-cache]` is confusing because it looks like another store option.

dimitri-yatsenko:
3. I am not too excited about sneaking in new handlers in `s3.py`.  It would be great to have a single place where all protocols are listed.  I would be in favor of moving out all the store implementations into separate modules and importing them all into `external.py` to be listed in a single place.

datajoint/external.py
```diff
@@ -6,6 +6,203 @@
...
+
+        if len(missing):
+            raise DataJointError(
+                'Store "{s}" incorrectly configured for "{t}", missing: {m}'
+                .format(s=store, t=storetype, m=missing))
+
+    def hash_obj(self, obj):
+        blob = pack(obj)
+        hash = long_hash(blob) + self._store[len('external-'):]
+        return (blob, hash)
```
dimitri-yatsenko: `return blob, hash`

datajoint/external.py
```diff
@@ -6,6 +6,203 @@
...
+    def __init__(self, store, database):
+        super().__init__(store, database)
+        self.check_required(store, 's3', RawFileHandler.required)
+        self._location = self._spec['location']
+
+    def get_folder(self):
+        return os.path.join(self._location, self._database)
+
+    def put(self, obj):
+        (blob, hash) = self.hash_obj(obj)
```
dimitri-yatsenko: `blob, hash = self.hash_obj(obj)`

datajoint/external.py
```diff
@@ -6,6 +6,203 @@
...
+        if not os.path.isfile(full_path):
+            try:
+                with open(full_path, 'wb') as f:
+                    f.write(blob)
+            except FileNotFoundError:
+                os.makedirs(folder)
+                with open(full_path, 'wb') as f:
+                    f.write(blob)
+
+        return (blob, hash)
```
dimitri-yatsenko: `return blob, hash`

datajoint/external.py
```diff
@@ -42,68 +239,20 @@ def put(self, store, obj):
...
-                    with open(full_path, 'wb') as f:
-                        f.write(blob)
-                except FileNotFoundError:
-                    os.makedirs(folder)
-                    with open(full_path, 'wb') as f:
-                        f.write(blob)
-        else:
-            raise DataJointError('Unknown external storage protocol {protocol} for {store}'.format(
-                store=store, protocol=protocol))
+        (blob, hash) = ExternalFileHandler.get_handler(
```
dimitri-yatsenko: `blob, hash = ...`

datajoint/external.py
```diff
@@ -6,6 +6,203 @@
...
+
+    def get_cache_folder(self):
+        return os.path.join(self._cache_location, self._database)
+
+    def _clean_cache(self):
+        pass  # TODO: implement _clean_cache
+
+    def _put_cache(self, blob, hash):
+
+        self._clean_cache()
```
dimitri-yatsenko: why clean cache at every put?

datajoint/external.py
```diff
@@ -6,6 +6,203 @@
...
+                'External Cache configuration missing "location"')
+
+        self._cache_spec = cache_spec
+        self._cache_location = cache_spec['location']
+
+    def get_cache_folder(self):
+        return os.path.join(self._cache_location, self._database)
+
+    def _clean_cache(self):
+        pass  # TODO: implement _clean_cache
```
dimitri-yatsenko: incomplete TODO.

Please use TODOs for work in progress only.  If you propose a new feature, please use the issue tracker.

datajoint/external.py
```diff
@@ -6,6 +6,203 @@
...
+        else:
+            pass  # TODO: track recent file usage to assist _clean_cache
+
+        return (blob, hash)
+
+    def put(self, obj):
+        return self._put_cache(*super().put(obj))
+
+    def get_blob(self, hash):
+        blob = super().get_blob(hash)
```
dimitri-yatsenko: The object is not retrieved from the cache.  What's the point of the cache then?

datajoint/s3.py
```diff
@@ -150,3 +189,64 @@ def delete(self, rpath):
...
+            return self._location + '/' + hash
+
+    def put(self, obj):
+        (blob, hash) = self.hash_obj(obj)
+        rpath = self.make_path(hash)
+        self._bucket.put(BytesIO(blob), rpath)
+        return (blob, hash)
+
+    def get_blob(self, hash):
+        rpath = self.make_path(hash)
```
dimitri-yatsenko: why `make_path` in `get_blob`?  Shouldn't you return an error if it's not there?

datajoint/s3.py
```diff
@@ -150,3 +189,64 @@ def delete(self, rpath):
...
+        '''
+        if(len(self._location) == 0 or self._location == '/'):
+            return hash
+        if self._location[0] == '/':
+            return self._location[1:] + '/' + hash
+        else:
+            return self._location + '/' + hash
+
+    def put(self, obj):
+        (blob, hash) = self.hash_obj(obj)
```
dimitri-yatsenko: `blob, hash = ...`

datajoint/s3.py
```diff
@@ -150,3 +189,64 @@ def delete(self, rpath):
...
+
+        S3 has no path.join; '' is root, '/' is delimiter.
+
+        Leading slashes will create an empty-named top-level folder.
+        therefore we leading slashes from location and do not create them.
+
+        This is done here and not in the Bucket class since it is not
+        technically 'illegal' and Bucket encapsulates low-level access.
+        '''
+        if(len(self._location) == 0 or self._location == '/'):
```
dimitri-yatsenko:
```python
if not self._location or self._location == '/':
```

datajoint/s3.py
```diff
@@ -150,3 +189,64 @@ def delete(self, rpath):
...
+            return hash
+        if self._location[0] == '/':
```
dimitri-yatsenko:
```python
return self._location.lstrip('/') + hash
```

datajoint/s3.py
```diff
@@ -150,3 +189,64 @@ def delete(self, rpath):
...
+            return self._location[1:] + '/' + hash
+        else:
+            return self._location + '/' + hash
+
+    def put(self, obj):
+        (blob, hash) = self.hash_obj(obj)
+        rpath = self.make_path(hash)
+        self._bucket.put(BytesIO(blob), rpath)
+        return (blob, hash)
```
dimitri-yatsenko: return blob, hash

datajoint/s3.py
```diff
@@ -115,17 +134,37 @@ def put(self, lpath=None, rpath=None):
...
+        Retrieve a file from the bucket into a 'bytes-like-object'
+
+        :param rpath: remote path within bucket
+        :param obj: local object
+        """
+        try:
+            self.connect()
+            self._s3.Object(self._bucket, rpath).download_fileobj(obj)
+        except Exception as e:
+            # XXX: risk of concatenating huge object? hmm.
```
dimitri-yatsenko: make comment meaningful.

datajoint/s3.py
```diff
@@ -1,51 +1,25 @@
...
+logger = logging.getLogger(__name__)
 
-        if aws_secret_access_key is None:  # pragma: no cover
-            aws_secret_access_key = getpass(
-                "Please enter AWS Secret Access Key: "
-            )
 
-        bucket.bucket = Bucket(aws_access_key_id, aws_secret_access_key)
-    return bucket.bucket
+def bucket(**kwargs):
```
dimitri-yatsenko:
```python
bucket = Bucket.get_bucket    # alias for the Bucket factor function
```

datajoint/s3.py
```diff
@@ -150,3 +189,64 @@ def delete(self, rpath):
...
+        self._location = self._spec['location']
+
+    def make_path(self, hash):
+        '''
+        create a remote s3 path for the given hash.
+
+        S3 has no path.join; '' is root, '/' is delimiter.
+
+        Leading slashes will create an empty-named top-level folder.
+        therefore we leading slashes from location and do not create them.
```
dimitri-yatsenko: ... we *strip* leading slashes ...

datajoint/external.py
dimitri-yatsenko: Or we could do cache in the `ExternalTable.put`/`get` and avoid problems with inheritance.

datajoint/external.py
ixcat: fundamentally disagree for the case of 'verifying key presence' especially when this is the part of the code directly responsible for validating configuration (aka verifying key presence) or membership checks are straightforward and dont require complex (exception causing error prone) logic 

a) 4 lines of code vs 2
b) exponentially? more places where bad indentation can bite you (especially in nested try blocks) 
d) is (try <> except KeyError <>) really more legible than 'if something not in container a else b'?

my take on exceptions is that they are just that - exceptional case to be handled via an alternate control structure not expressed clearly inline in the nominal path of the code.

also: this is not the same thing as running 50 duck-type tests for the same parameters on each stack frame pushed.

datajoint/external.py
ixcat: because the logic to clean the cache will be complex and function-worthy -
the thinking was clean cache itself can determine if it needs to actually clean anything
(yes this is a frame - but the alternative is having multiple functions tracking cache cleaning state)

datajoint/external.py
ixcat: I agree generally -
However, here, planning ahead for known-immediately-pending features is useful to prevent poor structure which will force subsequent rewriting; annotating and saving this effort is similarly useful.

PR #405, opened 2017-12-01
Revamp of restriction

### LINKED ISSUE #406: add a string equivalent of `dj.key` in fetch
Currently, `dj.key` is used to fetch the primary key.  However, this requires importing datajoint.  I propose to add a string equivalent of `dj.key`.  For example, this could be '<key>'.

Before:
```python
import datajoint as dj
import analysis
alphas, keys = analysis.Hyperparam().fetch('alpha', dj.key)
```
After:
```python
import analysis
alphas, keys = analysis.Hyperparam().fetch('alpha', '<key>')
```

dimitri-yatsenko:
Simplify and correct the logic of restrictions.

This PR includes #403 and should be merged after #403 

Also, fix #406:  `"KEY"` can be used in place of `dj.key`.

### DIFF
--- a/datajoint/__init__.py
+++ b/datajoint/__init__.py
@@ -69,7 +69,7 @@ from .connection import conn, Connection
 from .s3 import bucket, Bucket
 from .base_relation import FreeRelation, BaseRelation
 from .user_relations import Manual, Lookup, Imported, Computed, Part
-from .relational_operand import Not, AndList, OrList, U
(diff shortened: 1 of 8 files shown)


eywalker:
Please pull and merge the upstream branch to update the state of this PR.

eywalker:
This is a much cleaner implementation - awesome.

PR #407, opened 2017-12-04
Class methods

dimitri-yatsenko:
Allow to treat any DataJoint class like an instance of the class.
Instantiating is not longer required.

The following operations are now allowed:
```python
experiment.Session & 'animal_id=13077'
experiment.Scan.insert1(rec)
```

This PR includes #405 and should not be merged until #405 is.

This PR also implements local caching for external storage.


tests/schema_external.py
```diff
@@ -24,9 +25,10 @@
     'user': 'djtest',
     'token': '2e05709792545ce'}
 
-dj.config['cache'] = {
-    'protocol': 'file',
-    'location': '/media/dimitri/ExtraDrive1/dj-store/cache'}
+cache_path = 'dj-cache'
```
eywalker: I prefer that we would use a [temp directory](https://docs.python.org/3.6/library/tempfile.html#tempfile.mkdtemp) for the external and cache testing purpose rather than running the risk of dumping stuff into a  potentially meaningful directory during the test .

tests/schema_external.py
```diff
@@ -24,9 +25,10 @@
...
+if not os.path.exists(cache_path):
+    os.makedirs(cache_path)
+dj.config['cache'] = cache_path
```
eywalker: Do we have explicit test cases for external store with and without using the caching? Based on this, it appears that we now only test case with caching. I think it'll make sense to have explicit tests for external storage with and without storage by switching cases.

datajoint/external.py
```diff
@@ -87,24 +100,38 @@ def get(self, hash):
         """
         store = hash[STORE_HASH_LENGTH:]
         store = 'external' + ('-' if store else '') + store
-        try:
-            spec = config[store]
-        except KeyError:
-            raise DataJointError('Store `%s` is not configured' % store)
+        cache_file = os.path.join(config['cache'], hash) if 'cache' in config else None
```
eywalker: I also think that we should skip caching if `config['cache'] = None` (or any falsy values e.g. `False`), so change this to `... if 'cache' in config and config['cache'] else None`

datajoint/external.py
```diff
@@ -87,24 +100,38 @@ def get(self, hash):
         """
         store = hash[STORE_HASH_LENGTH:]
```
eywalker: Would you do me a favor and replace the variable name `hash` with something else? `hash` is the name of a built-in Python function and I much rather not have it's name used as variable.

tests/schema_external.py
dimitri-yatsenko: agreed

tests/schema_external.py
dimitri-yatsenko: Why is this better?

datajoint/external.py
dimitri-yatsenko: agreed

datajoint/external.py
dimitri-yatsenko: `hash_`? `hashed`? `blob_hash`?

tests/schema_external.py
eywalker: this will create a local directory called `dj-cache` or even put new content into the directory should it already exist. It's better to not have an unnecessary side effects during testing.

dimitri-yatsenko: got it

dimitri-yatsenko: It's quite tricky to use these tempfile objects in testing.  They clean up when the object is destroyed.  I am leaving the test as is for now but we should revise our general approach to handling temporary files used during testing.

tests/schema_external.py
```diff
@@ -24,9 +25,8 @@
     'user': 'djtest',
     'token': '2e05709792545ce'}
 
-dj.config['cache'] = {
-    'protocol': 'file',
-    'location': '/media/dimitri/ExtraDrive1/dj-store/cache'}
+dj.config['cache'] = 'dj-cache'
```
eywalker: why not just do `dj.config['cache'] = tempfile.mkdtemp('dj-cache')`

datajoint/external.py
eywalker: Too late now, but probably `blob_hash`

PR #410, opened 2017-12-12
eliminate unnecessary copy of external file

dimitri-yatsenko:
The earlier use of filecopy was a mistake.  Should have always been os.rename

### DIFF
--- a/datajoint/external.py
+++ b/datajoint/external.py
@@ -1,5 +1,4 @@
 import os
-import shutil
 from . import config, DataJointError
 from .hash import long_hash
 from .blob import pack, unpack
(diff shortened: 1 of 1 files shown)


datajoint/external.py
```diff
@@ -17,7 +16,7 @@ def safe_write(filename, blob):
     temp_file = filename + '.saving'
     with open(temp_file, 'bw') as f:
         f.write(blob)
-    shutil.copyfile(temp_file, filename)
+    os.rename(temp_file, filename)
```
eywalker: I should have caught this.

PR #411, opened 2017-12-13
[WIP] implement garbage collection for external storage

dimitri-yatsenko:


datajoint/s3.py
```diff
@@ -42,6 +42,12 @@ def get(self):
...
-        return r['ResponseMetadata']['HTTPStatusCode'] == 204
+
+def delete_all_except(except_list, bucket, aws_access_key_id, aws_secret_access_key, location, database, **_):
+    cred = (aws_access_key_id, aws_secret_access_key)
+    if cred not in sessions:
+        # cache sessions
+        session = boto3.Session(aws_access_key_id=aws_access_key_id,
+                                aws_secret_access_key=aws_secret_access_key)
+        sessions[cred] = session.resource('s3')
+    raise NotImplementedError
```
ixcat: via http://boto3.readthedocs.io/en/latest/guide/migrations3.html, raw boto delete example is:
```
# Boto 3
for key in bucket.objects.all():
    key.delete()
```
so basically this + checking against the file list should do it.
didn't fork/implement just yet since there is an open question about refactoring s3.py to zap the dj Bucket class & not 100% how this will map into this.

datajoint/external.py
```diff
@@ -113,29 +90,93 @@ def get(self, blob_hash):
...
+    def drop_quick(self):
+        """drop the external table -- works only when it's empty"""
+        if self:
+            raise DataJointError('Cannot drop a non-empty external table. Please use delete_garabge to clear it.')
+        self.drop_quick()
+
+    def delete_garbage(self):
+        """
+        Delete items that are no longer referenced.
+        This operation is safe to perform at any time.
```
ixcat: Correct me if wrong, but isn't ExternalTable.put called within base relation outside of a transaction in some cases? If so, there conceivably is a minuscule chance of something being added to the backing store (file, s3, etc) and external tracking table before it is referenced in the concrete 'real' relation table, which if deleted here would cause the subsequent update to the user relation to fail; concievably this would leave the backing file 'orphaned' and therefore deletable by clean_store before the issue could be caught and corrected.

not sure if this and the similar race you allude to in the case of clean_store are worth complicating the code with, especially since we are envisioning cache cleaning as a manual maintenance sort of effort and the chances of issues here are very low, but possibly something to note; If i'm correct, this operation is somewhat as risky as clean_store (with the difference that it is only eference rather than real data being purged in the case of clean_store)

it might make sense to flip the external logic to insert the record and then save the blob (rather than save-then-insert ecord) - while this wouldn't prevent this issue as-is, it would allow it to be prevented if we were to lock the entire database during cleanup - writers would then block on the 1st hash insertion and only a 'coherent' database would be cleaned; the other case of failed blob writing would trigger an exception so the external record would remain unreferenced in the main table (and therefore also cleanable)

again, possibly too much given the slim chance; simplest is to document this needs to be done carefully.

dimitri-yatsenko:
Combined into #417

PR #417, opened 2017-12-21
Local context; garbage collection

### LINKED ISSUE #368: Problematic behavior of `locals()` inside a funtion
Use of `locals()` inside a function leads to unexpected behavior. For an example, the following code snippet

```python
def make_tables():
    schema = dj.schema('test', locals())
    
    @schema
    class A():
        definition = """
        id:  int     # id
        """
    @schema
    class B():
        definition = """
         -> A
         """
```
Calling the fucntion `make_tables()` will result in name resolution error for the foreign key reference `-> A` inside the definition of `B` despite the fact that the following works just fine:

```python
schema = dj.schema('test', locals())
    
@schema
class A():
    definition = """
    id:  int     # id
    """
@schema
class B():
    definition = """
     -> A
     """
```

The problem arise due to the fact that, for some reasons, the dictionary returned by the `locals()` do *not* stay updated when used inside a function. This can be demonstrated by the following code:

```python
def test_locals():
    names = locals()
    print(names.keys())   # will print out []
    # add new variables
    a = 10
    b = 5 
    print(names.keys())   # still print out []!
    print(locals().keys())  # ['names', 'a', 'b']
    print(names.keys())   # ['names', 'a', 'b']
```

This makes the use of `locals()` highly unreliable inside a function.

### LINKED ISSUE #379: Error when accessing schema that already has ~log table but you don't have access to
```python
natexp = dj.create_virtual_module('scadena_natimexperiment', 'scadena_natimexperiment')
```

    ---------------------------------------------------------------------------

    OperationalError                          Traceback (most recent call last)

    /usr/local/lib/python3.5/dist-packages/datajoint/base_relation.py in insert(self, rows, replace, ignore_errors, skip_duplicates, ignore_extra_fields)
        262                         placeholders=','.join('(' + ','.join(row['placeholders']) + ')' for row in rows)),
    --> 263                     args=list(itertools.chain.from_iterable((v for v in r['values'] if v is not None) for r in rows)))
        264             except pymysql.err.OperationalError as err:

    /usr/local/lib/python3.5/dist-packages/datajoint/connection.py in query(self, query, args, as_dict)
        124                 warnings.simplefilter("ignore")
    --> 125                 cur.execute(query, args)
        126         except err.OperationalError as e:

    /usr/local/lib/python3.5/dist-packages/pymysql/cursors.py in execute(self, query, args)
        165 
    --> 166         result = self._query(query)
        167         self._executed = query

    /usr/local/lib/python3.5/dist-packages/pymysql/cursors.py in _query(self, q)
        321         self._last_executed = q
    --> 322         conn.query(q)
        323         self._do_get_result()

    /usr/local/lib/python3.5/dist-packages/pymysql/connections.py in query(self, sql, unbuffered)
        855         self._execute_command(COMMAND.COM_QUERY, sql)
    --> 856         self._affected_rows = self._read_query_result(unbuffered=unbuffered)
        857         return self._affected_rows

    /usr/local/lib/python3.5/dist-packages/pymysql/connections.py in _read_query_result(self, unbuffered)
       1056             result = MySQLResult(self)
    -> 1057             result.read()
       1058         self._result = result

    /usr/local/lib/python3.5/dist-packages/pymysql/connections.py in read(self)
       1339         try:
    -> 1340             first_packet = self.connection._read_packet()
       1341 

    /usr/local/lib/python3.5/dist-packages/pymysql/connections.py in _read_packet(self, packet_type)
       1013         packet = packet_type(buff, self.encoding)
    -> 1014         packet.check_error()
       1015         return packet

    /usr/local/lib/python3.5/dist-packages/pymysql/connections.py in check_error(self)
        392             if DEBUG: print("errno =", errno)
    --> 393             err.raise_mysql_exception(self._data)
        394 

    /usr/local/lib/python3.5/dist-packages/pymysql/err.py in raise_mysql_exception(data)
        106     errorclass = error_map.get(errno, InternalError)
    --> 107     raise errorclass(errno, errval)
    

    OperationalError: (1142, "INSERT command denied to user 'ewang'@'10.21.17.251' for table '~log'")

    
    During handling of the above exception, another exception occurred:

    DataJointError                            Traceback (most recent call last)

    <ipython-input-3-ad23ec27eb78> in <module>()
          4 detect = dj.create_virtual_module('detect', 'detect')
          5 sort = dj.create_virtual_module('sort', 'sort')
    ----> 6 natexp = dj.create_virtual_module('scadena_natimexperiment', 'scadena_natimexperiment')
          7 natimg = dj.create_virtual_module('scadena_synth_naturalimages', 'scadena_synth_naturalimages')

    /usr/local/lib/python3.5/dist-packages/datajoint/__init__.py in create_virtual_module(modulename, dbname)
         84     """
         85     mod = ModuleType(modulename)
    ---> 86     s = schema(dbname, mod.__dict__)
         87     s.spawn_missing_classes()
         88     mod.__dict__['schema'] = s

    /usr/local/lib/python3.5/dist-packages/datajoint/schema.py in __init__(self, database, context, connection, create_tables)
         71                 else:
         72                     self.log('created')
    ---> 73         self.log('connect')
         74         connection.register(self)
         75 

    /usr/local/lib/python3.5/dist-packages/datajoint/base_relation.py in __call__(self, event)
        581                 version=version + 'py',
        582                 host=platform.uname().node,
    --> 583                 event=event), ignore_errors=True, ignore_extra_fields=True)
        584         except pymysql.err.OperationalError:
        585             logger.info('could not log event in table ~log')

    /usr/local/lib/python3.5/dist-packages/datajoint/base_relation.py in insert1(self, row, **kwargs)
        123         For kwargs, see insert()
        124         """
    --> 125         self.insert((row,), **kwargs)
        126 
        127     def insert(self, rows, replace=False, ignore_errors=False, skip_duplicates=False, ignore_extra_fields=False):

    /usr/local/lib/python3.5/dist-packages/datajoint/base_relation.py in insert(self, rows, replace, ignore_errors, skip_duplicates, ignore_extra_fields)
        264             except pymysql.err.OperationalError as err:
        265                 if err.args[0] == server_error_codes['command denied']:
    --> 266                     raise DataJointError('Command denied:  %s' % err.args[1])
        267                 else:
        268                     raise

    DataJointError: Command denied:  INSERT command denied to user 'ewang'@'10.21.17.251' for table '~log'

dimitri-yatsenko:
# Garbage Collection
Implement garbage collection for external. 

# Use of local context by the schema decorator

In creation of the `schema` decorator, the context keyword now defaults to the local context.

Before:
```python
schema = dj.schema('pipeline_tuning', locals())
```
After
```python
schema = dj.schema('pipeline_tuning')
```

The default context value should also fix #368.

Also fixed #379.

### DIFF
--- a/datajoint/external.py
+++ b/datajoint/external.py
@@ -177,7 +177,7 @@ class ExternalTable(BaseRelation):
                 for ref in self.references) or "TRUE")
         print('Deleted %d items' % self.connection.query("SELECT ROW_COUNT()").fetchone()[0])
 
-    def clean_store(self, store):
+    def clean_store(self, store, display_progress=True):
         """
         Clean unused data in an external storage repository from unused blobs.
         This must be performed after delete_garbage during low-usage periods to reduce risks of data loss.
@@ -191,11 +191,13 @@ class ExternalTable(BaseRelation):
         except KeyError:
             raise DataJointError('Storage {store} config is missing the protocol field'.format(store=store))
 
+        progress = tqdm if display_progress else lambda x: x
+
         if protocol == 'file':
             folder = os.path.join(spec['location'], self.database)
             delete_list = set(os.listdir(folder)).difference(self.fetch('hash'))
             print('Deleting %d unused items from %s' % (len(delete_list), folder), flush=True)
-            for f in tqdm(delete_list):
+            for f in progress(delete_list):
                 os.remove(os.path.join(folder, f))
         else:
             raise DataJointError('Unknown external storage protocol {protocol} for {store}'.format(
--- a/datajoint/external.py
+++ b/datajoint/external.py
@@ -70,7 +70,7 @@ class ExternalTable(BaseRelation):
                     safe_write(full_path, blob)
         else:
             raise DataJointError('Unknown external storage protocol {protocol} for {store}'.format(
-                store=store, protocol=protocol))
+                store=store, protocol=spec['protocol']))
 
         # insert tracking info
         self.connection.query(
@@ -126,16 +126,6 @@ class ExternalTable(BaseRelation):
         FROM information_schema.key_column_usage
         WHERE referenced_table_name="{tab}" and referenced_table_schema="{db}"
         """.format(tab=self.table_name, db=self.database), as_dict=True)
-
-    @property
-    def garbage_count(self):
-        """
-        :return: number of items that are no longer referenced
-        """
-        return self.connection.query(
-            "SELECT COUNT(*) FROM `{db}`.`{tab}` WHERE ".format(tab=self.table_name, db=self.database) +
-            (" AND ".join('hash NOT IN (SELECT {column_name} FROM {referencing_table})'.format(**ref)
-                          for ref in self.references) or "TRUE")).fetchone()[0]
 
     def delete(self):
         return self.delete_quick()
--- a/datajoint/external.py
+++ b/datajoint/external.py
@@ -161,7 +161,7 @@ class ExternalTable(BaseRelation):
         raise DataJointError('Please use delete_garbage instead')
 
     def drop_quick(self):
-        if not self:
+        if self:
             raise DataJointError('Cannot non-empty external table. Please use delete_garabge to clear it.')
         self.drop_quick()
 
@@ -193,10 +193,9 @@ class ExternalTable(BaseRelation):
 
         if protocol == 'file':
             folder = os.path.join(spec['location'], self.database)
-            lst = set(os.listdir(folder))
-            lst.difference_update(self.fetch('hash'))
-            print('Deleting %d unused items from %s' % (len(lst), folder), flush=True)
-            for f in tqdm(lst):
+            delete_list = set(os.listdir(folder)).difference(self.fetch('hash'))
+            print('Deleting %d unused items from %s' % (len(delete_list), folder), flush=True)
+            for f in tqdm(delete_list):
                 os.remove(os.path.join(folder, f))
         else:
             raise DataJointError('Unknown external storage protocol {protocol} for {store}'.format(
--- a/datajoint/external.py
+++ b/datajoint/external.py
@@ -140,7 +140,7 @@ class ExternalTable(BaseRelation):
     def drop_quick(self):
         """drop the external table -- works only when it's empty"""
         if self:
-            raise DataJointError('Cannot non-empty external table. Please use delete_garabge to clear it.')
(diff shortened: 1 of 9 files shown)


datajoint/external.py
```diff
@@ -113,29 +103,88 @@ def get(self, blob_hash):
...
+        SELECT concat('`', table_schema, '`.`', table_name, '`') as referencing_table, column_name
+        FROM information_schema.key_column_usage
+        WHERE referenced_table_name="{tab}" and referenced_table_schema="{db}"
+        """.format(tab=self.table_name, db=self.database), as_dict=True)
+
+    def delete(self):
+        return self.delete_quick()
+
+    def delete_quick(self):
+        raise DataJointError('The external table does not support delete. Please use delete_garbage instead.')
```
eywalker: Why do we not support delete/drop on external? Given that all dependent tables that use external file refers to the external table, the deletion should safely propagate.

dimitri-yatsenko: external is not really a table from the user's point of view.  They don't need to know or think about it.  It does not show in the ERD.  In a way, it is outside the regular data model.  The only correct way to delete objects is to delete their entries from their tables and then do garbage cleanup if the external table.

eywalker: I see. It would still be nice to have a way to delete entries in external for which the correspkndkng file is missing.

dimitri-yatsenko: That's what `delete_garbage` does.

dimitri-yatsenko: Wait, if the file is missing, it indicates a  problem.  That would be a diagnostic tool

dimitri-yatsenko: The normal workflow is: 
1. Delete from referencing table
2. `schema.external.delete_garbage()`  -- to delete from the external table
3. `schema.external.clean_store()`  -- to delete the files

eywalker: Yes, if file is missing, that does indicate a problem, but then it would be nice to be able to run delete from the external table to delete any table entry that references the missing entry. I guess all I want is `delete` on External perhaps with option `force` to be able to delete when needed.

dimitri-yatsenko: I am hesitant to add such a feature because it might give someone the wrong idea.  In normal use, no one should be deleting from the external table unless for garbage collection.

eywalker: Ok, I guess we can keep such deletion feature out of `External` for now and think about it when the need arise. Given that the situation of missing external file is a possible scenario (while it does certainly indicate some problem in the setup), it would be important to have a way for users to be able to identify which tables reference the now invalid entry, but perhaps this need not happen as `delete`.

Issue #416, opened 2017-12-21
Problems with pymysql 0.8.0


aecker:
I just ran into an issue where DJ seemed to be unable to connect to the database:
```
import datajoint as dj
dj.conn()

Connecting aecker@52.202.15.166:3306
---------------------------------------------------------------------------
DataJointError                            Traceback (most recent call last)
<ipython-input-1-1e1724e7ecff> in <module>()
     11 
     12 import datajoint as dj
---> 13 dj.conn()
     14 
     15 import tensorflow as tf

/usr/local/lib/python3.5/dist-packages/datajoint/connection.py in conn(host, user, password, init_fun, reset)
     40             password = getpass(prompt="Please enter DataJoint password: ")
     41         init_fun = init_fun if init_fun is not None else config['connection.init_function']
---> 42         conn.connection = Connection(host, user, password, init_fun)
     43     return conn.connection
     44 

/usr/local/lib/python3.5/dist-packages/datajoint/connection.py in __init__(self, host, user, password, init_fun)
     72             self.connection_id = self.query('SELECT connection_id()').fetchone()[0]
     73         else:
---> 74             raise DataJointError('Connection failed.')
     75         self._conn.autocommit(True)
     76         self._in_transaction = False

DataJointError: Connection failed.
```

The problem turned out to be the pymysql version (0.8.0). A downgrade to an older version (I used 0.7.11, which was working for me in the past) solved the issue. I'm not sure what exactly is the problem and don't have the time to investigate it in depth at the moment, but I thought I'd post the issue and its solution here in case someone else runs into a similar issue.

ixcat:
Thanks for the report.
It looks like pymysql's 'ping' function used to return a value we were using to check connection active, but now apparently relies on exceptions to convey this information.

(see also: https://github.com/PyMySQL/PyMySQL/commit/8009ec5059f3bf47ab77118bb48cbdc88c34f89a#diff-884ad974e6382c5ab0dee054b727f7cd)

Hacking in the following allows our unit tests to run; obviously this is not a real fix - we are discussing how to handle this further but I wanted to post status.

```
diff --git a/datajoint/connection.py b/datajoint/connection.py
index cc4bc61..513fff7 100644
--- a/datajoint/connection.py
+++ b/datajoint/connection.py
@@ -102,7 +102,8 @@ class Connection:
         """
         Returns true if the object is connected to the database server.
         """
-        return self._conn.ping()
+        # return self._conn.ping()
+        return True
 
[...]

PR #419, opened 2017-12-22
connection.py: change Connection.is_connected to use exception handli…

ixcat:
…ng logic

Reported as issue #416 against PyMySQL v0.8.0;
Apparrently related to changes in PyMySQL rev 8009ec5.

### DIFF
--- a/datajoint/connection.py
+++ b/datajoint/connection.py
@@ -102,7 +102,11 @@ class Connection:
         """
         Returns true if the object is connected to the database server.
         """
-        return self._conn.ping()
+        try:
+            self._conn.ping()
+            return True
+        except:
+            return False
 
     def query(self, query, args=(), as_dict=False, suppress_warnings=True):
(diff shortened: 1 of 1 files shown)


datajoint/connection.py
```diff
@@ -102,7 +102,11 @@ def is_connected(self):
         """
         Returns true if the object is connected to the database server.
         """
-        return self._conn.ping()
+        try:
+            self._conn.ping()
```
dimitri-yatsenko: What exactly happens when no connection?

eywalker: If a stale connection is present, previously `ping` caused the re-establishment of the stale connection. This solution is not ideal as it would not work properly for previous versions of the PyMySQL's `ping`. To be compatible to both the new and old, use `try... return self.__conn.ping()` or require the new Pymysql version in the requirements.

Issue #300, opened 2017-03-13
correct cascaded deletes across renamed foreign keys


dimitri-yatsenko:
a more detailed descriptions will follow.

dimitri-yatsenko:
This was actually not fixed.  #371 stated that restructuring was done for #300 but the fix was not completed.

ixcat:
note:
tests/test_foreign_keys.py
test_aliased_fk
test_delete
may need adjustment as part of fixing this.

PR #424, opened 2017-12-28
Revamp delete

### LINKED ISSUE #418: Automatically create the cache folder used for external files
Setting the cache folder (`dj.config['cache'] = '/tmp/dj-cache'`) to a folder that doesn't exist throws a FileNotFoundError when fetching. It would be nice if it created it automatically
Here's the error
```
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/dj-cache/su7VnvVwCN901A7maY3lX4es-_digGrC3mFQtcSXdHEbl3d.saving'
```

### LINKED ISSUE #430: Type error when using `max_calls` in the autopopulate function
The argument `max_calls` in the autopopulate function added in https://github.com/datajoint/datajoint-python/commit/7299058ee37c44216a5a53acbcb737d68f47fdc4 throws a type error:

```
--> 126             if max_calls is not None and call_count >= max_calls:
    127                 break

TypeError: unorderable types: itertools.count() >= int()
```

dimitri-yatsenko:
* Optimize delete.  
Previously entries were counted first and then deleted.  Now they are deleted within a transaction with the count reported to the users.  Only afterwards, the user is asked to confirm before commit.

* Complete class-based relational algebra.
Previously, all relational operators required instantiating user classes.  Now all operators can be performed directly on the classes.  
For example, one can do this now
```python 
stimulus.Sync & (stimulus.Trial * stimulus.Trippy.proj())
```
Previously this query would have required instantiating the relations:
```python 
stimulus.Sync() & (stimulus.Trial() * stimulus.Trippy().proj())
```
This change is fully backward compatible.

* Add efficient iterator capability to relational operands:
```python
for session in stimulus.Sync:
    print(session)

for trippy_trial in stimulus.Trial * stimulus.Trippy & 'animal_id=16342':
    plt.imshow(trippy_trial.movie[:,:,0])
```
This uses memory efficiently by retrieving only the primary keys first and then fetching each individual item in its entirety at each iteration.  However, when the relation only has primary keys, then then only one fetch operation is performed.  The following two operations are essentially equivalent:

```python
for key in stimulus.Sync.proj():
    print(key)

for key in stimulus.Sync.fetch('KEY'):
    print(key)
```

* Fix #418 

* Fix #430

* This PR is based off #422


tests/test_foreign_keys.py
```diff
@@ -1,11 +1,13 @@
...
+from nose.tools import assert_equal, assert_false, assert_true, raises
 from datajoint.declare import declare
+from datajoint import DataJointError
 
 from . import schema_advanced
 
 context = schema_advanced.schema.context
 
 
+@raises(DataJointError)     # TODO: remove after fixing issue #300
```
ixcat: would it be better to add a ```raise SkipTest('fix after issue 300')``` for these? 

That would theoretically make the 'brokenness' of the test slightly more visible so this isn't lost..

Either way, I've added a note in issue 300 itself.

datajoint/base_relation.py
```diff
@@ -341,35 +349,42 @@ def delete(self):
...
                 if not already_in_transaction:
-                    self.connection.commit_transaction()
-                print('Done')
+                    self.connection.cancel_transaction()
+            else:
+                if already_in_transaction:
+                    if verbose:
+                        print('The delete is pending within the ongoing transaction.')
+                else:
+                    if not safe or user_choice("Proceed?", default='no') == 'yes':
```
ixcat: My one concern with this approach is that it has the potential to lock the table when waiting for user input..
I suppose this is a tradeoff between potentially inconsistent prompting notification and actual delete actions..
Don't think it is a reason to block and I think I do prefer this approach, but thought I'd mention

tests/test_foreign_keys.py
dimitri-yatsenko: yes, thanks.  I just learned about `SkipTest`.

dimitri-yatsenko: actually, I do want to test that it throws a DataJoint error when an attempt to cascade delete across renamed foreign keys until #300 is fixed.  So I will keep it unchanged.

datajoint/base_relation.py
dimitri-yatsenko: The blocking is a bit more complex than this.  Block would have occurred under the previous approach as well after the consistent read was made.

tests/test_foreign_keys.py
eywalker: Yes keep this in until #300 is fixed. Until then, the right behavior is to throw `DataJointError`. (but `SkipTest` is good to know - I didn't know about it)

datajoint/base_relation.py
```diff
@@ -289,27 +289,35 @@ def check_fields(fields):
...
+    def delete(self, verbose=True):
         """
         Deletes the contents of the table and its dependent tables, recursively.
         User is prompted for confirmation if config['safemode'] is set to True.
         """
+        already_in_transaction = self.connection.in_transaction
+        safe = config['safemode']
+        if already_in_transaction and safe:
+            raise DataJointError('Cannot delete within a transaction in safemode. '
+                                 'Set dj.config["setmode"] = False or complete the ongoing transaction first.')
```
eywalker: Typo in the error message: should be `Set dj.config["safemode"] = False or...`

datajoint/base_relation.py
```diff
@@ -289,27 +289,35 @@ def check_fields(fields):
...
 
-    def delete(self):
+    def delete(self, verbose=True):
         """
         Deletes the contents of the table and its dependent tables, recursively.
         User is prompted for confirmation if config['safemode'] is set to True.
         """
+        already_in_transaction = self.connection.in_transaction
+        safe = config['safemode']
+        if already_in_transaction and safe:
```
eywalker: What's the rationale for disallowing deletion when already in transaction if in safe mode? This differs from previous behavior as far as I recall.

datajoint/base_relation.py
```diff
@@ -389,6 +404,8 @@ def drop(self):
         Drop the table and all tables that reference it, recursively.
         User is prompted for confirmation if config['safemode'] is set to True.
         """
+        if self.restriction:
+            raise DataJointError('A relation with an applied restriction condition cannot be dropped.')
```
eywalker: Append how to solve this in the message. Something like `Call drop() directly on the unrestricted UserRelation`

datajoint/base_relation.py
dimitri-yatsenko: Because there is no way to under the delete in the middle of a transaction without rolling back the entire transaction.  So there is no *safe* delete within a transaction.

datajoint/base_relation.py
dimitri-yatsenko: agreed.

datajoint/base_relation.py
eywalker: Sounds fair

eywalker: Sounds fair

PR #443, opened 2018-01-31
networkx-2 support 

ixcat:
Fixes for networkx version>2

Not heavily tested - unit tests and manual ERD draw for simple single schema set of relations OK.

### DIFF
--- a/datajoint/dependencies.py
+++ b/datajoint/dependencies.py
@@ -7,8 +7,13 @@ from . import DataJointError
 class Dependencies(nx.DiGraph):
     """
     The graph of dependencies (foreign keys) between loaded tables.
+
+    Note: the 'connnection' argument should normally be supplied;
+    Empty use is permitted to facilliate use of networkx algorithms which
+    internally create objects with the expectation of empty constructors.
+    See also: https://github.com/datajoint/datajoint-python/pull/443
     """
-    def __init__(self, connection):
+    def __init__(self, connection=None):
         self._conn = connection
         self._node_alias_count = itertools.count()
         super().__init__(self)
@@ -100,6 +105,9 @@ class Dependencies(nx.DiGraph):
         :param full_table_name:  In form `schema`.`table_name`
         :return: all dependent tables sorted in topological order.  Self is included.
         """
-        nodes = nx.algorithms.dag.descendants(self, full_table_name)
-        sort = nx.algorithms.dag.topological_sort(self)
-        return [full_table_name] + [s for s in sort if s in nodes]
+
+        nodes = self.subgraph(
+            nx.algorithms.dag.descendants(self, full_table_name))
+
+        return [full_table_name] + list(
+            nx.algorithms.dag.topological_sort(nodes))
--- a/datajoint/dependencies.py
+++ b/datajoint/dependencies.py
@@ -100,5 +100,8 @@ class Dependencies(nx.DiGraph):
         :param full_table_name:  In form `schema`.`table_name`
         :return: all dependent tables sorted in topological order.  Self is included.
         """
-        nodes = nx.algorithms.dag.descendants(self, full_table_name)
-        return [full_table_name] + nx.algorithms.dag.topological_sort(self, nodes)
+        # nx v1
+        # nodes = nx.algorithms.dag.descendants(self, full_table_name)
+        # return [full_table_name] + nx.algorithms.dag.topological_sort(self, nodes)
+        desc = list(nx.algorithms.dag.descendants(self, full_table_name))
+        return [full_table_name] + desc
--- a/datajoint/dependencies.py
+++ b/datajoint/dependencies.py
@@ -100,8 +100,6 @@ class Dependencies(nx.DiGraph):
         :param full_table_name:  In form `schema`.`table_name`
         :return: all dependent tables sorted in topological order.  Self is included.
         """
-        # nx v1
-        # nodes = nx.algorithms.dag.descendants(self, full_table_name)
-        # return [full_table_name] + nx.algorithms.dag.topological_sort(self, nodes)
-        desc = list(nx.algorithms.dag.descendants(self, full_table_name))
-        return [full_table_name] + desc
+        nodes = nx.algorithms.dag.descendants(self, full_table_name)
+        sort = nx.algorithms.dag.topological_sort(self)
+        return [full_table_name] + [s for s in sort if s in nodes]
--- a/datajoint/erd.py
+++ b/datajoint/erd.py
@@ -216,7 +216,6 @@ else:
             new_names = [mapping.values()]
             if len(new_names) > len(set(new_names)):
                 raise DataJointError('Some classes have identical names. The ERD cannot be plotted.')
-            # v2: either copy=True or make 'real graph' in subgraph call above
             nx.relabel_nodes(graph, mapping, copy=False)
             return graph
 
--- a/datajoint/erd.py
+++ b/datajoint/erd.py
@@ -206,8 +206,8 @@ else:
                 nx.algorithms.boundary.node_boundary(nx.DiGraph(self).reverse(), self.nodes_to_show))
             nodes = self.nodes_to_show.union(a for a in gaps if a.isdigit)
             # construct subgraph and rename nodes to class names
-            graph = nx.DiGraph(self).subgraph(nodes)
-            nx.set_node_attributes(graph, 'node_type', {n: _get_tier(n) for n in graph})
+            graph = nx.DiGraph(nx.DiGraph(self).subgraph(nodes))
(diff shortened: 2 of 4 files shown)


datajoint/erd.py
```diff
@@ -206,8 +206,8 @@ def _make_graph(self):
                 nx.algorithms.boundary.node_boundary(nx.DiGraph(self).reverse(), self.nodes_to_show))
             nodes = self.nodes_to_show.union(a for a in gaps if a.isdigit)
             # construct subgraph and rename nodes to class names
-            graph = nx.DiGraph(self).subgraph(nodes)
-            nx.set_node_attributes(graph, 'node_type', {n: _get_tier(n) for n in graph})
+            graph = nx.DiGraph(nx.DiGraph(self).subgraph(nodes))
```
eywalker: what's the reason for the redundancy?

requirements.txt
```diff
@@ -3,6 +3,6 @@ pymysql>=0.7.2
 pyparsing
 ipython
 tqdm
-networkx~=1.11
-pydotplus
+networkx
+pydot
```
eywalker: why `pydot` vs `pydotplus`?

datajoint/erd.py
ixcat: The call now returns a subgraph 'view', which is immutable and cannot be relabeled later on in line 219's ```nx.relabel_nodes(graph, mapping, copy=False)```. I also tried doing a ```return nx.relabel_nodes(graph, mapping, copy=True)``` in that line but it didn't work on 1st pass - from here I didn't really investigate the API change further - since copy=True imples making a copy and the line you are mentioning would also makes a a copy I viewed these as equivalent and moved on..

requirements.txt
ixcat: Networkx apparently flipped in 2.0; my memory from research a few months back was that pydot went unmaintained for a bit, pydotplus was a python 3 + added feature fork, subsequently pydot development picked up and supports pydot3 and pydotplus appears dormant.

1.11 docs:

https://networkx.github.io/documentation/networkx-1.11/reference/drawing.html#module-networkx.drawing.nx_pydot

```
Pydot

Import and export NetworkX graphs in Graphviz dot format using pydotplus.

Either this module or nx_agraph can be used to interface with graphviz.

See also

PyDotPlus
    https://github.com/carlos-jenkins/pydotplus
```

2.1 docs:

https://networkx.github.io/documentation/stable/reference/drawing.html#module-networkx.drawing.nx_pydot

```
Pydot

Import and export NetworkX graphs in Graphviz dot format using pydot.

Either this module or nx_agraph can be used to interface with graphviz.

See also

pydot
    https://github.com/erocarrera/pydot
Graphviz
```

Thanks for pointing it out - this reminds me I should double check rendering on Windows; my recollection from the initial testing of pydotplus to avoid the graphviz issues is that both seem to work but I'm not 100% on this.

datajoint/dependencies.py
```diff
@@ -101,4 +101,5 @@ def descendants(self, full_table_name):
         :return: all dependent tables sorted in topological order.  Self is included.
         """
         nodes = nx.algorithms.dag.descendants(self, full_table_name)
-        return [full_table_name] + nx.algorithms.dag.topological_sort(self, nodes)
+        sort = nx.algorithms.dag.topological_sort(self)
+        return [full_table_name] + [s for s in sort if s in nodes]
```
ixcat: alternate/better performing implementation suggestion @eywalker:
create a subgraph from full_table_name and descendants and then sort

datajoint/erd.py
eywalker: Yes I'm imagining that `self.subgraph` should suffice. @dimitri-yatsenko do you recall why we have to do `nx.Digraph(self).subgraph`? I have a feeling that this is legacy code

ixcat: working on the speedtest now - it seems the constructor is called internally and the ```__init___``` signatures between the classes vary, though I haven't dug further just yet & theoretically calling via superclass would work around this..

ixcat: it seems that networkx's ```digraph.py`` calls H.__class__() to create an instance copy within the ```subgraph``` function - since this refers to ```Dependencies``` which requires ```connection``` as a parameter, the call fails. Making this an optional argument allows ```self.subgraph(...)``` to work, however, this could potentially be undesirable w/r/t the rest of the code depending on opinions...

datajoint/dependencies.py
ixcat: Based on a test of calling ```descendants``` 10000 times for every class within the graph of 41 classes discussed elsewhere, I was able to get the following results locally:

1) initial code above: ~250 iterations of the subclasses-for-each-classes per second
2) Commented code in patch below: ~125 iterations/sec
3) Live code in patch below: ~450 iterations/sec, however this requires making ```connection``` optional in the ```Dependencies``` constructor

can provide the test code elsewhere if that might be useful.

Patch:
--8<--
```
diff --git a/datajoint/dependencies.py b/datajoint/dependencies.py
index 6aa169b..2c0167f 100644
--- a/datajoint/dependencies.py
+++ b/datajoint/dependencies.py
@@ -8,7 +8,7 @@ class Dependencies(nx.DiGraph):
     """
     The graph of dependencies (foreign keys) between loaded tables.
     """
-    def __init__(self, connection):
+    def __init__(self, connection=None):
         self._conn = connection
         self._node_alias_count = itertools.count()
         super().__init__(self)
@@ -95,7 +95,7 @@ class Dependencies(nx.DiGraph):
         return dict(p[1:3] for p in self.out_edges(table_name, data=True)
                     if primary is None or p[2]['primary'] == primary)
 
-    def descendants(self, full_table_name):
+    def descendants_old(self, full_table_name):
         """
         :param full_table_name:  In form `schema`.`table_name`
         :return: all dependent tables sorted in topological order.  Self is included.
@@ -103,3 +103,18 @@ class Dependencies(nx.DiGraph):
         nodes = nx.algorithms.dag.descendants(self, full_table_name)
         sort = nx.algorithms.dag.topological_sort(self)
         return [full_table_name] + [s for s in sort if s in nodes]
+
+    def descendants(self, full_table_name):
+        """
+        :param full_table_name:  In form `schema`.`table_name`
+        :return: all dependent tables sorted in topological order.  Self is included.
+        """
+
+        # nodes = nx.DiGraph(self).subgraph(
+        #     nx.algorithms.dag.descendants(self, full_table_name))
+
+        nodes = self.subgraph(
+            nx.algorithms.dag.descendants(self, full_table_name))
+
+        return [full_table_name] + list(
+            nx.algorithms.dag.topological_sort(nodes))
```

this implies that a similar modification to ```erd.py``` line 209 discussed in other comments could also improve performance; again, however, this requires making ```connection``` optional in the ```Dependencies``` code. The previous implementation running at ~250 iterations shouldn't be impacted by a similar change since it is not constructing copied instances within the main method body. As a side note, unit tests still pass with ```connnection`` an optional argument.

datajoint/erd.py
eywalker: We probably should override `subgraph` so `connection` gets passed down - although it also makes sense to check if `connection` is needed in the subgraphs.

ixcat: Thinking to audit for connection use and proceed or just take the performance hit -

Maintaining a random graph related function in-tree to avoid this issue doesn't seem architecturally correct since this class isn't *really* a 'graph subclass' in terms of 'providing additional graph functionality' (even though it is *technically* a subclass) .. 

That said it isn't such a big deal to copy the function in, annotate the reasoning and sync when needed with upstream. Networkx is BSD licencsed btw so this wouldn't impact DJ.

@dimitri-yatsenko care to weigh in?

dimitri-yatsenko: If everything works, then no need to pass the connection

ixcat: @dimitri-yatsenko apologies if unclear - so does this mean the additional patch to ```datajoint/dependencies.py ``` mentioned in the comments here can be added to this PR after comment cleanup?

datajoint/dependencies.py
```diff
@@ -100,5 +105,9 @@ def descendants(self, full_table_name):
         :param full_table_name:  In form `schema`.`table_name`
         :return: all dependent tables sorted in topological order.  Self is included.
         """
-        nodes = nx.algorithms.dag.descendants(self, full_table_name)
-        return [full_table_name] + nx.algorithms.dag.topological_sort(self, nodes)
+
+        nodes = self.subgraph(
```
eywalker: I think it would make sense to just use `nx.DiGraph(self).subgraph(...)` here instead just like we do in the `ERD` and don't make `connection` optional. Also, the variable name `nodes` is a bit misleading - I'd say give it a name that suggests that it is a graph and not a mere list of nodes.

PR #456, opened 2018-03-05
Fix  #431,  #451, #458, #463, #466

### LINKED ISSUE #431: ERD Tooltip does not always substitute class names
From issue  #395 and fix #422, #423 -- the class name does not always get expanded in foreign key references, especially when classes are constructed through `schema.spawn_missing_classes` or `dj.create_virtual_module`

### LINKED ISSUE #451: Inserting with `skip_duplicates=True` fails with error message "attr X is ambiguous"
I'm encountering a case in which inserting with `skip_duplicates=True` results in an error. The effective query performed by the insert is:

```sql
'INSERT INTO `cortex_ex_machina_mesonet_analysis2018_02_20`.`__high_oracle_units` (`group_id`,`data_hash`,`readout_key`,`neuron_id`) SELECT `group_id`,`data_hash`,`readout_key`,`neuron_id` FROM (SELECT * FROM `cortex_ex_machina_mesonet_analysis2018_02_20`.`__target_dataset__unit` WHERE ((`data_hash`=\'a502f8a1ad8fab1f3fea0b3638af26a4\') AND (`group_id`=4)) AND ((`group_id`,`readout_key`,`unit_id`,`animal_id`) in (SELECT `group_id`,`readout_key`,`unit_id`,`animal_id` FROM (SELECT `group_id`,`animal_id`,`session`,`scan_idx`,`window`,`unit_id`,`pearson`,concat_ws("-", animal_id, session, scan_idx) as `readout_key` FROM `cortex_ex_machina_mesonet_data`.`__meso_net_all_oracle__unit_scores`) as `_s16` WHERE ((`group_id`,`animal_id`,`unit_id`,`readout_key`) in (SELECT `group_id`,`animal_id`,`unit_id`,`readout_key` FROM `cortex_ex_machina_mesonet_analysis2018_02_20`.`__target_dataset__unit` WHERE ((`data_hash`=\'a502f8a1ad8fab1f3fea0b3638af26a4\') AND (`group_id`=4)))) AND (pearson > 0.395305)))) as `_s17` ON DUPLICATE KEY UPDATE `group_id`=`group_id`'
```
 When the above query is attempted, it fails with an error: `
InternalError: (1052, "Column 'group_id' in field list is ambiguous")`

It appears to have to do with the fact that the attribute `group_id` is present on both target and source of the insert. I'll need to look into this further, but it may be the case that our "workaround" for duplicate entry (i.e. the `ON DUPLICATE KEY UPDATE group_id=group_id`) is not compatible when inserting from another relation.

### LINKED ISSUE #458: len(rel) is slow
I noticed that if I have a long table (relation) `rel`, then calling `len(rel)` can be quite slow. It takes about 2 seconds on our setup to return if the length is say, 300000 tuples. Here's the code in `RelationalOperand`:

```python
    def __len__(self):
        """
        number of tuples in the relation.
        """
        return U().aggr(self, n='count(*)').fetch1('n')
```
After looking at the code for `RelationalOperand.preview()`, I noticed that it uses `len(self.proj())` instead of `len(self)`, which returns the same result in about half the time. Could `__len__` be sped up? Maybe just replace it with `len(self.proj())`?

### LINKED ISSUE #463: (A & B) * B results in an error
Currently the following query:

```python
(A & B) * B
```

throws an error like:
```python
Cannot join relations on dependent attribute `xxx`
```
where `xxx` is a non-primary key attribute of the relation `B`

### LINKED ISSUE #466: Improve error message for missing @schema decorator
If you forget to decorate a datajoint class with its schema, the error message is now more cryptic than it used to be.

```python
class Mouse(dj.Manual):
    definition = """
    mouse_id : int  
    ----
    dob  : date 
    sex  : enum('M', 'F', 'unknown')
    """
    pass

Mouse()
```
```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
/opt/conda/lib/python3.6/site-packages/IPython/core/formatters.py in __call__(self, obj)
    700                 type_pprinters=self.type_printers,
    701                 deferred_pprinters=self.deferred_printers)
--> 702             printer.pretty(obj)
    703             printer.flush()
    704             return stream.getvalue()

/opt/conda/lib/python3.6/site-packages/IPython/lib/pretty.py in pretty(self, obj)
    393                             if callable(meth):
    394                                 return meth(obj, self, cycle)
--> 395             return _default_pprint(obj, self, cycle)
    396         finally:
    397             self.end_group()

/opt/conda/lib/python3.6/site-packages/IPython/lib/pretty.py in _default_pprint(obj, p, cycle)
    508     if _safe_getattr(klass, '__repr__', None) is not object.__repr__:
    509         # A user-provided repr. Find newlines and replace them with p.break_()
--> 510         _repr_pprint(obj, p, cycle)
    511         return
    512     p.begin_group(1, '<')

/opt/conda/lib/python3.6/site-packages/IPython/lib/pretty.py in _repr_pprint(obj, p, cycle)
    699     """A pprint that just redirects to the normal repr function."""
    700     # Find newlines and replace them with p.break_()
--> 701     output = repr(obj)
    702     for idx,output_line in enumerate(output.splitlines()):
    703         if idx:

/opt/conda/lib/python3.6/site-packages/datajoint/relational_operand.py in __repr__(self)
    345 
    346     def __repr__(self):
--> 347         return super().__repr__() if config['loglevel'].lower() == 'debug' else self.preview()
    348 
    349     def preview(self, limit=None, width=None):

/opt/conda/lib/python3.6/site-packages/datajoint/relational_operand.py in preview(self, limit, width)
    351         returns a preview of the contents of the relation.
    352         """
--> 353         heading = self.heading
    354         rel = self.proj(*heading.non_blobs)
    355         if limit is None:

/opt/conda/lib/python3.6/site-packages/datajoint/base_relation.py in heading(self)
     43             self._heading = Heading()  # instance-level heading
     44         if not self._heading:  # lazy loading of heading
---> 45             self._heading.init_from_database(self.connection, self.database, self.table_name)
     46         return self._heading
     47 

/opt/conda/lib/python3.6/site-packages/datajoint/heading.py in init_from_database(self, conn, database, table_name)
    132         initialize heading from a database table.  The table must exist already.
    133         """
--> 134         info = conn.query('SHOW TABLE STATUS FROM `{database}` WHERE name="{table_name}"'.format(
    135             table_name=table_name, database=database), as_dict=True).fetchone()
    136         if info is None:

AttributeError: 'NoneType' object has no attribute 'query'

---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
/opt/conda/lib/python3.6/site-packages/IPython/core/formatters.py in __call__(self, obj)
    343             method = get_real_method(obj, self.print_method)
    344             if method is not None:
--> 345                 return method()
    346             return None
    347         else:

/opt/conda/lib/python3.6/site-packages/datajoint/relational_operand.py in _repr_html_(self)
    373 
    374     def _repr_html_(self):
--> 375         heading = self.heading
    376         rel = self.proj(*heading.non_blobs)
    377         info = heading.table_info

/opt/conda/lib/python3.6/site-packages/datajoint/base_relation.py in heading(self)
     43             self._heading = Heading()  # instance-level heading
     44         if not self._heading:  # lazy loading of heading
---> 45             self._heading.init_from_database(self.connection, self.database, self.table_name)
     46         return self._heading
     47 

/opt/conda/lib/python3.6/site-packages/datajoint/heading.py in init_from_database(self, conn, database, table_name)
    132         initialize heading from a database table.  The table must exist already.
    133         """
--> 134         info = conn.query('SHOW TABLE STATUS FROM `{database}` WHERE name="{table_name}"'.format(
    135             table_name=table_name, database=database), as_dict=True).fetchone()
    136         if info is None:

AttributeError: 'NoneType' object has no attribute 'query'
```

dimitri-yatsenko:
Fix #431 -- ERD hovertip shows classnames
Fix #451 -- disambiguate fields when skipping duplicates in insert from relation
Fix #458 -- speed up queries
Fix #463 -- Join arguments are enclosed in subqueries if they are restricted
Fix #466 -- Improve error message when missing `@schema`


eywalker:
Could you add a test case to test and verify that this solution works?

eywalker:
This is still waiting a test case to be added before I can accept this PR.

datajoint/schema.py
```diff
@@ -101,7 +101,7 @@ def spawn_missing_classes(self):
         in the context.
         """
         # if self.context is not set, use the calling namespace
-        context = self.context if self.context is not None else inspect.currentframe().f_back.f_locals
+        context = self.context if self.context is not None else inspect.currentframe().f_back.f_globals
```
eywalker: Should search in both locals and globals, in that sequence.

datajoint/base_relation.py
```diff
@@ -434,11 +439,13 @@ def show_definition(self):
         logger.warning('show_definition is deprecated.  Use describe instead.')
         return self.describe()
 
-    def describe(self, printout=True):
+    def describe(self, context=None, printout=True):
         """
         :return:  the definition string for the relation using DataJoint DDL.
             This does not yet work for aliased foreign keys.
         """
+        if context is None:
```
eywalker: Locals + globals here as well

datajoint/base_relation.py
```diff
@@ -42,7 +42,11 @@ def heading(self):
         if self._heading is None:
             self._heading = Heading()  # instance-level heading
         if not self._heading:  # lazy loading of heading
-            self._heading.init_from_database(self.connection, self.database, self.table_name)
+            if self.connection is None:
+                raise DataJointError(
+                    'DataJoint class is missing a database connection. Missing class decorator?')
```
eywalker: Change to something like `'DataJoint class is missing a database connection. Missing schema decorator on class? (i.e. @schema)'`

datajoint/erd.py
```diff
@@ -84,10 +85,9 @@ def __init__(self, source, context=None):
 
             # get the caller's locals()
             if context is None:
-                import inspect
                 frame = inspect.currentframe()
                 try:
-                    context = frame.f_back.f_locals
+                    context = frame.f_back.f_globals
```
eywalker: Locals + globals

eywalker:
What was the issue?

dimitri-yatsenko:
Spawn_missing_classes is special because it uses the context to populate it not just for lookup.

datajoint/base_relation.py
dimitri-yatsenko: done

datajoint/relational_operand.py
```diff
@@ -467,7 +468,12 @@ def __len__(self):
         """
         number of tuples in the relation.
         """
-        return U().aggr(self, n='count(*)').fetch1('n')
+        return self.connection.query(
+            'SELECT ' + (
+                'count(DISTINCT `{pk}`)'.format(pk='`,`'.join(self.primary_key)) if self.distinct else 'count(*)') +
```
ixcat: While this is a definite improvement, I seem to recall`count(*)` can be much slower than `count(some-indexed-or-just-small-field)` in some cases since `count` is an aggregation function performed on the lower-level result set.

Unfortunately I'm having trouble locating any good queries to verify this with (need to have dj generate sufficient joins where the `count(*)` operates on a subquery I believe). I _think_ the lower level implication being that a subselect requires creating a temporary table to hold the `*` data, on which an index lookup is done to compute the result; in cases where `*` is sufficiently small, this temporary table is stored in-memory (`< tmp_table_size`), in other cases, it is written to disk. If only some small field is used, the resulting temporary table is smaller and therefore fits in memory. Even for those cases where an on-disk table is used, a smaller dataset would require less transient data copy. I don't think there is anything to be done for the case of row-scans on large tables (e.g when using a `WHERE` clause). I haven't pinned down all of these things 100% however.

In any event, could the same pkey-splicing logic be used more generally (theoretically just using the first pkey field would work)?

This could potentially be a topic for further investigation; the current change does imply some improvement and I'm not concrete enough about the details here to entirely say for sure.

datajoint/schema.py
```diff
@@ -101,7 +102,12 @@ def spawn_missing_classes(self):
         in the context.
         """
         # if self.context is not set, use the calling namespace
-        context = self.context if self.context is not None else inspect.currentframe().f_back.f_locals
+        if self.context is not None:
+            context = self.context
+        else:
+            frame = inspect.currentframe().f_back
+            context = frame.f_locals
```
eywalker: Why not do the combination of locals and globals as was done in ERD init? In general, that combination is likely to yield the lookup of identifiers as expected by the user.

datajoint/schema.py
```diff
@@ -223,8 +229,4 @@ def external_table(self):
         """
         if self._external is None:
             self._external = ExternalTable(self.connection, self.database)
-        return self._external
-
-    def erd(self):
```
eywalker: Why is this getting removed? Given it was already part of the interface, would need a good reason for a removal, and I don't think this was discussed.

datajoint/relational_operand.py
eywalker: Very good point brought up. It does appear that the `pkey` subselection could already be a general thing and I have been using this implementation of count locally to get a speed ups for all counts operation.

datajoint/relational_operand.py
```diff
@@ -569,7 +575,7 @@ def make_argument_subquery(arg):
         """
         Decide when a Join argument needs to be wrapped in a subquery
         """
-        return Subquery.create(arg) if isinstance(arg, (GroupBy, Projection)) else arg
+        return Subquery.create(arg) if isinstance(arg, (GroupBy, Projection)) or arg.restriction else arg
```
eywalker: This will work but with a potential slow downs as there could be cases where restriction can be passed through without a need for a subquery, and thus this can be a slow down.

datajoint/schema.py
dimitri-yatsenko: This is inside `spawn_missing_classes` and the `context` variable is used for placing new objects, not just looking for existing objects.  So here it works a bit differently from other places.

eywalker: Ah, good point. I should have checked what the `context` was getting used for - we are actually modifying the local space. Thanks for the clarification.

datajoint/relational_operand.py
dimitri-yatsenko: true.  We value correctness over efficiency.  however, we will soon release a new optimization redesign that's based on slightly different overall approach.

eywalker: Between correctness and efficiency, there is no question asked - we always strive for correctness, and there should be no exception to this. This will be a notable optimization point as subquery is known to dramatically decrease query speed. I don't know what different approach you are thinking of, but I agree that this is not the only place that needs attention on query optimization, and might be better dealt together. I'll likely create an issue on this, so we can track improvements in the future.

datajoint/relational_operand.py
dimitri-yatsenko: With a subquery, you are correct, perhaps count(*) causes the full subquery result to be generated rather than just inferring how many rows it should have.  This would require a principally new approach, which we are working on now.

eywalker: Indeed - I don't care to have this addressed in this PR. Let's just keep a note of this and other queries with known performance issues for future improvements.

dimitri-yatsenko: Yes, the upcoming redesign is based on a new approach, leading to fewer subqueries.  But it's a major redesign and I would like to roll it into rev 0.11.0

datajoint/schema.py
dimitri-yatsenko: I don't think `schema.erd` was used much.  It was never documented and it works differently from all other ways to produce the ERD.

eywalker: Ok, as many things in DataJoint has been **undocumented**, I usually don't like removing a feature based on that argument. There has been a lot of things that were *unofficially documented*, say by word of mouth etc. That being said, we do provide clear documentation on how to use the alternative method of `dj.ERD`. There may be still some utility in schema and even class level `erd` methods, but for consistency, I think I can be fine with the removal.

datajoint/connection.py
```diff
@@ -93,7 +94,7 @@ def connect(self):
             warnings.filterwarnings('ignore', '.*deprecated.*')
             self._conn = client.connect(init_command=self.init_fun,
                                         sql_mode="NO_ZERO_DATE,NO_ZERO_IN_DATE,ERROR_FOR_DIVISION_BY_ZERO,"
-                                                 "STRICT_TRANS_TABLES,NO_AUTO_CREATE_USER,NO_ENGINE_SUBSTITUTION",
+                                                 "STRICT_TRANS_TABLES,NO_AUTO_CREATE_USER,NO_ENGINE_SUBSTITUTION", charset="utf8",
```
eywalker: Why not let charset be a parameter to be passed into `dj.conn`? I don't like this kind of hard coding. In fact this probably should be part of `conn_info` with default value of `utf8`.

dimitri-yatsenko: making it part of `dj.config` ...

eywalker: If you are going to do this, there has to be associated handling of `dj.config` where if the option is not given then it defaults to `utf8`.

eywalker: We should consider making it possible to set different connection character set for each DataJoint connection. I'll likely follow up with such change, but I'm fine with this current change.

PR #486, opened 2018-09-06
Return error messages instead of Error objects, fixes #461

### LINKED ISSUE #461: Errors in make() cause memory leak during populate(suppress_errors=True)
I've noticed that the memory usage of the ipython process in which we typically run datajoint can grow quite quickly during a big `populate(suppress_errors=True)` job, which continues despite the errors. Memory usage gets into the multi GB range. Eliminating the errors raised in `make()` eliminated the memory leak. I'm not sure if the memory leak is due to `suppress_errors=True`, or if that just makes it easier to notice.

Any ideas on how to create a minimal example? Is there a public test database or something? I guess all that's needed is to call `make` many times with many keys, and raise an error on each call while suppressing those errors with `populate(suppress_errors=True)`

mspacek:


eywalker:
Makes sense to me. Only question is whether there would be a good usecase for actually returning the error objects (like the way we were doing before). If yes, then we should include a flag that'll let you chose between objects vs message.

mspacek:
OK, done and lightly tested.

ixcat:
A list of messages will still result in memory growth, although at a potentially much slower pace, no?

I'm ok either way, but just wanted to point it out; perhaps keeping the list this is tolerable/preferable.

mspacek:
@ixcat , yes, it should still result in memory growth, but negligibly. From testing on our database, I found no difference in memory usage between appending and not appending error messages. Appending error objects however could quickly balloon memory usage into the GBs.

PR #493, opened 2018-10-09
Partial implementation of nullable foreign key

### LINKED ISSUE #301: foreign key properties
I propose to allow specification of foreign key properties `unique` and `required`.

I propose the following syntax.

```
-> [unique, nullable] lab.Animal 
```

"Unique" means that each animal in `lab.Animal` can be referenced only once from the referencing table.

"Optional" means that the foreign key values can be null.  When null, the unique constraint is not applicable. 

By default, foreign keys are not unique and not optional. 

This syntax is compatible with the syntax for renamed foreign keys:

```
(subject_id) -> [unique] lab.Animal
```

### LINKED ISSUE #491: insert-from-select for 'uninstantiated' tables doesn't work
Although we now support various operations on uninstantiated datajoint classes (e.g. `Foo.insert()`), 
this support does not work for the insert-from-select case. For example:

```Foo.insert(Bar)```

does not work, whereas:

```Foo.insert(Bar())```

does.

dimitri-yatsenko:
Work in progress, provided for preliminary review. Do not merge.

* Fix #301 -- WIP, still need to fix `describe` and `delete`
* Fix #491

### DIFF
--- a/datajoint/__init__.py
+++ b/datajoint/__init__.py
@@ -25,7 +25,7 @@ __all__ = ['__author__', '__version__',
            'config', 'conn', 'kill', 'BaseRelation',
            'Connection', 'Heading', 'FreeRelation', 'Not', 'schema',
            'Manual', 'Lookup', 'Imported', 'Computed', 'Part',
-           'AndList', 'OrList', 'ERD', 'U',
+           'AndList', 'ERD', 'U',
            'DataJointError', 'DuplicateError',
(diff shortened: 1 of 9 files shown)


dimitri-yatsenko:
Do not merge yet.  I still need to update `describe` and `delete`.

dimitri-yatsenko:
This PR will fix #301.

PR #495, opened 2018-10-11
Enable `unique` dependencies

### LINKED ISSUE #301: foreign key properties

### LINKED ISSUE #490: provide 'standard' mechanism for database schema name selection (database.prefix?)
We have been discussing the use of ‘configurable’ database strings for testing purposes in schema development work -

One useful mechanism to use is a ‘database prefix’ - a string which is prepended to the ‘common’ name of a database schema - for example:
```
prefix = ‘my_test_’
dbname = ‘database’
```
-> application code will use: `‘my_test_database’`

In the codebases where this pattern has been applied, we have implemented this mechanism via:
```
schema = dj.schema(dj.config.get('database.prefix', '') + 'database')
```
So, wondering if it would make sense to standardize this mechanism in some way since it is useful for schema development/testing and other uses - either directly in the `dj.schema` logic or some other means (utility function, etc)

### LINKED ISSUE #492: Restriction by a list with an empty list element fails
The following restriction fails:

```python
MyTable() & [ [] ]
```

This is caused by inappropriate parsing of the OrList elements expecting them to be strings, despite the fact that inner `[]` gets translated into `False` boolean value. `or_list` should be first parsed to exclude `False` elements.

### LINKED ISSUE #496: len() and bool() do not work on data class objects
Pretty much all functionality of user table classes can  be performed without explicitly instantiating.  Two exceptions are the length test and the truth test: `__len__` and `__bool__`.

Implement the `len` and `bool` on the class.

dimitri-yatsenko:
* Continue fix #301:  enabled declaration of unique dependencies.
* Fix #496: `len` and `bool` work on user table classes.
* Fix #492: Restrictions by lists containing empty lists work now.


eywalker:
I don't think that issue #490 has reached the stage where it should be implemented and made into master branch. I'll have to reject this PR with that feature included. That should be made into something like a dev branch for others to test. `master` branch should be considered release candidates and experimental features like #490 should not be immediately included.

dimitri-yatsenko:
Then I suggest to cancel #490 altogether.  Any implementation of this feature can be easier implemented on the application side.  This feature does not interact with anything internal to datajoint and it belongs outside.  I am rolling back the changes and I propose to simply close issue #490

eywalker:
I don't think it is quite necessary that we close that issue as it might be worth putting something in - I just don't think it's solidified enough to be made part of this PR. Just revert the changes pertaining to that feature, and we can just continue the discussion on #490.

PR #500, opened 2018-10-12
Fix #482, #494, #498, and #502

### LINKED ISSUE #478: dj.ERD renaming labels increment across subsequent calls
it seems that dj.ERD renamed-attribute label numbers are kept across subsequent calls to dj.ERD

e.g. if on first call, the renamed attribute labels 0,1,2 exist, on the next call there will be 3,4,5 instead, etc

### LINKED ISSUE #482: Query previews in Jupyter issue the same query twice
For some reason, previews in Jupyter notebook issue the query twice, bout the limited preview and the count.

### LINKED ISSUE #494: rename internal classes and modules to comply with current terminology?
Should we rename the internal class and module names to be consistent with terms we use in the documentation?

`relational_operand.py` -> `query.py`
`RelationalOperand` -> `Query`
`base_relation.py` -> `table.py`
`BaseRelation` -> `Table`
`user_relation.py` -> `user_table.py`
`UserRelation` -> `UserTable`
`FreeRelation` -> `FreeTable`

### LINKED ISSUE #498: Describe indexes
Currently, `table.describe` omits any additional indexes  and foreign key properties declared on tables.  Expand the functionality to include indexes and foreign key properties.

### LINKED ISSUE #502: Add Python 3.7 to the Travis test
Currently, we test 3.4, 3.5, and 3.6.  

Let's drop 3.4 testing.

Also see #452.

dimitri-yatsenko:
* Fix #478  -- no numbers show in ERD nodes representing derived dependencies
* Fix #482 -- Save an extra query during table declaration time.
* Fix #494 -- rename classes and modules to correspond to documentation.
* Unfix #496 because IPython calls `bool(cls)` repeatedly in its normal operation, triggering repeated queries.
* Fix #498 -- indexes and properties of dependencies are now properly reverse-engineered by `table.describe()`
* Fix #502 -- Python 3.7 is added to the Travis test


.travis.yml
```diff
@@ -6,6 +6,7 @@ python:
   - "3.4"
   - "3.5"
   - "3.6"
+  - "3.7"
```
eywalker: Replace `- "3.7"` with
```suggestion
# Enable 3.7 without globally enabling sudo and dist: xenial for other build jobs
matrix:
  include:
    - python: 3.7
      dist: xenial
      sudo: true
```

PR #506, opened 2018-10-22
implement projected dependencies (fix #436)

### LINKED ISSUE #436: Relational expressions in foreign key definitions
Foreign keys may be more expressively defined using the same relational operators used in queries.  

Multiple foreign keys made from the primary key, for example, may be viewed as the join:

```
-> Scan
-> Method
```
is equivalent to 
```
-> Scan.proj() * Method.proj()
```
Since only primary keys are used in foreign key references, then the `.proj()` is redundant and we could use simply
```
-> Scan * Method
```

A particular improvement would result for renamed foreign keys.  For example, in the current syntax, to reference to the `User` table as `experimenter, we use the syntax
```
(experimenter) -> User
```
Instead, we could use the `proj` operator for renaming
```
-> User.proj(user="experimenter")
```

This becomes particularly useful for more complex cases.  Users could actually test the expression before making it a foreign key.

For example, let's define a Synapse with references to presynaptic and postsynaptic Cells within a Scan. Currently, this would be declared as
```
# Synapse
-> Scan
(presyn) -> Cell(cell)
(postsyn) -> Cell(cell)
```
Instead it could be defined as 
```
# Synapse
-> Scan * Cell.proj(presyn="cell") * Cell.proj(postsyn="cell")
```
Finally, foreign keys to unions (See #324) use the union operator in the foreign key, resulting in more intricate expressions as detailed in that issue.

This approach uses the same set of concepts as relational expressions and helps consolidate these ideas in the student's mind.

Allowed operators are are join, proj, and union.  Since `aggr` and restriction have no way of affecting the primary key, they are not allowed.  Similarly, `dj.U` is not allowed because it is not referable. 

The change is not backward compatible for renamed foreign keys if we fully commit to the `proj` syntax.  Since renamed foreign keys have not yet gained wide use with less experienced users, I think this is the right time to introduce the change.  We could keep both forms for a while and gradually deprecated the old renaming scheme.

dimitri-yatsenko:
* Fix issue #436, including `describe`
* Backward compatibility is maintained
* Only projections are allowed in dependencies.  Dependencies on more complex expressions should be reissues as separate issues.

### DIFF
--- a/datajoint/declare.py
+++ b/datajoint/declare.py
@@ -16,7 +16,9 @@ HASH_DATA_TYPE = 'char(51)'
 logger = logging.getLogger(__name__)
 
 
-def build_foreign_key_parser():
+def build_foreign_key_parser_old():
+    # old-style foreign key parser. Superceded by expression-based syntax. See issue #436
+    # This will be deprecated in a future release.
     left = pp.Literal('(').suppress()
     right = pp.Literal(')').suppress()
     attribute_name = pp.Word(pp.srange('[a-z]'), pp.srange('[a-z0-9_]'))
@@ -29,6 +31,20 @@ def build_foreign_key_parser():
     ref_table = pp.Word(pp.alphas, pp.alphanums + '._').setResultsName('ref_table')
     ref_attrs = pp.Optional(left + pp.delimitedList(attribute_name) + right).setResultsName('ref_attrs')
     return new_attrs + arrow + options + ref_table + ref_attrs
+
+
+def build_foreign_key_parser():
+    left = pp.Literal('(').suppress()
+    right = pp.Literal(')').suppress()
+    attribute_name = pp.Word(pp.srange('[a-z]'), pp.srange('[a-z0-9_]'))
+    new_attrs = pp.Optional(left + pp.delimitedList(attribute_name) + right).setResultsName('new_attrs')
+    arrow = pp.Literal('->').suppress()
+    lbracket = pp.Literal('[').suppress()
+    rbracket = pp.Literal(']').suppress()
+    option = pp.Word(pp.srange('[a-zA-Z]'))
+    options = pp.Optional(lbracket + pp.delimitedList(option) + rbracket).setResultsName('options')
+    ref_table = pp.restOfLine.setResultsName('ref_table')
+    return new_attrs + arrow + options + ref_table
 
 
 def build_attribute_parser():
@@ -50,6 +66,7 @@ def build_index_parser():
     return unique + index + left + pp.delimitedList(attribute_name).setResultsName('attr_list') + right
 
 
+foreign_key_parser_old = build_foreign_key_parser_old()
 foreign_key_parser = build_foreign_key_parser()
 attribute_parser = build_attribute_parser()
 index_parser = build_index_parser()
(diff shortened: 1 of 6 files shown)


tests/schema.py
```diff
@@ -271,12 +271,13 @@ class DecimalPrimaryKey(dj.Lookup):
...
     contents = zip((0.1, 0.25, 3.99))
 
+
 @schema
 class IndexRich(dj.Manual):
     definition = """
     -> Experiment
     ---
-    (first) ->[unique, nullable] User
+    -> [unique, nullable] User.proj(first="username")
```
eywalker: Actually, given that we are keeping forward compatibility, we need to ensure that old syntax still works. Please add back the old syntax to the test.

tests/schema_advanced.py
```diff
@@ -94,7 +94,7 @@ class Cell(dj.Manual):
 class InputCell(dj.Manual):
     definition = """  # a synapse within the slice
     -> Cell
-    (input)-> Cell(cell)
+    -> Cell.proj(input="cell")
```
eywalker: Same as above

tests/schema_advanced.py
```diff
@@ -108,8 +108,9 @@ class LocalSynapse(dj.Manual):
 
 @schema
 class GlobalSynapse(dj.Manual):
+    # Mix old-style and new-style projected foreign keys
     definition = """
     # a synapse within the slice
-    (pre_slice, pre_cell) -> Cell(slice, cell)
+    -> Cell.proj(pre_slice="slice", pre_cell="cell")
```
eywalker: Same as above

tests/schema.py
dimitri-yatsenko: This kept both syntaxes.  The first dependency uses the old syntax.

dimitri-yatsenko: I did leave some dependencies in the old syntax.  `IndexRich` is a new test, so I used the new syntax.  Old tests use old syntax.

tests/schema_advanced.py
dimitri-yatsenko: While this table uses the new syntax, many existing tests use the old syntax.  They don't show in the diff because they remained unchanged.

tests/schema_advanced.py
dimitri-yatsenko: The current tests cover both syntaxes.

datajoint/declare.py
```diff
@@ -77,16 +90,22 @@ def compile_foreign_key(line, context, attributes, primary_key, attr_sql, foreig
...
+    from .query import Projection
+
+    new_style = True   # See issue #436.  Old style to be deprecated in a future release
     try:
         result = foreign_key_parser.parseString(line)
-    except pp.ParseException as err:
-        raise DataJointError('Parsing error in line "%s". %s.' % (line, err))
+    except pp.ParseException:
+        try:
+            result = foreign_key_parser_old.parseString(line)
```
eywalker: I think it would be best to throw a warning of some kind to indicate that they should update their syntax.

dimitri-yatsenko: Perhaps in a future release after the complete documentation is in place.  How about we make the warning in release 0.12 and raise an exception in 0.13?

eywalker: Ok I'm cool with waiting to throw a warning till 0.12. Let's see about turning it into an exception. It may just be more appropriate to make a release of 1.0.0 where we switch on exception for all warnings...

eywalker:
Must say that the review could be more thorough but will perform thorough follow up testing.

Issue #508, opened 2018-10-25
Suggest to run `dj.config.save_local` after a password change


eywalker:
It so commonly occurs that the user forgets to update their local/global DJ config file upon changing their password in DataJoint. I suggest that `dj.set_password` to prompt along the line of "would you like to update the password entry in the config?" if the config file is present and password entry is not empty.

PR #509, opened 2018-10-25
admin.py: prompt user to update dj_local_conf.json after password update (#508)

ixcat:


datajoint/admin.py
```diff
@@ -14,6 +15,10 @@ def set_password(new_password=None, connection=None):   # pragma: no cover
     connection.query("SET PASSWORD = PASSWORD('%s')" % new_password)
     print('Password updated.')
 
+    if re.match('([Yy]|[Yy][Ee][Ss])', input('update dj_local_conf.json? ')):
```
eywalker: Cool - would you mind adding a keyword argument `prompt=True` or something like that so if set to `False`, it will not prompt to update? I'm thinking of the case we want to use this in a script or in batch where a prompt will be annoying :)

eywalker: Or maybe `update=None` and if it is set to `True` or `False`, it will use that as the response without prompting but if `None` (default), it prompts to udpate?

ixcat: all set

PR #512, opened 2018-10-30
fix bug introduced in last PR #509

dimitri-yatsenko:
In #509, `admin.set_password` attempted to use `dj.config` without importing `dj.config`.


eywalker:
Ok I'm to be blamed for this. Thanks for fixing @dimitri-yatsenko

PR #514, opened 2018-10-30
Fix #511: prohibit direct inserts outside the make callback

### LINKED ISSUE #511: Prohibit inserts from outside `make` in auto-populated tables.
Auto-populated tables should not permit inserts from outside their `make` method.   Raise an error when the user attempts to do so.

dimitri-yatsenko:


eywalker:
I'm not entirely sure if I agree about adding this, at least without some discussions. I completely understand where this is coming from, but I want to make sure we flesh out all imaginable cases before we put in what would be a fairly strong restriction.

dimitri-yatsenko:
This feature is introduced due to a user issue when they did not understand how auto-populate worked.  They first inserted into master and then called `populate` to have the parts populated.  It took me a while to catch the bug because I did not imagine someone inserting into a computed table.

This feature is important for new users to understand how to properly use DataJoint.

If the user must insert directly, they can set `allow_direct_insert=True` to bypass the restriction.

eywalker:
Ok makes sense and ability to bypass was something I was going to request. Ok - I'd still like to think about this before I will fully accept. I also would prefer that you would create an issue describing this so that this PR can be considered as a fix for that issue.

dimitri-yatsenko:
It is a fix of issue #511

eywalker:
Ah I oversaw that.

datajoint/table.py
```diff
@@ -172,6 +174,16 @@ def insert(self, rows, replace=False, skip_duplicates=False, ignore_extra_fields
             warnings.warn('Use of `ignore_errors` in `insert` and `insert1` is deprecated. Use try...except... '
                           'to explicitly handle any errors', stacklevel=2)
 
+        # prohibit direct inserts into auto-populated tables
+        try:
+            allow = allow_direct_insert or self._allow_insert  # only present in AutoPopulate
```
eywalker: Can you just replace the `self._allow_insert` with `getattr(self, '_allow_insert', True)` instead of using try except block? I much prefer a logic that doesn't rely on raising exceptions when not necessary.

dimitri-yatsenko: agreed. Done

eywalker: Awesome

Issue #483, opened 2018-08-17
order of attributes does not respect primary key first convention in projection results


spapa013:
A projection query like this
```python
nda2.Trace.proj('trace', presyn='segment_id') & key
```
gives

```
*scan_idx    *segmentation  trace      *presyn   
+----------+ +------------+ +--------+ +--------+
5            1              =BLOB=     421208    
 (1 tuples)
```

so that primary keys are not grouped at the beginning. Would make sense for all table results to consistently give primary key attributes at the beginning.

dimitri-yatsenko:
Python 3.6+?

spapa013:
yes

Issue #516, opened 2018-11-08
Renaming foreign key attributes shuffles the properties of the table attributes


ecobost:
With these declarations:
```python3
@schema
class A(dj.Manual):
     definition="""
     my_int: int
     my_string: varchar(64)
     ---
     """

@schema
class B(dj.Manual):
     definition = """
     -> A.proj(renamed_int='my_int')
     ---
     """

@schema
class C(dj.Manual):
    definition = """
    (renamed_int) -> A(my_int)
    ---
    """
```
I get this output:
```python3
In [2]: import datajoint as dj                                                                           

In [3]: dj.__version__                                                                                   
Out[3]: '0.11.0'

In [4]: from pipeline import stack     # this is where I defined the classes above
Loading local settings from pipeline_config.json
Connecting ecobost@at-database.ad.bcm.edu:3306

In [5]: stack.A.heading                                                                                  
Out[5]: 
# 
my_int               : int                          # 
my_string            : varchar(64)                  # 

In [6]: stack.B.heading                                                                                  
Out[6]: 
# 
my_string            : int                          # 
renamed_int          : varchar(64)                  # 

In [7]: stack.C.heading                                                                                  
Out[7]: 
# 
my_string            : int                          # 
renamed_int          : varchar(64)                  # 


```

Projecting sends the renamed variable to the end of the queue but the attribute properties are not shuffled accordingly so, for instance, if I have three attributes p1:int, p2:boolean, p3:varchar and rename p2->p2' I will get p1:int, p3:boolean, p2':varchar.

Cheers,

dimitri-yatsenko:
looks serious.  Will fix today.

dimitri-yatsenko:
Closing via #517

Issue #515, opened 2018-11-01
Preview Error


FedeClaudi:
Hi, 

I'm running DataJoint with Python3 on Windows with a locally hosted server. 
I reliably get an error when trying to inspect the content of a class:

Example class definition:

```
@schema
class BehaviourRecording(dj.Manual):
     definition = """
     # Individual recordings within a session
     -> Session
     recording_number: int               # recording number within a session
     ---
     metadata_path : varchar(128)       # name of the experimenter
     video_path : varchar(128)       # name of the experimenter
     tracked : enum('Y', 'N')   # asdas
     dlc_data: varchar(128)      # path to .h5 produced as a result of dlc tracking
     """
```

Example error:
```
recordings = BehaviourRecording()
recordings.preview()
Traceback (most recent call last):
  File "C:\Users\Federico\Miniconda3\envs\CVenv\lib\site-packages\IPython\core\interactiveshell.py", line 3265, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-10-5ee312724859>", line 1, in <module>
    recordings.preview()
  File "C:\Users\Federico\Miniconda3\envs\CVenv\lib\site-packages\datajoint\relational_operand.py", line 362, in preview
    tuples = rel.fetch(limit=limit+1)
  File "C:\Users\Federico\Miniconda3\envs\CVenv\lib\site-packages\datajoint\fetch.py", line 59, in __call__
    cur = self._relation.cursor(as_dict=as_dict, limit=limit, offset=offset, order_by=order_by)
  File "C:\Users\Federico\Miniconda3\envs\CVenv\lib\site-packages\datajoint\relational_operand.py", line 530, in cursor
    return self.connection.query(sql, as_dict=as_dict)
  File "C:\Users\Federico\Miniconda3\envs\CVenv\lib\site-packages\datajoint\connection.py", line 136, in query
    cur.execute(query, args)
  File "C:\Users\Federico\Miniconda3\envs\CVenv\lib\site-packages\pymysql\cursors.py", line 170, in execute
    result = self._query(query)
  File "C:\Users\Federico\Miniconda3\envs\CVenv\lib\site-packages\pymysql\cursors.py", line 328, in _query
    conn.query(q)
  File "C:\Users\Federico\Miniconda3\envs\CVenv\lib\site-packages\pymysql\connections.py", line 515, in query
    self._execute_command(COMMAND.COM_QUERY, sql)
[...]

PR #517, opened 2018-11-13
Fix issues #483 and #516

dimitri-yatsenko:
This pull request follows PR #514

* Update `.proj` to preserve the original order of attributes 
* Renames `Query` -> `Expression` and `query.py` -> `expression.py`

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -46,4 +46,4 @@ def teardown_package():
     cur = conn.query('SHOW DATABASES LIKE "{}\_%%"'.format(PREFIX))
     for db in cur.fetchall():
         conn.query('DROP DATABASE `{}`'.format(db[0]))
-    conn.query('SET FOREIGN_KEY_CHECKS=1')
+    conn.query('SET FOREIGN_KEY_CHECKS=1').close()
--- a/tests/test_relational_operand.py
+++ b/tests/test_relational_operand.py
@@ -2,7 +2,7 @@ import random
 import string
 
 import numpy as np
-from nose.tools import assert_equal, assert_false, assert_true, raises, assert_set_equal
+from nose.tools import assert_equal, assert_false, assert_true, raises, assert_set_equal, assert_list_equal
 
 import datajoint as dj
 from .schema_simple import A, B, D, E, L, DataA, DataB, TestUpdate, IJ, JI, ReservedWord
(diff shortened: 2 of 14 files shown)


tests/test_relational_operand.py
```diff
@@ -13,12 +13,12 @@ def setup():
     """
     module-level test setup
     """
-    A().insert(A.contents, skip_duplicates=True)
-    L().insert(L.contents, skip_duplicates=True)
-    B().populate()
-    D().populate()
-    E().populate()
-    Experiment().populate()
+    A.insert(A.contents, skip_duplicates=True)
```
eywalker: We should mix in instance based inserts as that's supposed to work too. In general, please do not remove test cases but rather just add.

tests/__init__.py
```diff
@@ -46,4 +46,4 @@ def teardown_package():
     cur = conn.query('SHOW DATABASES LIKE "{}\_%%"'.format(PREFIX))
     for db in cur.fetchall():
         conn.query('DROP DATABASE `{}`'.format(db[0]))
-    conn.query('SET FOREIGN_KEY_CHECKS=1')
+    conn.query('SET FOREIGN_KEY_CHECKS=1').close()
```
eywalker: Are you sure that addition of `.close()` is a safe operation? The connection object is largely shared throughout the test so we have to ensure that it doesn't get closed in the middle.

dimitri-yatsenko: This does not close the connection but the cursor.  However, I experimented with closing the cursor after every query to attempt to address #515 but I do not think that was the problem. I took all `.close()` out but forgot this one.

tests/test_relational_operand.py
dimitri-yatsenko: Yes, we have a good mixture.  If you review the tests, the majority of tests still instantiate the classes.

tests/__init__.py
dimitri-yatsenko: Each query creates its own cursor so closing the cursor should not make any difference

tests/test_relational_operand.py
eywalker: As in every operation that's supposed to work for either instance or class ought to be tested on both. And it still holds that in general you shouldn't just be replacing previous test cases if they are supposed to continue to work.

tests/__init__.py
eywalker: Fair

tests/test_relational_operand.py
dimitri-yatsenko: This is not meant for testing, just the regular setup.  It works.

dimitri-yatsenko: We do have testing for both instances and classes.

Issue #205, opened 2016-04-26
Table names less then 64 characters


fabiansinz:
There is a [64 character limit](http://dev.mysql.com/doc/refman/5.0/en/identifiers.html) on database and table names. DataJoint should check that the table name doesn't get longer than 64 characters. This might be a problem with  `Part` tables.

dimitri-yatsenko:
What's the message that SQL throws?  Is it not informative enough?

PR #529, opened 2018-12-01
Fix issue #205 - check table length before declaring

dimitri-yatsenko:


datajoint/declare.py
```diff
@@ -196,6 +197,10 @@ def declare(full_table_name, definition, context):
     :param definition: DataJoint table definition
     :param context: dictionary of objects that might be referred to in the table.
     """
+
+    table_name = full_table_name.strip('`').split('.')[1]
+    if len(table_name) > MAX_TABLE_NAME_LENGTH:
+        raise DataJointError('Table name `%s` is too long')
```
eywalker: I think it'll be helpful to show what is the maximum table length along with this error message.

eywalker:
This PR depends on #528 - will only accept upon that PR accepted.

datajoint/declare.py
dimitri-yatsenko: fixed

Issue #519, opened 2018-11-18
Populate with `reserve_jobs=True` doesn't work if `datetime` is a primrary key attribute


eywalker:
Related to #428. Blob doesn't support handling `datetime` field and as a result, inserting a key that contains datetime field fails.

dimitri-yatsenko:
fixed in #530

PR #530, opened 2018-12-02
Fix #279, 428, and 519

### LINKED ISSUE #279: Auto reconnect to database after connection timeout
Currently when connection gets timed out (say because you left the session sitting for some time), you have to explicitly reconnect using `dj.conn().connect()`. It would be far nicer if DataJont would automatically attempt a reconnect prior to attempting to issue another query.

We should be careful about auto reconnect if disconnect occurs in the middle of a transaction. When disconnect occurs, transaction is automatically rolled back on the database server, but we should also make sure that the system *do not* reconnect and attempt to resume the transaction process in the middle. If disconnect is detected in a middle of a transaction process, error should be issued to kick process out of a transaction process (e.g. `populate`). Prior to issuing any additional query, if the process is *not* in a transaction, then auto-reconnect can occur safely.

### LINKED ISSUE #428: Use of `decimal` type in primary key prevents `reserve_jobs=True`
Currently insertion of `decimal` datatype into blob is not supported. Because of this, when you have a computed table with `decimal` primary key attribute, `populate(reserve_jobs=True)` fails due to the failure of inserting the primary key dictionary (containing the decimal key attribute value) into the `jobs` table `key` blob field. Worse yet, when this error occurs, this occurs silently, such that it appears as if the whole `populate` jobs completed very quickly.

eywalker:
* Fix #279 - Connection to the database is automatically recovered. If the connection was in a transaction, this triggers a `DataJointError`
* Fix #428 and #519 - Packing of `Decimal` and `Datetime` objects into a blob is now supported by converting them into supported datatypes of float and string, respectively.

### DIFF
--- a/datajoint/connection.py
+++ b/datajoint/connection.py
@@ -156,7 +156,7 @@ class Connection:
                     logger.debug("Re-executing SQL")
                     cur = self.query(query, args=args, as_dict=as_dict, suppress_warnings=suppress_warnings, reconnect=False)
             else:
-                logger.debug("Caught InterfaceError/OperationalError that are not related to connection issue")
+                logger.debug("Caught InterfaceError/OperationalError.")
                 raise
         except err.ProgrammingError as e:
             if e.args[0] == server_error_codes['parse error']:
--- a/datajoint/connection.py
+++ b/datajoint/connection.py
@@ -116,7 +116,7 @@ class Connection:
         except:
             return False
 
-    def query(self, query, args=(), as_dict=False, suppress_warnings=True, reconnect=True):
+    def query(self, query, args=(), as_dict=False, suppress_warnings=True, reconnect=None):
         """
         Execute the specified query and return the tuple generator (cursor).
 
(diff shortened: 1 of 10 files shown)


datajoint/connection.py
```diff
@@ -134,16 +145,18 @@ def query(self, query, args=(), as_dict=False, suppress_warnings=True):
...
-                logger.debug("Re-executing SQL: " + query[0:300])
-                cur.execute(query, args)
+                if self._in_transaction:
+                    self.cancel_transaction()
+                    raise DataJointError("Connection was lost during a transaction.")
+                else:
+                    logger.debug("Re-executing SQL")
+                    cur = self.query(query, args=args, as_dict=as_dict, suppress_warnings=suppress_warnings, reconnect=False)
             else:
+                logger.debug("Caught InterfaceError/OperationalError that are not related to connection issue")
```
dimitri-yatsenko: This could be a connection issue.

Issue #497, opened 2018-10-11
Configurable blobs


dimitri-yatsenko:
After some discussions, we have decided to change how "external storage" is managed.   Instead of the `external` datatype, we will have `blob-*` datatypes.  For backward compatibility, `external` and `external-*` datatypes will continue to be supported.  

The main advantage is that `blob-*` can be configured to be regular blob datatype, stored in the database or stored in external bulk storage.  So the immediately supported storage protocols will be `table`, `file`, and `s3`.

dimitri-yatsenko:
Here are examples of syntax:
```python
dj.config['store'] = dict(protocol='blob', compress='zlib')
dj.config['store-images'] = dict(protocol='s3', location='/mybucket/store/', secretkey='')
dj.config['store-raw'] = dict(protocol='file', location='/store/', folding=(2,2))
dj.config['store-shared'] = dict(protocol='s3', folding=(1,2))

dj.config['attach'] = dict(store='store')
dj.config['attach-scope'] = dict(store='store-images')
dj.config['blob-images'] = dict(store='store-images')
```


Internally, the DataJoint attribute declaration 
```
field : blob-   # comment
```
is translated into SQL column declaration
```
field longblob NOT NULL COMMENT "blob-:commment"
```

and
```
field : blob-shared
```
into
```sql
field char(51)  COMMENT "blob-shared:"
```
and
```
field : attach-shared
```
into
```sql
field char(51)  COMMENT "attach-shared:comment"
```

dimitri-yatsenko:
Fixed with #532

Issue #480, opened 2018-08-06
File attachments


dimitri-yatsenko:
# Purpose

Enable storage of entire files intact, without having to read and serialize them, similar to attachments.  Enable the use of these attachments through GUI drag-and-drop features.  The attached files are not browsable outside DataJoint.  DataJoint must maintain data integrity.

# Operation

Attachments use a mechanism similar to external storage using file hashes as the filenames in a storage repository.   The datatype should be either attach or attach-suffix where the suffix can be up to eight character long.  The storage protocol and location (filesystem, S3, globus) is specified similar to external storage. 

The table `~attach` is created and maintained similar to how `~external` is used for external storage.  The only difference is that it also stores the original filename of the stored file.

Just as with external, this mechanism ensures that inserting the same file multiple times within the same schema does not result in duplicated storage.

## Insert 
Insert works by inserting the full path to the file to be inserted.  The insert is asynchronous and commits the insert into the table into ~attach.
In its initial implementation, the entire file will be first read into memory for hashing and writing into storage. 

## Fetch
Fetch downloads the attached files.  The download will be implemented asynchronously in the initial implementation.  While the file is being downloaded, it is given the extension .download.   When the download is complete the file is renamed into its target name.

The destination folder can be specified in the configuration.  That can be overridden by the optional argument download_path of fetch.  If neither is identified the download is performed into the current directory. 

In all cases, the full path to the downloaded file is returned in the fetch output as attribute values.

Similar to external storage, fetch can be configured to use local caching. 

If the same file retrieved multiple time as part of the same fetch, the file is only retrieved once and the same path in returned for each requested attribute value.

If the file already exists, the downloaded file is given an incrementing suffix.

## Delete

Similar to external, delete only operates on the host tables and does not remove entries from ~attach or from the file repository.

Cleanup of ~attach and of storage folder work identically to the corresponding mechanisms of external storage.

Cleanup commands:
```python
schema.attachment.delete_garbage()   # transaction-safe
schema.attachment.cleanup_store()   # not transaction-safe
```

dimitri-yatsenko:
Attachments will use the existing blob functionality and will be configured the same way.  Thus fields of type `attach` will use the configuration from `dj.config['blob']` and `attach-images` will use `dj.config[blob-images]`.  The internal implementation will use the same storage tracking table `~external` for both blobs and attachments.

The insert will take filepaths and will store the  files as binary strings along with the filename. 

[...]

Issue #475, opened 2018-07-23
subfolders for external storage


dimitri-yatsenko:
To avoid millions of files in one folder, break up external storage files into subfolders.

1. Add subfolder scheme to the external storage configuration in `dj.config`.  The structure may be specified as a tuple, e.g. (2,2,2), which indicates three levels of subfolders, each taking the next 2 characters of the hash and leaving the rest for the filename.  The default scheme is (), i.e. no subfolders.

2.  When creating a new external storage folder, create the file `djexternal.json` and save the subfolder scheme.  

3. When using an existing external storage folder, find `djexternal.json` and read the subfolder scheme from it, ignoring the one specified in `dj.config`.

4. Break up the hash filename according to the subfolder scheme for both reading and writing.

dimitri-yatsenko:
@ixcat  Are you about to submit a solution for this?

ixcat:
@dimitri-yatsenko 'about' is somewhat relative - it's on my active todo list about 3 items down but I haven't begun work. expecting to begin tomorrow.

dimitri-yatsenko:
ok, instead of of `djexternal.json`, let's call it `djstore.json`

dimitri-yatsenko:
We revised this as specified in #480.  The configuration will contain the field `folding` for each store.  Folding will be a sequence of numbers indicating the folding, e.g. `(2,2)` means folding as `
`/ab/cd/abcdefghiklmno`

dimitri-yatsenko:
Does it make sense to have folding for S3 storage or only for files?  It seems like S3 would not care.

eywalker:
it would make translating between S3 and filesystems simpler if folding is supported regardless of the system. I don't see a reason to not support folding for S3

dimitri-yatsenko:
Fixed with #577

Issue #467, opened 2018-04-25
Interpret None as NULL in inserts


ixcat:
In tables with `NULL` default values, insert behavior is inconsistent:

If a dictionary-style insert is performed using a dictionary without a key corresponding to the default-null field, the operation proceeds. In the case of a dictionary-style insert where the default-null field is present and set to `None`, or in the case of a list-style insert with an appropriately given positional `None`,  the operation fails. 

However, on the fetch side, records are returned with `None` as a present and populated value.

```
>>> print(OtherNull.definition)

    otherid: integer
    ---
    otherthing=NULL: varchar(255)
    
>>> OtherNull().insert1({'otherid': 1})
>>> OtherNull().insert1({'otherid': 2, 'otherthing': None})
Traceback (most recent call last):
  File "<console>", line 1, in <module>
  File "/Users/chris/Workspace/mlims-pipeline.ve/lib/python3.6/site-packages/datajoint/base_relation.py", line 137, in insert1
    self.insert((row,), **kwargs)
  File "/Users/chris/Workspace/mlims-pipeline.ve/lib/python3.6/site-packages/datajoint/base_relation.py", line 278, in insert
    itertools.chain.from_iterable((v for v in r['values'] if v is not None) for r in rows)))
  File "/Users/chris/Workspace/mlims-pipeline.ve/lib/python3.6/site-packages/datajoint/connection.py", line 133, in query
    cur.execute(query, args)
  File "/Users/chris/Workspace/mlims-pipeline.ve/lib/python3.6/site-packages/pymysql/cursors.py", line 163, in execute
    query = self.mogrify(query, args)
  File "/Users/chris/Workspace/mlims-pipeline.ve/lib/python3.6/site-packages/pymysql/cursors.py", line 142, in mogrify
    query = query % self._escape_args(args, conn)
TypeError: not enough arguments for format string
>>> OtherNull().insert1((2, None,))
Traceback (most recent call last):
  File "<console>", line 1, in <module>
  File "/Users/chris/Workspace/mlims-pipeline.ve/lib/python3.6/site-packages/datajoint/base_relation.py", line 137, in insert1
    self.insert((row,), **kwargs)
  File "/Users/chris/Workspace/mlims-pipeline.ve/lib/python3.6/site-packages/datajoint/base_relation.py", line 278, in insert
    itertools.chain.from_iterable((v for v in r['values'] if v is not None) for r in rows)))
[...]

PR #532, opened 2018-12-03
Attachments and configurable blobs

dimitri-yatsenko:
* Nans are interpreted as NULL for floating-point attributes at insert #467
* implement configurable subfolders in external storage #475 
* implement file attachment datatype #480 
* implement configurable blobs #497 
* add `dj.get_schema_names()`


datajoint/__init__.py
```diff
@@ -15,29 +15,27 @@
 """
 
 __author__ = "Dimitri Yatsenko, Edgar Y. Walker, and Fabian Sinz at Baylor College of Medicine"
-__date__ = "Nov 15, 2018"
+__date__ = "January 14, 2018"
```
austin-hilberg: Do you mean 2019?

datajoint/declare.py
```diff
@@ -291,32 +291,41 @@ def compile_attribute(line, in_key, foreign_key_sql):
...
+    if is_configurable:
         if in_key:
-            raise DataJointError('External attributes cannot be primary in:\n%s' % line)
+            raise DataJointError('Configurable attributes cannot be primary in:\n%s' % line)
+        match['comment'] = ':{type}:{comment}'.format(**match)  # insert configurable type into comment
         store_name = match['type'].split('-')
-        if store_name[0] != 'external':
-            raise DataJointError('External store types must be specified as "external" or "external-<name>"')
+        if store_name[0] not in ('external', 'blob', 'attach'):
+            raise DataJointError('Invalid attribute type in:\n%s' % line)
```
austin-hilberg: The previous error message gave a more explicit indication of what to do to fix the problem. I find the new one less informative.

datajoint/erd.py
```diff
@@ -50,8 +50,8 @@ class ERD:
         """
 
         def __init__(self, *args, **kwargs):
-            warnings.warn('ERD functionality depends on matplotlib and pygraphviz. Please install both of these '
-                          'libraries to enable the ERD feature.')
+            warnings.warn('ERD functionality depends on matplotlib, networkx, and pygraphviz. '
+                          'Please install both of these libraries to enable the ERD feature.')
```
austin-hilberg: There are now three libraries mentioned.

tests/test_legacy_external_class.py
```diff
@@ -0,0 +1,63 @@
+from nose.tools import assert_true, assert_list_equal, raises
+from numpy.testing import assert_almost_equal
+import datajoint as dj
+from . import schema_legacy_external as modu
+
+
+def test_heading():
+    heading = modu.Simple().heading
```
austin-hilberg: You removed the `()` in line 8 of `test_external_class.py` above. Should you do so here as well?

tests/test_legacy_external_class.py
```diff
@@ -0,0 +1,63 @@
...
+    assert_true('item' in heading)
+    assert_true(heading['item'].is_external)
+
+
+def test_insert_and_fetch():
+    original_list = [1, 3, 8]
+    modu.Simple().insert1(dict(simple=1, item=original_list))
```
austin-hilberg: Same as above; should the `()` be removed throughout this method?

tests/test_legacy_external_class.py
```diff
@@ -0,0 +1,63 @@
...
+    q = (modu.Simple() & {'simple': 1}).fetch1()
+    assert_list_equal(list(q['item']), original_list)
+    # test with cache
+    dj.config['cache'] = previous_cache
+    q = (modu.Simple() & {'simple': 1}).fetch1()
+    assert_list_equal(list(q['item']), original_list)
+
+
+def test_populate():
+    image = modu.Image()
```
austin-hilberg: Can we remove these `()` throughout this and the next two methods?

tests/test_legacy_external_class.py
dimitri-yatsenko: This does not really matter.  Accessing attributes of classes directly still triggers instantiation of an object, so a more efficient way is to instantiate.  The omission of parentheses was provided for user convenience.

tests/test_legacy_external_class.py
dimitri-yatsenko: Yes, we could but, when using the same table multiple times, it's more efficient to instantiate once.

dimitri-yatsenko: This particular snippet is inside `test_legacy_external_class`, which tests old functionality, so it's a good place to test whether the old syntax still works.

tests/test_legacy_external_class.py
dimitri-yatsenko: This is inside legacy tests.  We support both syntaxes.

datajoint/erd.py
dimitri-yatsenko: fixed

datajoint/version.py
```diff
@@ -1 +1 @@
-__version__ = "0.11.1"
+__version__ = "0.12.0"
```
eywalker: Please change this to `"0.12.dev"`

PR #534, opened 2018-12-03
Add pandas support and support order_by "KEY" (issues #459, #537, #538, #541)

### LINKED ISSUE #459: Add pandas.DataFrame methods
Ever since we've started using datajoint, I've been frustrated by how hard it is to access data in a table. I often just want to get a few rows and/or columns so I can examine some of the data, and to do that, I don't want to have to type a complicated restriction followed by a fetch. I've added some convenience methods/properties to a mix-in class that we inherit from for all of our tables. Here's a quick description of some of them:
- `df`: fetches the whole table, returns it in a DataFrame - this can be slow for big tables
- `head`: fetches just the first `n=25` tuples in the table, returns it in a DataFrame - very fast
- `tail`: fetches just the last `n=25` tuples in the table, returns it in a DataFrame - slower than `head`
- `full`: prints out the entire table in a DataFrame, without skipping over any rows

Once in a DataFrame, the data is trivial to get at using direct column and row indexing. Another benefit is that the DataFrame does a nicer job of displaying as much data as possible in the given screen space.

The code for the above is quite trivial. Would there be any interest in integrating these into datajoint? I'd be happy to do it.

(Sorry for not using the proper database terminology, but the terms "table", "row" and "column" are much more intuitive to most users than "relation", "tuple" and "attribute").

### LINKED ISSUE #537: `pandas` support
Add comprehensive support for `pandas` for represent datajoint entity sets.
1. `pandas` will become a required package for datajoint
2.  `dj.config` will provide the option `fetch_format` to specify the default format for `fetch` output: `pandas.DataFrame` or `numpy.rec_array` or list of `dict`s.
3.  `fetch` will support returning pandas dataframes when configured. `fetch` will provide a keyword argument to select the output format.  The returned dataframe will be indexed by the primary key.
4.  Restriction operator will accept `pandas` as an argument as in `table & df`
5.  `insert` will accept a pandas dataframe as input for inserting.
6.  Add the methods `head` and `tail` as a shortcut to fetch the first and the last few elements (ordered by primary key) from a query expressions. The result will be in the default format specified in `dj.config`.

This issue subsumes issue #459.

### LINKED ISSUE #538: Reword `n tuples` at the end of query previews.
Currently, query previews report the total number of entries as, for example,
```
919 tuples
```
The word `tuples` comes from the pure relational model and not DataJoint's more refined model.

We could replace it with, for example,
```
Total: 919
```

### LINKED ISSUE #541: Enable order_by the primary key in fetch.
I propose to add the ability to order by primary key by supplying the attributes "KEY" or "KEY DESC" in the `order_by` list.  MySQL already sorts by the primary key in simple cases but it may not hold in more complex queries.

dimitri-yatsenko:
* Fix #459 - pandas support
* Fix #537 - pandas support revised
* Fix #538 - "n tuples" -> "Total: n"
* Fix #541 - Allow order_by by "KEY" and "KEY DESC"
* add `expr.head()` and `expr.tail()` shortcuts for limited key-sorted fetches


dimitri-yatsenko:
@eywalker What is your verdict?  Perhaps we could leave this feature undocumented until we have a more comprehensive strategy of working with pandas, including inserts from pandas frameworks.

I would like to group these methods a bit more so that they clearly stand out when listing object methods.  Perhaps this could be done by encapsulating them into a single attribute, e.g. `df`:
```python
table.df(**kwargs)   # fetch as a pandas frame 
table.df.get_head(n=20)   
table.df.head    # shortcut to fetch head
table.df.tail    # shortcut to fetch tail
table.df.get_tail(n=20)     # fetch tail of specified length
table.df.tail    # shortcut to fetch tail
```

Thoughts?

dimitri-yatsenko:
Insert can be implemented straightforwardly without requiring installing pandas by duck-typing calling `to_records` on the inserted object and proceeding to insert normally.

mspacek:
> I would like to group these methods a bit more so that they clearly stand out when listing object methods. Perhaps this could be done by encapsulating them into a single attribute, e.g. `pd`:
> 

That makes sense organizationally, but then it's less convenient, which starts to defeat the purpose for me.

dimitri-yatsenko:
@eywalker and I discussed this in person and decided that datajoint should support `pandas` more extensively and we will make it part of this pull request.  This solution is described in #537.  This is a more principles and comprehensive solution than currently proposed in this PR.

dimitri-yatsenko:
Okay, @mspacek and @eywalker  I have implemented pandas support as discussed.  

You can now do this:
```python
dj.config['fetch_format'] = 'frame'
Table.fetch()  # returns a pandas.DataFrame
Table.head()  # return the first 25 entries as a pandas.DataFrame ordered by primary key
Table.tail()  # return the last 25 entries as a pandas.DataFrame ordered by primary key
dj.config['fetch_format'] = 'array'
Table.fetch()  # returns a numpy.record_array
Table.head()  # return the first 25 entries as a numpy.record_array ordered by primary key
Table.tail()  # return the last 25 entries as a a numpy.record_array ordered by primary key
Table.head(format="frame")  # now as a pandas.DataFrame
```

Please review this before I add tests.

ixcat:
I think this is great -

there was some question r.e. if pandas should/should not be included as dependency -
I just remembered we don't actually require matplotlib/pydotplus for ERD support, they
are listed in 'test_requirements.txt' - if this concern is still there, perhaps this should be the mechanism for 'optional' extensions, including this?

either way, just recalled this mechanism in use in erd.py so thought I'd mention.

mspacek:
@dimitri-yatsenko that seems very flexible, which is cool. But my concern is that by default when we (our lab) do a fetch, we want to get the usual numpy record array, yet we also want to have quick access to the table (or its head or tail) as a pandas DataFrame. The first is for actually doing stuff with the data, the second is typically just for inspection. Doing so with these proposed changes would require constantly changing `dj.config['fetch_format'] `, which is inconvenient. Just my opinion. Maybe our use case isn't typical.

dimitri-yatsenko:
@mspacek I believe the current solution accomplishes what you need but in a more general way. 

We now provide methods `.tail()` and `.head()`, which are shortcuts for `.fetch(order_by='KEY', limit=25)` and `.fetch(order="KEY DESC", limit=25)`, respectively.  You can also change the format for any `fetch` call, including `head` and `tail` using the `format` keyword.  Sounds like your preference is to have different default formats for `fetch` and for `head` and `tail`.

datajoint/expression.py
```diff
@@ -342,6 +346,26 @@ def fetch1(self):
...
+        :param limit:  number of entries
+        :param fetch_kwargs: kwargs for fetch
+        :return: query result
+        """
+        return self.fetch(order_by="KEY", limit=limit, **fetch_kwargs)
+
+    def tail(self, limit=25, **fetch_kwargs):
+        """
+        shortcut to fetch the last few entries from query expression.
+        Equivalent to fetch(order_by="KEY DESC", limit=25)
```
eywalker: not quite - it's equivalent to `fetch(order_by="KEY DESC", limit=25)[::-1]`

datajoint/fetch.py
```diff
@@ -24,6 +27,16 @@ def to_dicts(recarray):
         yield dict(zip(recarray.dtype.names, rec.tolist()))
 
 
+def _flatten_attribute_list(expression, attr):
```
eywalker: I'll find this function a bit more general and useful (while I know it's a private function) if it takes in list of primary key as the first argument instead of the `expression` object. The only reason we are passing in `expression` is to grab `primary_key`.

datajoint/fetch.py
```diff
@@ -33,36 +46,59 @@ class Fetch:
...
-                                 'Use proj() to select attributes or set as_dict=False')
+                                 'Use '
+                                 'proj() to select attributes or set as_dict=False')
+        # format should not be specified with attrs or is_dict=True
+        if format is not None and (as_dict or attrs):
+            raise DataJointError('Cannot specify output format when as_dict=True or '
+                                 'when attributes are selected to be fetched separately.')
+
+        if format not in {None, "array", "frame"}:
+            raise DataJointError('Fetch output format must be in {"array", "frame"}')
```
eywalker: I always appreciate error message that tells me the wrong value I gave. So could we modify this to say 

```suggestion
            raise DataJointError('Fetch output format must be in {{"array", "frame"}} but "{}" was given'.format(format))
```

datajoint/fetch.py
```diff
@@ -33,36 +46,59 @@ class Fetch:
...
+
+        if not (attrs or as_dict) and format is None:
+            format = config.get('fetch_format', 'array')  # default to array
+            if format not in {"array", "frame"}:
+                raise DataJointError('Invalid entry in datajoint.config["fetch_format"]: use "array" or "frame"')
```
eywalker: Similar to above:

```suggestion
                raise DataJointError('Invalid entry "{}" in datajoint.config["fetch_format"]: use "array" or "frame"'.format(format))
```

datajoint/expression.py
dimitri-yatsenko: right

datajoint/fetch.py
```diff
@@ -33,36 +46,59 @@ class Fetch:
...
+        # format should not be specified with attrs or is_dict=True
+        if format is not None and (as_dict or attrs):
+            raise DataJointError('Cannot specify output format when as_dict=True or '
+                                 'when attributes are selected to be fetched separately.')
+
+        if format not in {None, "array", "frame"}:
+            raise DataJointError('Fetch output format must be in {"array", "frame"}')
+
+        if not (attrs or as_dict) and format is None:
+            format = config.get('fetch_format', 'array')  # default to array
```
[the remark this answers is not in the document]
dimitri-yatsenko: right.  I forgot about that.

datajoint/fetch.py
dimitri-yatsenko: right.

datajoint/fetch.py
dimitri-yatsenko: agreed.

PR #552, opened 2018-12-28
datajoint/table.py: add 'see allow_direct_insert' to direct_insert error message

ixcat:
Add clarification about how to override the direct_insert restriction.

... not sure on precise text since the idea of adding this flag was to protect data purity 
& don't want to encourage blind usage, but at the same time, pointers might be useful..

### DIFF
--- a/datajoint/table.py
+++ b/datajoint/table.py
@@ -170,7 +170,7 @@ class Table(QueryExpression):
         # prohibit direct inserts into auto-populated tables
         if not (allow_direct_insert or getattr(self, '_allow_insert', True)):  # _allow_insert is only present in AutoPopulate
             raise DataJointError(
-                'Auto-populate tables can only be inserted into from their make methods during populate calls.')
+                'Auto-populate tables can only be inserted into from their make methods during populate calls. (see allow_direct_insert)')
 
         heading = self.heading
         if inspect.isclass(rows) and issubclass(rows, QueryExpression):   # instantiate if a class


datajoint/table.py
```diff
@@ -170,7 +170,7 @@ def insert(self, rows, replace=False, skip_duplicates=False, ignore_extra_fields
         # prohibit direct inserts into auto-populated tables
         if not (allow_direct_insert or getattr(self, '_allow_insert', True)):  # _allow_insert is only present in AutoPopulate
             raise DataJointError(
-                'Auto-populate tables can only be inserted into from their make methods during populate calls.')
+                'Auto-populate tables can only be inserted into from their make methods during populate calls. (see allow_direct_insert)')
```
dimitri-yatsenko:
```suggestion
                'Auto-populate tables refuse direct inserts from outside their make methods during populate calls, unless overridden with the allow_direct_insert argument.')
```

ixcat: Hmm.. I see combining the two, but think the original 'error' description to me seems more straightforward- what do you think about:

```
Auto-populate tables can only be inserted into from their make methods during populate calls, unless overridden with the allow_direct_insert argument.```

or

```Auto-populate tables can only be inserted into from their make methods during populate calls. To override, use the the allow_direct_insert argument.```

was thinking something like this at 1st but deliberately was a bit obscure so as to not 'encourage' bad behavior by users looking for a 'quick fix' to the issue..

will plan to go with your choice of any from here.

Comment (dimitri-yatsenko) [reply datajoint/table.py]: Agreed.
```suggestion
                'Auto-populate tables can only be inserted into from their make methods during populate calls.' \
                ' To override, use the the allow_direct_insert argument.')
```

Issue #110, opened 2015-06-25
A method for altering the definitions of tables with data


dimitri-yatsenko:
I propose to have a single `alter` method in `dj.Relation`.  It will take the current definition and update the table definition.  

This will require computing the difference between the current definition and the new definition and generating alter commands to update the definition. 

Two methods:
`preview_alter(self)` -- shows the required changes
`alter(self, data_loss_ok=False, preview_only=False)` -- executes the changes

fabiansinz:
Agreement from my side.

dimitri-yatsenko:
This may be helpful: https://github.com/djrobstep/migra

Issue #17, opened 2014-08-28
Display relation contents


dimitri-yatsenko:
`dj.Relational.__str__`  should display a representation of the contents of the relation

Issue #14, opened 2014-08-28
simple delete


dimitri-yatsenko:
restricted delete without cascading

PR #559, opened 2019-01-22
add alter() & preview_alter()

kimehta:
A method for altering the definitions of tables with data #110 

Only supports table comment, and secondary attribute changes (ADD,DROP,CHANGE).
new attributes are nullable if no default value is specified.

This is the best approach I have aside from maybe using a third party library or doing the same parsing on mysql ddl.

### DIFF
--- a/datajoint/table.py
+++ b/datajoint/table.py
@@ -73,6 +73,182 @@ class Table(QueryExpression):
         else:
             self._log('Declared ' + self.full_table_name)
 
+        def preview_alter(self,new_definition,_preview=True):
+        """
+        Returns changes required to go from current defition to new defition.
+        Does not execute any changes. Use alter() method to finalize any changes.
+        Only supports table comment, and SECONDARY attribute changes(ADD,DROP,CHANGE).
+        Reordering currently not supported.
+        Use # {<old_name>} <comment> to rename given attribute.(insert values between '< >') 
+
+        :param new_definition: string, same format used to create the table.
+        :param _preview: True - Human friendly preview of changes. False - MYSQL ALTER preview.
+        :return: preview changes based on _preview. 
+        """
+
+        output_str = '' #user-friendly preview
+        alter_sql = ''  #alter sql
+
+        #change in table comment
+        new_definition = re.split(r'\s*\n\s*', new_definition.strip())
+        new_table_comment = new_definition.pop(0)[1:].strip() if new_definition[0].startswith('#') else ''
+        
+        if (self.heading.table_info['comment'] != new_table_comment):    
+            output_str += 'change table comment: "'+self._heading.table_info['comment']+'" -> "'+new_table_comment + '".\n'
+            alter_sql += ('COMMENT = "' + new_table_comment + '", ')
+  
+        #similar to heading.attributes.dependentAttributes. Used only to generate output_str
+        newAttributes = [] 
+        
+        in_key = True
+        
+        for line in new_definition:
+            
+            if line.startswith('#'):
+                continue
+            elif line.startswith('---') or line.startswith('___'):
+                in_key = False
+            elif is_foreign_key(line):
+                continue
+            elif re.match(r'^(unique\s+)?index[^:]*$', line, re.I):   # index
+                continue
+            # change in secondary attributes
+            else:
+                if not in_key:
+
+                    try:
+                        match = attribute_parser.parseString(line+'#', parseAll=True)
+                    except pp.ParseException as err:
+                        raise DataJointError('Declaration error in position {pos} in line:\n  {line}\n{msg}'.format(
+                            line=err.args[0], pos=err.args[1], msg=err.args[2]))
+
+                    #external not supported
+                    if match['type'].startswith('external'):
+                        continue
+
+                    #old_name and name are the same if no rename
+                    match['old_name'] = match['name']
+                    match['comment'] = match['comment'].rstrip('#')
+                    
+                    #extract rename, and comment if necessary. brackets after any alphanumerical are ignored.
+                    a = match['comment'].find('{')
+                    b = match['comment'].find('}')
+                    rename = a >= 0 and b >= a+2 and not any(re.match(r"^[a-zA-Z0-9_]*$",c,re.I) for c in match['comment'][0:a])
+
+                    if rename:
+                        match['old_name'] = match['comment'][a+1:b].strip()
+                        match['comment'] = match['comment'][b+1:]
+
+                    if rename and not self.heading.attributes.__contains__(match['old_name']):
+                        raise DataJointError('Incorrect definition line: "' + line + '". ')
+
+                    match['comment'] = match['comment'].replace('"', '\\"')   # escape double quotes in comment
+
+                    #Extract changes in default, and nullable
+                    if 'default' not in match:
+                        match['default'] = ''
+                    match = {k: v.strip() for k, v in match.items()}
+                    match['nullable'] = match['default'].lower() == 'null'
+
+                    literals = ['CURRENT_TIMESTAMP'] 
+                    if match['nullable']:
+                        if in_key:
+                            raise DataJointError('Primary key attributes cannot be nullable in line %s' % line)
+                        match['default'] = 'DEFAULT NULL'
+                    else:
+                        if match['default']:
+                            quote = match['default'].upper() not in literals and match['default'][0] in '"\''
+                            match['default'] = ('"'+match['default'][1:len(match['default'])-1]+'"' if quote else match['default'])
+                        else:
+                            match['default'] = None
+
+                    #add attribute
+                    if match['name'] not in self.heading.attributes and not rename:
+                        output_str += ('Add attribute: \n' + match['name'] +
+                            (' = ' + match['default'] if match['default'] else '') + 
+                            (" : " + match['type']) +
+                            (' # ' + match['comment'] if match['comment'] else '') + '.\n')
+                        alter_sql += ("ADD COLUMN " + match['name'] + ' ' + match['type'] + 
+                            ((' DEFAULT '+ match['default']) if match['default'] else '') + 
+                            ((' COMMENT "' + match['comment'] + '"') if match['comment'] else '') + ', ')
+                        
+                        newAttributes.append(match)
+                        continue
+
+                    #change attribute
+                    old_dtype = self.heading.attributes[match['old_name']].type
+                    old_nullable = self.heading.attributes[match['old_name']].nullable
+                    old_default = self.heading.attributes[match['old_name']].default
+                    old_comment = self.heading.attributes[match['old_name']].comment
+                    new_dtype = match['type']
+                    new_nullable = match['nullable']
+                    new_default = match['default']
+                    new_comment = match['comment']
+
+                    new_definition = new_dtype + (' NOT NULL' if not new_nullable else '') + ((' DEFAULT '+new_default) if new_default else '') + ((' COMMENT "' + new_comment + '"') if new_comment else '')
+                    if rename or old_dtype != new_dtype or old_nullable != new_nullable or old_default != new_default or old_comment != new_comment:
+                        output_str += ('Change attribute: \n' + match['old_name'] +
+                            (' = ' + old_default if old_default else '') + 
+                            (" : " + old_dtype) +
+                            (' # ' + old_comment if old_comment else '') + '.\n' + 
+                            '->\n' + match['name'] +
+                            (' = ' + new_default if new_default else '') + 
+                            (" : " + new_dtype) +
+                            (' # ' + new_comment if new_comment else '') + '.\n')
+                        alter_sql += ('CHANGE COLUMN ' + match['old_name'] + ' ' + match['name'] + ' ' + new_definition + ', ')
+
+                    newAttributes.append(match)
+
+        #Drop attribute
+        for old_attribute in self.heading.dependent_attributes:
+            if (not any(old_attribute == new_attribute['name'] or old_attribute == new_attribute['old_name'] for new_attribute in newAttributes) and 
+                not self.heading.attributes[old_attribute].type.startswith('external') and
+                old_attribute not in np.array(list(self.heading.indexes.keys())).flatten()):
+                    output_str += ('Drop attribute: "' + old_attribute + '".\n')
+                    alter_sql += ('DROP COLUMN ' + old_attribute + ', ')
+
+        #polish
+        if alter_sql != '':
+            alter_sql = alter_sql[0:-2] + ';'
+
+        if _preview:
+            if output_str == '':
+                return None
+            return output_str[0:-1]
+        else:
+            if alter_sql == '':
+                return None
+            return alter_sql        
+
+    def alter(self, new_definition = None, alter_statement = None):
+        """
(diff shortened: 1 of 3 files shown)


kimehta:
indentation error. haven't updated test cases for alter yet.

datajoint/table.py
```diff
@@ -73,6 +73,182 @@ def declare(self, context=None):
...
+        Only supports table comment, and SECONDARY attribute changes(ADD,DROP,CHANGE).
+        Reordering currently not supported.
+        Use # {<old_name>} <comment> to rename given attribute.(insert values between '< >') 
+
+        :param new_definition: string, same format used to create the table.
+        :param _preview: True - Human friendly preview of changes. False - MYSQL ALTER preview.
+        :return: preview changes based on _preview. 
+        """
+
+        output_str = '' #user-friendly preview
```
dimitri-yatsenko: Let's drop the user-friendly preview for now.  The SQL code should be sufficient to present to user for review.

datajoint/table.py
```diff
@@ -73,6 +73,182 @@ def declare(self, context=None):
...
+        #change in table comment
+        new_definition = re.split(r'\s*\n\s*', new_definition.strip())
+        new_table_comment = new_definition.pop(0)[1:].strip() if new_definition[0].startswith('#') else ''
+        
+        if (self.heading.table_info['comment'] != new_table_comment):    
+            output_str += 'change table comment: "'+self._heading.table_info['comment']+'" -> "'+new_table_comment + '".\n'
+            alter_sql += ('COMMENT = "' + new_table_comment + '", ')
+  
+        #similar to heading.attributes.dependentAttributes. Used only to generate output_str
+        newAttributes = []
```
dimitri-yatsenko: Let's use PEP8: `new_attributes`

datajoint/table.py
```diff
@@ -73,6 +73,182 @@ def declare(self, context=None):
...
+                        continue
+
+                    #old_name and name are the same if no rename
+                    match['old_name'] = match['name']
+                    match['comment'] = match['comment'].rstrip('#')
+                    
+                    #extract rename, and comment if necessary. brackets after any alphanumerical are ignored.
+                    a = match['comment'].find('{')
+                    b = match['comment'].find('}')
+                    rename = a >= 0 and b >= a+2 and not any(re.match(r"^[a-zA-Z0-9_]*$",c,re.I) for c in match['comment'][0:a])
```
dimitri-yatsenko:
```suggestion
                    rename = a >= 0 and b >= a+2 and not any(
                            re.match(r"^[a-zA-Z0-9_]+$", c, re.I) for c in match['comment'][:a])
```

datajoint/table.py
```diff
@@ -73,6 +73,182 @@ def declare(self, context=None):
...
+                    #external not supported
+                    if match['type'].startswith('external'):
+                        continue
+
+                    #old_name and name are the same if no rename
+                    match['old_name'] = match['name']
+                    match['comment'] = match['comment'].rstrip('#')
+                    
+                    #extract rename, and comment if necessary. brackets after any alphanumerical are ignored.
+                    a = match['comment'].find('{')
```
dimitri-yatsenko: Use a regular expression or pyparsing to find matching brackets.

datajoint/table.py
```diff
@@ -73,6 +73,182 @@ def declare(self, context=None):
...
+                    b = match['comment'].find('}')
+                    rename = a >= 0 and b >= a+2 and not any(re.match(r"^[a-zA-Z0-9_]*$",c,re.I) for c in match['comment'][0:a])
+
+                    if rename:
+                        match['old_name'] = match['comment'][a+1:b].strip()
+                        match['comment'] = match['comment'][b+1:]
+
+                    if rename and not self.heading.attributes.__contains__(match['old_name']):
```
dimitri-yatsenko:
```suggestion
                    if rename and match['old_name'] not in self.heading.attributes:
```

datajoint/table.py
```diff
@@ -73,6 +73,182 @@ def declare(self, context=None):
...
+                            (' # ' + match['comment'] if match['comment'] else '') + '.\n')
+                        alter_sql += ("ADD COLUMN " + match['name'] + ' ' + match['type'] + 
+                            ((' DEFAULT '+ match['default']) if match['default'] else '') + 
+                            ((' COMMENT "' + match['comment'] + '"') if match['comment'] else '') + ', ')
+                        
+                        newAttributes.append(match)
+                        continue
+
+                    #change attribute
+                    old_dtype = self.heading.attributes[match['old_name']].type
```
dimitri-yatsenko: It's inefficient to each attribute into its own variable.

datajoint/table.py
```diff
@@ -73,6 +73,182 @@ def declare(self, context=None):
...
+                            (' # ' + new_comment if new_comment else '') + '.\n')
+                        alter_sql += ('CHANGE COLUMN ' + match['old_name'] + ' ' + match['name'] + ' ' + new_definition + ', ')
+
+                    newAttributes.append(match)
+
+        #Drop attribute
+        for old_attribute in self.heading.dependent_attributes:
+            if (not any(old_attribute == new_attribute['name'] or old_attribute == new_attribute['old_name'] for new_attribute in newAttributes) and 
+                not self.heading.attributes[old_attribute].type.startswith('external') and
+                old_attribute not in np.array(list(self.heading.indexes.keys())).flatten()):
```
dimitri-yatsenko:
```suggestion
                old_attribute not in self.heading.indexes):
```

datajoint/table.py
```diff
@@ -73,6 +73,182 @@ def declare(self, context=None):
...
+
+    def alter(self, new_definition = None, alter_statement = None):
+        """
+        Execute an ALTER TABLE statement.
+        :param new_definition: new definition.
+        :param alter_statement: optional alter statements. Do not include ALTER TABLE <table_name>.
+        if provided, alter_statement overrides new_definition.
+        """
+
+        if new_definition == None and alter_statement == None:
```
dimitri-yatsenko:
```suggestion
        if new_definition is None and alter_statement is None:
```

datajoint/table.py
```diff
@@ -73,6 +73,182 @@ def declare(self, context=None):
...
+        #polish
+        if alter_sql != '':
+            alter_sql = alter_sql[0:-2] + ';'
+
+        if _preview:
+            if output_str == '':
+                return None
+            return output_str[0:-1]
+        else:
+            if alter_sql == '':
```
dimitri-yatsenko:
```suggestion
            return alter_sql or None
```

datajoint/table.py
```diff
@@ -73,6 +73,182 @@ def declare(self, context=None):
...
+        if self.connection.in_transaction:
+            raise DataJointError("Table definition cannot be altered during a transaction.")
+        
+        if alter_statement == None:
+            alter_statement = self.preview_alter(new_definition,False)
+
+        if alter_statement == None:
+            return
+        sql = 'ALTER TABLE %s %s' % (self.full_table_name, alter_statement)
+        try:
```
dimitri-yatsenko: Is this `try ... except ...` helpful?

datajoint/table.py
```diff
@@ -73,6 +73,182 @@ def declare(self, context=None):
...
+                            (' = ' + new_default if new_default else '') + 
+                            (" : " + new_dtype) +
+                            (' # ' + new_comment if new_comment else '') + '.\n')
+                        alter_sql += ('CHANGE COLUMN ' + match['old_name'] + ' ' + match['name'] + ' ' + new_definition + ', ')
+
+                    newAttributes.append(match)
+
+        #Drop attribute
+        for old_attribute in self.heading.dependent_attributes:
+            if (not any(old_attribute == new_attribute['name'] or old_attribute == new_attribute['old_name'] for new_attribute in newAttributes) and
```
dimitri-yatsenko: apply De Morgan's rule to simplify

datajoint/table.py
kimehta: removed

kimehta:
still need to do something about make methods,

datajoint/table.py
```diff
@@ -72,7 +72,145 @@ def declare(self, context=None):
...
+                        continue
+
+                    #old_name and name are the same if no rename
+                    attr['old_name'] = attr['name']
+                    attr['comment'] = attr['comment'].rstrip('#')
+                    
+                    #extract rename, and new comment if necessary. brackets after any alphanumerical are ignored.
+                    rename = re.match(r"[^\w]*{[a-z][a-z0-9_]*}", attr['comment'])
+                    if rename:
+                        rename = True
```
dimitri-yatsenko: unnecessary.  `bool(rename)` is already `True`.

datajoint/table.py
```diff
@@ -72,7 +72,145 @@ def declare(self, context=None):
...
+                    #external not supported
+                    if attr['type'].startswith('external'):
+                        continue
+
+                    #old_name and name are the same if no rename
+                    attr['old_name'] = attr['name']
+                    attr['comment'] = attr['comment'].rstrip('#')
+                    
+                    #extract rename, and new comment if necessary. brackets after any alphanumerical are ignored.
+                    rename = re.match(r"[^\w]*{[a-z][a-z0-9_]*}", attr['comment'])
```
dimitri-yatsenko: Try
```suggestion
                    rename = re.match(r'({(?P<name>\w+)\s*})?\s*(?P<comment>.*)', attr['comment'])
                    if rename.group('name'):
                        ...
                    
```

datajoint/table.py
```diff
@@ -72,7 +72,145 @@ def declare(self, context=None):
...
+                        new_attributes.append(attr)
+                        continue
+
+                    #change attribute
+                    column_definition = (attr['type']
+                                        + (' NOT NULL' if not attr['nullable'] else '')
+                                        + (' DEFAULT ' + attr['default'] if attr['default'] else '') 
+                                        + (' COMMENT "' + attr['comment'] + '"' if attr['comment'] else ''))
+                    if (rename
+                        or self.heading.attributes[attr['old_name']].type != attr['type']
```
dimitri-yatsenko: these comparisons want to be a loop

kimehta: why would it be a loop? it is inside 
https://github.com/datajoint/datajoint-python/blob/eb6410770f9ebd154c44c4252938462b78a7c2c8/datajoint/table.py#L100
.
.
.
https://github.com/datajoint/datajoint-python/blob/eb6410770f9ebd154c44c4252938462b78a7c2c8/datajoint/table.py#L167-L168

datajoint/table.py
kimehta: `r'{\s*(?P<name>\w+)\s*}\s*(?P<comment>.*)'`
legal: 234_f. allows for illegal attribute names like digit only

`r'{\s*(?P<name>[a-z][a-z0-9_]*)\s*}\s*(?P<comment>.*)'`
How about this: alphanumerical lowercase must start with a letter
fails: # {er} sf. space after # in attribute comment

this?: added \s* in the front to deal with blank space before '{'
`r'\s*{\s*(?P<name>[a-z][a-z0-9_]*)\s*}\s*(?P<comment>.*)'`

dimitri-yatsenko: Oh, the main thing I was pointing out is that you can use `rename.group('name')` inside the condition rather than parsing again.   So yes, please use the more precise regular expression.

kimehta: https://github.com/datajoint/datajoint-python/blob/c12dd9a936553bc52610b7193ab26208778f1bbf/datajoint/table.py#L129-L132

datajoint/table.py
kimehta: https://github.com/datajoint/datajoint-python/blob/c12dd9a936553bc52610b7193ab26208778f1bbf/datajoint/table.py#L168-L171

datajoint/table.py
kimehta: https://github.com/datajoint/datajoint-python/blob/c12dd9a936553bc52610b7193ab26208778f1bbf/datajoint/table.py#L175-L180

datajoint/table.py
```diff
@@ -72,7 +74,142 @@ def declare(self, context=None):
...
+                    else:
+                        if attr['default']:
+                            quote = attr['default'].upper() not in literals and attr['default'][0] in '"\''
+                            attr['default'] = ('"'+attr['default'][1:len(attr['default'])-1]+'"' if quote else attr['default'])
+                        else:
+                            attr['default'] = None
+
+                    #add attribute
+                    if attr['name'] not in self.heading.attributes and not rename:
+                        alter_sql += ("ADD COLUMN"
```
dimitri-yatsenko: Using `str.format` would be look clearer, e.g. something like
```python
"ALTER COLUMN {name} {type}...".format(**attr)
```

datajoint/table.py
```diff
@@ -72,7 +74,142 @@ def declare(self, context=None):
...
+                        alter_sql += ("ADD COLUMN" 
+                                    + (' ' + attr['name'])
+                                    + (' ' + attr['type']) 
+                                    + (' ' + 'DEFAULT ' + attr['default'] if attr['default'] else '') 
+                                    + (' ' + 'COMMENT "' + attr['comment'] + '"' if attr['comment'] else '') + ', ')
+                        new_attributes.append(attr)
+                        continue
+
+                    #change attribute
+                    column_definition = (attr['type']
```
dimitri-yatsenko: use `str.format` for better clarity

datajoint/table.py
```diff
@@ -72,7 +74,142 @@ def declare(self, context=None):
...
+                                        + (' NOT NULL' if not attr['nullable'] else '')
+                                        + (' DEFAULT ' + attr['default'] if attr['default'] else '') 
+                                        + (' COMMENT "' + attr['comment'] + '"' if attr['comment'] else ''))
+                    if (rename
+                        or any(getattr(self.heading.attributes[attr['old_name']],attr_def) != attr[attr_def] 
+                            for attr_def in ('type','nullalble','default','comment'))):
+                        alter_sql += ('CHANGE COLUMN ' + attr['old_name'] + ' ' + attr['name'] + ' ' + column_definition + ', ')
```
dimitri-yatsenko: use `str.format`

kimehta:
@dimitri-yatsenko 

Like this? Should i include tests for all(dj.imported,dj.computed,dj.manual) as well?

https://github.com/datajoint/datajoint-python/blob/d807be0bf9818c0ed3728f833f759b6737a18504/tests/test_relation.py#L246-L299

kimehta:
the alter commands generated by my code are the same for each version of python.
Travis ci passing python 3.7, failing 3.6. the only difference(i think) is mysql version(5.7 & 5.6 respectively). 
I tried running generated alters on db-fiddle and it works fine on both mysql versions.
pymysql version are the same for each python.
the code works fine on my machine for different version of python with datajoint mysql docker.

I am trying to find the cause but if anyone would like to take a look.

#17 is passing without modifying enum values, but with reorder support.(python 3.6)
#14 is passing with modified enum, but without reorder(after colum) support.(python 3.6)
python 3.7 is passing everything.
https://travis-ci.com/kimehta/datajoint-python/builds
here "passing" is not throwing mysql error#1292 incorrect date value. I am failing everything to collect debug info.

dimitri-yatsenko:
closing for a pending resubmission

PR #569, opened 2019-02-28
Revised implementation of configurable external storage

dimitri-yatsenko:
* External storage uses UUIDs
* A separate external table is created in each schema for each store
* External table cleanup is done with `.delete()` and `.clean()`
* `schema.external` is now a container that manages external tables


datajoint/external.py
```diff
@@ -192,8 +178,32 @@ def clean_store(self, store, verbose=True):
...
+        self.schema = schema
+        self.external_tables = {}
+
+    def __getitem__(self, store):
+        if store not in self.external_tables:
+            self.external_tables[store] = ExternalTable(
+                connection=self.schema.connection, store=store, database=self.schema.database)
+        return self.external_tables[store]
+
+    def __iter__(self):
```
ixcat: can we get a `keys()` too?
andor alternately could external mapping generally not be done via some more complete and well thought out notion of `DefaultDict(lambda k: ExternalTable(config[store]))` ?

datajoint/external.py
```diff
@@ -21,63 +21,65 @@ class ExternalTable(Table):
...
-        store = ''.join(store.split('-')[1:])
-        spec = config.get_store_spec(store)
-        blob_hash = long_hash(blob) + store
-        if spec['protocol'] == 'file':
-            folder = os.path.join(spec['location'], self.database, *subfold(blob_hash, spec['subfolding']))
-            full_path = os.path.join(folder, blob_hash)
+        blob_hash = uuid_from_buffer(blob)
+        if self.spec['protocol'] == 'file':
+            folder = os.path.join(self.spec['location'], self.database, *subfold(blob_hash.hex, self.spec['subfolding']))
+            full_path = os.path.join(folder, blob_hash.hex)
```
ixcat: clearly, as this line illustrates, not a complicated conversion, however, would it be preferable to save directly to the uuid string so hash:filename is 1:1? (but still with subpaths path splitting?)

datajoint/external.py
ixcat: actually, thinking more not sure if i like the 'simple' dict approach since custom class could allow things like maintenance methods, like 'consistency check', 'bulk cleanup' etc to happen at 'schema' level.

ixcat: getting keys still useful - unless i'm mistaken there is not a clear/simple way to retrieve 'names' of stores w/o knowing internals of the mapping since `list(schema.external)[0].spec` has config 'values' but not 'names'.. but could be overlooking something.

dimitri-yatsenko: Yes, I am thinking that `ExternalMapping` should derive from `collections.Mapping` and act like a dict with additional methods.

datajoint/external.py
dimitri-yatsenko: It uses `binary(16)` to store UUIDs.  When you view the external table in DataJoint, the UUIDs show as UUID strings. If you need to view them in SQL, you can do something like `SELECT hex(hash) FROM external_local`

datajoint/external.py
dimitri-yatsenko: This has been fixed now. `schema.external`is a Mapping.

tests/schema_external.py
```diff
@@ -74,8 +63,8 @@ class Image(dj.Computed):
     -> Seed
     -> Dimension
     ----
-    img  : blob-raw    #  objects are stored as specified by dj.config['stores'][-raw']
-    neg : blob-    # objects are stored as specified by dj.config['stores']['-']
+    img : blob@raw     #  objects are stored as specified by dj.config['stores'][-raw']
```
eywalker: Please update this comment in the future to reflect the correct config.

datajoint/declare.py
```diff
@@ -6,14 +6,41 @@
...
+    DECIMAL=r'(decimal|numeric)(\s*\(.+\))?(\s+unsigned)?$',
+    FLOAT=r'(double|float|real)(\s*\(.+\))?(\s+unsigned)?$',
+    STRING=r'(var)?char\s*\(.+\)$',
+    ENUM=r'enum\s*\(.+\)$',
+    BOOL=r'bool(ean)?$',   # aliased to tinyint(1)
+    TEMPORAL=r'(date|datetime|time|timestamp|year)(\s*\(.+\))?$',
+    INTERNAL_BLOB=r'(tiny|small|medium|long|)blob$',
+    INTERNAL_ATTACH=r'attach$',
+    EXTERNAL_ATTACH=r'attach@(?P<store>[a-z]\w*)$',
+    EXTERNAL_BLOB=r'blob@(?P<store>[a-z]\w*)$',
```
eywalker: I'm totally going to reorder this in the future so that `EXTERNAL_BLOB` comes right after `INTERNAL_BLOB`

PR #573, opened 2019-03-17
Altering table definition (Fix #110)

### LINKED ISSUE #110: A method for altering the definitions of tables with data

dimitri-yatsenko:
* Fix #110


datajoint/declare.py
```diff
@@ -5,9 +5,16 @@
...
-
+from collections import defaultdict
 from .errors import DataJointError
 
+import sys
+if sys.version_info[1] < 6:
+    from collections import OrderedDict
+else:
+    # use dict in Python 3.6+ -- They are already ordered and look nicer
+    OrderedDict = dict
```
guzman-raphael: Replace the dict import here with:

`from .utils import OrderedDict`

PR #576, opened 2019-03-20
fix issue #572 -- support for scalars in blobs

### LINKED ISSUE #572: np.float32 not supported for longblobs yet

Error message:

dimitri-yatsenko:
Fix #572


dimitri-yatsenko:
This will be released along with other blobs fixes in #577

PR #577, opened 2019-03-20
dj0: expanded blob functionality

### LINKED ISSUE #244: Serializing non `ndarray` objects

### LINKED ISSUE #392: Serializing dict as blob assumes string keys
It seems dict serialization was added last year in #248 by @eywalker . I'm trying to add a dict as a mediumblob to our database. The dict has integer keys, not string keys, and they map to tuples. It looks something like this:
```
{1:(0, 100), 2:(0, 200)}
```
Currently, datajoint assumes that keys are always strings:
https://github.com/datajoint/datajoint-python/blob/master/datajoint/blob.py#L286

I tried changing `pack_string(k)` to `pack_obj(v)`, which works fine while pack, but later I get a byte count assertion error when unpacking the blob here:
https://github.com/datajoint/datajoint-python/blob/master/datajoint/blob.py#L111

Any ideas?

### LINKED ISSUE #427: Expand blob serialization functionality
This issue pulls together issues related to blob serialization to propose a single solution to address them.

This includes issues #392 and #244. 

The goal is to allow storing arbitrary Python objects if MATLAB compatibility is not important.  Clarify what objects are portable between MATLAB and Python and allow configuring alternative serialization and compression protocols.

### LINKED ISSUE #475: subfolders for external storage

### LINKED ISSUE #520: Python Blob cannot store list of Strings
When storing a list of data inside a blob, string datatypes are not functional in Python. 

When attempting to store a string list into a blob an error like the following will be thrown:
```
~/code/Python_3/lib/python3.7/site-packages/datajoint/blob.py in pack_array(array)
    252         array, imaginary = np.real(array), np.imag(array)
    253 
--> 254     type_number = rev_class_id[array.dtype]
    255 
    256     if dtype_list[type_number] is None:

KeyError: dtype('<U1')
```

### LINKED ISSUE #572: np.float32 not supported for longblobs yet

### LINKED ISSUE #594: uuid string insert/restrict support
Since uuid's have a clear string to value mapping, using string versions may be more user friendly in some cases where it would be possible to support them.

For the moment, this is:

- record insert
- restrict via dictionary

dimitri-yatsenko:
Blob now supports serialization of simple Python data types: `dict`, `tuple`, `list`, `set`, `string`, and `bytes`, `np.array` of `dtype('O')`, numeric scalars, `bool`, matlab cell arrays, and matlab struct arrays, `datetime.datetime`, `datetime.date`, `datetime.time`.

* Fix #572
* Fix #520 
* Fix #427
* Fix #392 
* Fix #244

* Fix #594 -- support string values for UUID attributes.


guzman-raphael:
Following my review with @eywalker, the following questions came up:

1. Why is the record array representation denoted by `F` as opposed to `f`? Thought this was agreed to be lower case.

https://github.com/dimitri-yatsenko/datajoint-python/blob/25965894484a82de68a6d4dde46e64291214e997/datajoint/blob.py#L121

2. Was there not a lighter way to pack a decimal e.g. 2 integers? Curious why string was utilized.

https://github.com/dimitri-yatsenko/datajoint-python/blob/25965894484a82de68a6d4dde46e64291214e997/datajoint/blob.py#L253-L256

3. Using the default `encode` method utilizes UTF8. Are we confident that all python string values can be represented in UTF8?

https://github.com/dimitri-yatsenko/datajoint-python/blob/25965894484a82de68a6d4dde46e64291214e997/datajoint/blob.py#L261-L264

datajoint/blob.py
```diff
@@ -34,286 +37,384 @@
...
+                b"".join(self.pack_recarray(array[f]) if array[f].dtype.fields else self.pack_array(array[f])
+                         for f in array.dtype.names))
+
+    def read_sparse_array(self):
+        raise DataJointError('datajoint-python does not yet support sparse arrays. Issue (#590)')
+
+    def read_decimal(self):
+        return Decimal(self.read_string())
+
+    @staticmethod
```
guzman-raphael: Was there not a lighter way to pack a decimal e.g. 2 integers? Curious why string was utilized.

datajoint/blob.py
```diff
@@ -34,286 +37,384 @@
...
+    def pack_decimal(d):
+        s = str(d)
+        return b"d" + len_u64(s) + s.encode()
+
+    def read_string(self):
+        return self.read_binary(self.read_value()).decode()
+
+    @staticmethod
```
guzman-raphael: Using the default encode method utilizes UTF8. Are we confident that all python string values can be represented in UTF8?

datajoint/blob.py
```diff
@@ -34,286 +37,384 @@
...
+    def read_sparse_array(self):
+        raise DataJointError('datajoint-python does not yet support sparse arrays. Issue (#590)')
+
+    def read_decimal(self):
+        return Decimal(self.read_string())
+
+    @staticmethod
+    def pack_decimal(d):
+        s = str(d)
+        return b"d" + len_u64(s) + s.encode()
```
guzman-raphael: Was there not a lighter way to pack a decimal e.g. 2 integers? Curious why string was utilized.

datajoint/blob.py
```diff
@@ -34,286 +37,384 @@
...
+
+    def read_string(self):
+        return self.read_binary(self.read_value()).decode()
+
+    @staticmethod
+    def pack_string(s):
+        blob = s.encode()
+        return b"\5" + len_u64(blob) + blob
```
guzman-raphael: Using the default encode method utilizes UTF8. Are we confident that all python string values can be represented in UTF8?

datajoint/blob.py
dimitri-yatsenko: I do not know if two integers are enough. Could not find documentation.

datajoint/blob.py
```diff
@@ -34,286 +37,384 @@
...
+                "C": self.read_cell_array,    # matlab cell array
+                # Python-native
+                "\xFF": self.read_none,      # None
+                "\1": self.read_tuple,     # a Sequence
+                "\2": self.read_list,      # a MutableSequence
+                "\3": self.read_set,       # a Set
+                "\4": self.read_dict,      # a Mapping
+                "\5": self.read_string,    # a UTF8-encoded string
+                "\6": self.read_bytes,     # a ByteString
+                "F": self.read_recarray,   # numpy array with fields, including recarrays
```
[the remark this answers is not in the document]
dimitri-yatsenko: Just following the pattern that all arrays are uppercase.

datajoint/blob.py
dimitri-yatsenko: What is the alternative. UTF8 is dominant and continues to grow.

datajoint/blob.py
```diff
@@ -34,286 +37,384 @@
...
+                         for f in array.dtype.names))
+
+    def read_sparse_array(self):
+        raise DataJointError('datajoint-python does not yet support sparse arrays. Issue (#590)')
+
+    def read_decimal(self):
+        return Decimal(self.read_string())
+
+    @staticmethod
+    def pack_decimal(d):
```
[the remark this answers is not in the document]
dimitri-yatsenko: I have looked around and could not find the optimal robust representation. string seems common.

datajoint/blob.py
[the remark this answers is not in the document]
dimitri-yatsenko: I followed the pattern that arrays are capitalized.

chrisroat:
Do you know if this would allow storage of general pandas DataFrames as blobs?

ixcat:
hi @chrisroat , just happened to be reviewing this issue while working elsewhere - pandas inserts should be functional (as well as fetch(... format='frame') - see also: `https://github.com/datajoint/datajoint-python/pull/776`

PR #578, opened 2019-03-26
Fix #499: Proj accepts Ellipsis ... and '-attr' to exclude attributes

### LINKED ISSUE #499: support ellipsis in proj to signify "all remaining attributes"
At the moment, there isn't a convenient mechanism to rename an attribute or calculate a new attribute while keep all the other secondary attribute.  In matlab, we use '*' for that purpose.  I propose to use ellipsis. 

For example, 
```python
s.proj(a='b', ...)    # rename b to a, keep everything else.
```

dimitri-yatsenko:
Fix #499  to allow Ellipsis in `proj` and `aggr`

You may now do something like:
```python
q = Person.proj(..., '-date_of_birth')  # all attributes except the date of birth
```

Review after merging #569

### DIFF
--- a/datajoint/expression.py
+++ b/datajoint/expression.py
@@ -688,6 +688,27 @@ class Projection(QueryExpression):
             self._heading = arg.heading
             self._arg = arg._arg
 
+    @staticmethod
+    def prepare_attribute_lists(arg, attributes, named_attributes):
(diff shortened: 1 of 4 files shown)


eywalker:
Accepting for dev release.

PR #586, opened 2019-04-03
Cleanup and simplification, Fix #572

### LINKED ISSUE #572: np.float32 not supported for longblobs yet

dimitri-yatsenko:
This contains only cleanup and simplifications of the `blob` module without functional change.

This PR is intended to break up PR #577 into smaller chunks. 

The only functional change is that scalar values are now accepted in blobs, fixing #572. MATLAB will not be able to see these blobs and will read them as empty arrays.


dimitri-yatsenko:
closing due to nearing completion of #577

PR #589, opened 2019-04-03
updated blob.bypass_serialization patch

ixcat:
per feeedback in #578

### DIFF
--- a/datajoint/blob.py
+++ b/datajoint/blob.py
@@ -213,7 +213,7 @@ class BlobReader:
 
 
 def pack(obj, compress=True):
-    if bypass_serialization is True:
+    if bypass_serialization:
         assert obj.startswith(tuple(decode_lookup))
         return obj
 
@@ -311,7 +311,7 @@ def unpack(blob, **kwargs):
     if blob is None:
         return None
 
-    if bypass_serialization is True:
+    if bypass_serialization:
         assert blob.startswith(tuple(decode_lookup))
         return blob
 
--- a/datajoint/blob.py
+++ b/datajoint/blob.py
@@ -8,7 +8,6 @@ from decimal import Decimal
 from datetime import datetime
 import numpy as np
 from .errors import DataJointError
-from .settings import config
 
 
 mxClassID = OrderedDict((
--- a/datajoint/blob.py
+++ b/datajoint/blob.py
@@ -8,6 +8,8 @@ from decimal import Decimal
 from datetime import datetime
 import numpy as np
 from .errors import DataJointError
+from .settings import config
+
 
 mxClassID = OrderedDict((
     # see http://www.mathworks.com/help/techdoc/apiref/mxclassid.html
@@ -210,6 +212,9 @@ class BlobReader:
 
 
 def pack(obj, compress=True):
+    if config['blob.encode_bypass'] is True:
+        return obj
+
     blob = b"mYm\0"
     blob += pack_obj(obj)
 
@@ -300,8 +305,11 @@ def pack_dict(obj):
 
 
 def unpack(blob, **kwargs):
+
+    if config['blob.encode_bypass'] is True:
+        return blob
+
     if blob is None:
         return None
 
     return BlobReader(blob, **kwargs).unpack()
-
--- a/datajoint/blob.py
+++ b/datajoint/blob.py
@@ -38,6 +38,8 @@ decode_lookup = {
     b'ZL123\0': zlib.decompress
 }
 
+bypass_serialization = False  # runtime setting to bypass blob (en|de)code
+
 
 class BlobReader:
     def __init__(self, blob, squeeze=False, as_dict=False):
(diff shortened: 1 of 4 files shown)


ixcat:
one remaining task is failure mode - not sure best way / level to do this -

Could just sanity check the pack/unpack data checks..

Manually injecting bad data into a table seems like alot of work, but could be done if deemed necessary.

datajoint/blob.py
```diff
@@ -210,6 +213,11 @@ def __str__(self):
 
 
 def pack(obj, compress=True):
+    if bypass_serialization is True:
```
dimitri-yatsenko:
```suggestion
    if bypass_serialization:
```

datajoint/blob.py
```diff
@@ -210,6 +213,11 @@ def __str__(self):
...
+        assert(any([obj[:len(pattern)] == pattern
```
dimitri-yatsenko:
```suggestion
        assert obj.startswith(tuple(decode_loookup))
```
`assert` is not a function. Any can take an iterator and quit as soon as condition is met without forming a list. `str.startswith` can take a tuple of strings.

datajoint/blob.py
```diff
@@ -210,6 +213,11 @@ def __str__(self):
...
+               for pattern in decode_lookup]))
```
dimitri-yatsenko:
```suggestion
```

datajoint/blob.py
```diff
@@ -300,8 +308,13 @@ def pack_dict(obj):
 
 
 def unpack(blob, **kwargs):
+
     if blob is None:
         return None
 
-    return BlobReader(blob, **kwargs).unpack()
+    if bypass_serialization is True:
+        assert(any([blob[:len(pattern)] == pattern
```
dimitri-yatsenko:
```suggestion
        assert blob.startswith(tuple(decode_lookup))
```

datajoint/blob.py
```diff
@@ -300,8 +308,13 @@ def pack_dict(obj):
...
+               for pattern in decode_lookup]))
```
dimitri-yatsenko:
```suggestion
```

tests/test_bypass_serialization.py
```diff
@@ -0,0 +1,47 @@
...
+import numpy as np
+
+schema_in = dj.schema(PREFIX + '_test_bypass_serialization_in',
+                      connection=dj.conn(**CONN_INFO))
+
+schema_out = dj.schema(PREFIX + '_test_blob_bypass_serialization_out',
+                       connection=dj.conn(**CONN_INFO))
+
+
+tst_blob = np.array([1, 2, 3])  # test blob;
```
dimitri-yatsenko:
```suggestion
test_blob = np.array([1, 2, 3]) 
```

tests/test_bypass_serialization.py
```diff
@@ -0,0 +1,47 @@
...
+
+def test_bypass_serialization():
+    dj.blob.bypass_serialization = True
+    OutputTable.insert(InputTable.fetch(as_dict=True))
+    dj.blob.bypass_serialization = True
+
+    ins = InputTable.fetch(as_dict=True)
+    outs = OutputTable.fetch(as_dict=True)
+
+    assert_true(all(np.array_equals(i[0]['data'], tst_blob),
```
dimitri-yatsenko: This `assert_true` will never fail. It is not testing what you like it to test.

dimitri-yatsenko: This assumes that `ins` and `outs` are sorted the same way.  What guarantees that?

ixcat: there's only one record :) but, good point. will switch to fetch1.

dimitri-yatsenko:
progress?

ixcat:
updated/incorporated suggestions

tests/test_bypass_serialization.py
```diff
@@ -36,12 +36,9 @@ class OutputTable(dj.Manual):
...
-    ins = InputTable.fetch(as_dict=True)
-    outs = OutputTable.fetch(as_dict=True)
+    dj.blob.bypass_serialization = False
 
-    assert_true(all(np.array_equals(i[0]['data'], tst_blob),
-                np.array_equals(i[0]['data'], i[1]['data']))
-                for i in zip(ins, outs))
+    i = InputTable.fetch1()
+    o = OutputTable.fetch1()
+    assert_true(np.array_equal(i['data'], o['data']))
```
dimitri-yatsenko: you can do assert_array_equal

tests/test_bypass_serialization.py
```diff
@@ -0,0 +1,44 @@
...
+    definition = """
+    id:                 int
+    ---
+    data:               blob
+    """
+
+
+def test_bypass_serialization():
+    dj.blob.bypass_serialization = True
+    OutputTable.insert(InputTable.fetch(as_dict=True))
```
dimitri-yatsenko: hm.. this didn't really test the bypass

dimitri-yatsenko: also, this would pass if `InputTable` were empty, no matter what.

dimitri-yatsenko: ok, i will merge and add one more assert to fix this.

PR #593, opened 2019-04-12
Fix #592 and #595

### LINKED ISSUE #592: Avoid re-downloading file if it already exists locally (attach method)
Add the option to check if the existing local file is the same and not create new files with the same contents.
Check the file size first, which is cheap. In most cases, different files will differ in size.  Run the checksum only if the files are the same size

### LINKED ISSUE #595: Allow fetching a subset of attributes as a dictionary
When fetching a subset of attributes, datajoint should be able to return them as a dictionary array when specifying `as_dict=True`. With `as_dict=False`, the returning results are list of values, while with `as_dict=True`, all fields are packed as a list of dictionary.

dimitri-yatsenko:
* Fix #592  Avoid creating multiple copies of attachments.
* Fix #595 
* Implement `schema.code` and `schema.save()`

**Merge after merging #577 and #601**


guzman-raphael:
Following my review with @eywalker, the following questions came up...

datajoint/fetch.py
```diff
@@ -35,18 +37,46 @@ def to_dicts(recarray):
...
+        peek, size = extern.peek(uuid.UUID(bytes=data)) if attr.is_external else (data, len(data))
+        assert size is not None 
+        filename = peek.split(b"\0", 1)[0].decode()
+        size -= len(filename) + 1
+        filepath = os.path.join(download_path, filename)
+        if os.path.isfile(filepath) and size == os.path.getsize(filepath):
+            local_checksum = hash.uuid_from_file(download_path, filename)
+            remote_checksum = (uuid.UUID(bytes=data)
+                    if attr.is_external else hash.uuid_from_buffer(data))
+            if local_checksum == remote_checksum:
```
guzman-raphael: Is it necessary to do a checksum each time? Could we perhaps create a hidden folder and store in a file the logged sum and compare against it rather than compute each time?

dimitri-yatsenko: but how do you ensure that the file hasn't changed?

PR #601, opened 2019-04-25
Add testing coverage for S3 utilizing Minio.

guzman-raphael:


dimitri-yatsenko:
Having a minio server for tests is great. Thank you, @guzman-raphael !

PR #603, opened 2019-05-03
Filepath datatype (Fix #481)

### LINKED ISSUE #481: File management
# Purpose
Enable the management of data as collections of files browsable and accessible bypassing DataJoint and organized in a way that makes sense to the user.  DataJoint must maintain data integrity.

# Operation
File tracking keeps the filenames and organization visible to and controllable by external users.  The files are identified by their paths relative to the repository path.  The datatype should be either file or file-suffix where the suffix can be up to eight character long.  

Unlike external storage and attachments, file tracking does not rely on separate subfolders for each schema.  Files must be accessible under the same locations by multiple schemas.  Therefore, for tracked files, datajoint will create a local subfolder .datajoint where it will contain log files describing the tracking info.  For example, for the file tracked at `ephys/day1/file001.dat`, datajoint will create the log file `ephys/day1/.datajoint/file001.dat.log`

The log file contains the information about the storing schema and the storing configuration. 

```python
@schema
class Ephys:
    definition = """
    -> Session
    ---
    ephys_file  :  filepath@ephys  # in-place path
    """
```

## Insert 
Inserting performs these tasks:
1. If the source file is not in the target repository, copy it there
2. Once the file is in the repository, update the log file to indicate how it is tracked: server, schema, and checksum (when reasonable). 
3. Insert the relative path into the target table

```python
Ephys.insert1(1, '/sessions/2018-08-06/rec001.dat')
```

The file can be specified as the 2-tuple with the source and destination locations.

```python
Ephys.insert1(1, ['c:/tmp/rec001.dat', '/sessions/2018-08-06/rec001.dat'])
```

## Fetch
Fetch performs the following operations:
1. Copy the file from repository to the download folder, if necessary.
2. Return the full file path as the regular attribute value.

dimitri-yatsenko:
* Fix #481


guzman-raphael:
Per the review Edgar and I had on this on Friday, the following were the questions that came up.

datajoint/utils.py
```diff
@@ -2,6 +2,7 @@
 
 import re
 import os
+import shutil
 from .errors import DataJointError
```
guzman-raphael: Need to add OrderedDict definition here.

datajoint/external.py
```diff
@@ -45,44 +45,82 @@ def __init__(self, connection, store=None, database=None):
...
+        Raise exception if an external entry already exists with a different contents checksum.
+        Otherwise, copy (with overwrite) file to remote and
+        If an external entry exists with the same checksum, then no copying should occur
+        """
+        local_folder = os.path.dirname(local_filepath)
+        stage_folder = os.path.join(os.path.abspath(self.spec['stage']), '')
+        if not local_folder.startswith(stage_folder):
+            raise DataJointError('The path {path} is not in stage {stage}'.format(
+                path=local_folder, stage=stage_folder))
+        relative_filepath = local_filepath[len(stage_folder):]
```
guzman-raphael: Can we avoid string operations on `local_filepath`? `fput` assumes this is an absolute path. What about user-specified relative paths, symbolic links? We can add tests for those cases.

datajoint/external.py
```diff
@@ -45,44 +45,82 @@ def __init__(self, connection, store=None, database=None):
...
+            if self.spec['protocol'] == 's3':
+                self.s3.fput(relative_filepath, local_filepath, contents_hash=str(contents_hash))
+            else:
+                remote_file = os.path.join(self.spec['location'], relative_filepath)
+                safe_copy(local_filepath, remote_file, overwrite=True)
+            self.connection.query(
+                "INSERT INTO {tab} (hash, size, filepath, contents_hash) VALUES (%s, {size}, '{filepath}', %s)".format(
+                    tab=self.full_table_name, size=os.path.getsize(local_filepath),
+                    filepath=relative_filepath), args=(uuid.bytes, contents_hash.bytes))
+        return uuid
```
guzman-raphael: Don't see an implementation that signals on-going changes to a file (similar to vim swap file). This would prevent collisions if multiple users are updating the same entry.

datajoint/s3.py
```diff
@@ -4,72 +4,62 @@
...
+        return self.client.list_objects(self.bucket, '/'.join((self.remote_path, folder, '')), recursive=True)
 
-        failed_deletes = list(
-            self.client.remove_objects(self.bucket, itertools.islice(names, max_count)))
+    def remove_objects(self, objects_iter):
 
-        print('Deleted: %i S3 objects' % next(count))
-        return failed_deletes
+        failed_deletes = self.client.remove_objects(self.bucket, objects_iter=objects_iter)
+        return list(failed_deletes)
```
guzman-raphael: Seems that there is a possibility here that a delete will trigger a delete of other existing external blobs that do not have `filepath` defined.

datajoint/external.py
```diff
@@ -45,44 +45,82 @@ def __init__(self, connection, store=None, database=None):
...
+                self.spec['location'], self.database, *subfold(uuid.hex, self.spec['subfolding'])), uuid.hex)
+            safe_write(remote_file, blob)
         # insert tracking info
         self.connection.query(
-            "INSERT INTO {tab} (hash, size) VALUES (%s, {size}) "
-            "ON DUPLICATE KEY UPDATE timestamp=CURRENT_TIMESTAMP".format(
-                tab=self.full_table_name, size=len(blob)), args=(blob_hash.bytes,))
-        return blob_hash
+            "INSERT INTO {tab} (hash, size) VALUES (%s, {size}) ON DUPLICATE KEY "
+            "UPDATE timestamp=CURRENT_TIMESTAMP".format(
```
guzman-raphael: To simplify the management of granting user privileges, can we use `INSERT IGNORE` or `REPLACE INTO` here instead depending on what is the intended behavior? The former would require only `INSERT` privilege, the ladder would require `INSERT, DELETE`. The current requires `INSERT, UPDATE`.

datajoint/external.py
dimitri-yatsenko: We always convert to absolute path before any operations.

datajoint/external.py
dimitri-yatsenko: We use `ON DUPLICATE KEY UPDATE` as a way to handle duplicates without ignoring other types of errors as would be the case with `INSERT IGNORE`. This feature is also used in `Table.insert`. We have to give people `UPDATE` privilege to allow `skip_duplicates` functionality.

datajoint/external.py
guzman-raphael: Per discussion, agreed that `os.path.relpath` implementation here would be an improvement.

datajoint/utils.py
guzman-raphael: Addressed

datajoint/external.py
guzman-raphael: Per discussion, need to research S3/Minio behavior on simultaneous updates and develop a test for this. Currently, a transaction is started prior to the copy, however, it is not committed so there is a risk.

datajoint/s3.py
guzman-raphael: Per discussion, this is correct but suggestion is that documentation should state that `filepath` data and `external blob` data are required to be in separate `stores`.

PR #611, opened 2019-05-16
Update blob packing + incorporate test user creation into nose

guzman-raphael:
Add blob packing support for np.bool_ and numpy array of string + integrate creation of test users into unit tests.

### DIFF
--- a/datajoint/blob.py
+++ b/datajoint/blob.py
@@ -146,7 +146,7 @@ class Blob:
             return self.pack_recarray(np.array(obj))
         if isinstance(obj, np.number):
             return self.pack_array(np.array(obj))
-        if isinstance(obj, (bool, np.bool)):
+        if isinstance(obj, (bool, np.bool, np.bool_)):
             return self.pack_array(np.array(obj))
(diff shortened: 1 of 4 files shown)


datajoint/blob.py
```diff
@@ -206,7 +206,8 @@ def pack_array(self, array):
         is_complex = np.iscomplexobj(array)
         if is_complex:
             array, imaginary = np.real(array), np.imag(array)
-        type_id = rev_class_id[array.dtype]
+        type_id = (rev_class_id[array.dtype] if array.dtype.char != 'U'
```
dimitri-yatsenko:
```suggestion
        type_id = rev_class_id[array.dtype if array.dtype.char != 'U' else np.dtype('O')]
```

datajoint/blob.py
```diff
@@ -206,7 +206,8 @@ def pack_array(self, array):
...
+                else rev_class_id[np.dtype('O')])
```
dimitri-yatsenko:
```suggestion
```

PR #612, opened 2019-05-20
Update Connection.connect() to deal with MYSQL 8.0

Synicix:
MySQL Server 8.0 completely removed support for the sql_mode = "NO_AUTO_CREATE_USER" thus causing it to fail.

This code creates a connection first without that mode flag and then checks if the server version is greater then 8.0, if not then close the current connection and reconnect with the flag.

This is due to the fact that in 8.0 they enforce the rule of being unable to create accounts via the GRANT command, which is equivalent to this flag always being on.

https://dev.mysql.com/doc/refman/5.7/en/sql-mode.html#sqlmode_no_auto_create_user

This might not be the cleanest way to this, so I am be happy to hear feed back.

### DIFF
--- a/datajoint/connection.py
+++ b/datajoint/connection.py
@@ -96,8 +96,7 @@ class Connection:
                 sql_mode="NO_ZERO_DATE,NO_ZERO_IN_DATE,ERROR_FOR_DIVISION_BY_ZERO,"
                          "STRICT_ALL_TABLES,NO_ENGINE_SUBSTITUTION",
                 charset=config['connection.charset'],
-                **self.conn_info)
-                
+                **self.conn_info)    
         self._conn.autocommit(True)
 
     def close(self):
--- a/datajoint/connection.py
+++ b/datajoint/connection.py
@@ -97,16 +97,7 @@ class Connection:
                          "STRICT_ALL_TABLES,NO_ENGINE_SUBSTITUTION",
                 charset=config['connection.charset'],
                 **self.conn_info)
-
-            # If the database version is older then 8.0, then include 'NO_AUTO_CREATE_USER'
-            if self._conn.server_version < '8.0.0':
-                self.close()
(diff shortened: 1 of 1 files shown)


datajoint/connection.py
```diff
@@ -92,6 +92,16 @@ def connect(self):
...
             warnings.filterwarnings('ignore', '.*deprecated.*')
             self._conn = client.connect(
+                init_command=self.init_fun,
+                sql_mode="NO_ZERO_DATE,NO_ZERO_IN_DATE,ERROR_FOR_DIVISION_BY_ZERO,"
+                         "STRICT_ALL_TABLES,NO_ENGINE_SUBSTITUTION",
+                charset=config['connection.charset'],
+                **self.conn_info)
+
+            # If the database version is older then 8.0, then include 'NO_AUTO_CREATE_USER'
+            if self._conn.server_version > '8.0.0':
```
dimitri-yatsenko:
```suggestion
            if self._conn.server_version < '8.0.0':
```

But should probably use version `distutil.version` for version comparisons. https://stackoverflow.com/questions/11887762/how-do-i-compare-version-numbers-in-python 

Also, perhaps we should just drop `NO_AUTO_CREATE_USER` for all MySQL or move `sql_mode` into `dj.config`.

dimitri-yatsenko: For now, would you just remove `NO_AUTO_CREATE_USER` altogether?

Synicix: We could, but that will allow people using SQL server ver < 8.0 to create users via grants command, which is not a major issue since normally people don't create users via GRANT command. I just wanted to get feed back to see if there was anything that actually depends on that flag that I don't know about.

Synicix: Also thanks for flipping that '>' I was testing the second version to see if works.

PR #613, opened 2019-05-20
Add blob convert utility with forward/backward support.

guzman-raphael:


dimitri-yatsenko:
I am afraid these conversion scripts will add more confusion than they will address. I suggest adding the following statement to this release and allow users to make adjustments to their code.  

> Release 0.12.0 adds support for all basic Python data types (lists, tuples, set, dict, str, bytes, and numeric scalars) and their combinations (e.g. list of str, set of numbers, etc).
Prior to Release 0.12.0, datajoint-python only supported MATLAB data types. For example, a dict would be converted into a 1x1 MATLAB struct array.  When fetching data back, datajoint would convert it back into a dict. In the new release, this conversion no longer occurs. Users can explicitly insert Matlab datatypes using objects of types `dj.MatStruct` and `dj.MatCell` and fetch them in the same form. For this reason, some previous code may break when dicts were expected and now `dj.MatStruct`would be retrieved. Please make adjustments to your code to handle such occurrances.

eywalker:
The statement is not quite accurate and should be corrected. Previously a `dict` would be inserted as record array and was also retrieved as record array. So the code they had would break for the new behavior in which a dict is inserted and dict is returned.

dimitri-yatsenko:
Thank you. I have fixed the comment. Let's close this PR and issue after incorporating the comment into the release notes.

PR #620, opened 2019-06-01
Add SSL

guzman-raphael:


datajoint/connection.py
```diff
@@ -57,14 +57,16 @@ class Connection:
     :param port: port number
     :param init_fun: connection initialization function (SQL)
     """
-    def __init__(self, host, user, password, port=None, init_fun=None):
+    def __init__(self, host, user, password, init_fun=None, ssl=True, port=None):
```
dimitri-yatsenko:
```suggestion
    def __init__(self, host, user, password, *, port=None, ssl=True,  init_fun=None):
```
Since we change the order of keyword arguments, let's add the * to force people to name their arguments after host, user, and password. All following attributes must be explicitly named.

tests/test_ssl.py
```diff
@@ -0,0 +1,28 @@
+from nose.tools import assert_true, assert_false, assert_equal, \
+                        assert_list_equal, raises
+import datajoint as dj
+from . import CONN_INFO
+from pymysql.err import OperationalError
+
+
+class TestSSL:
+
+    # @staticmethod
```
dimitri-yatsenko: Remove commented code before merging PRs.

datajoint/connection.py
```diff
@@ -17,7 +17,7 @@
 logger = logging.getLogger(__name__)
 
 
-def conn(host=None, user=None, password=None, init_fun=None, reset=False):
+def conn(host=None, user=None, password=None, init_fun=None, ssl=True, reset=False):
```
dimitri-yatsenko:
```suggestion
def conn(*, host=None, user=None, password=None, init_fun=None, ssl=True, reset=False):
```
Adding `*` to force callers to use keyword arguments.  
Perhaps `ssl` should be `use_ssl`?
What about the fact that SSL has evolved into TLS?

datajoint/connection.py
guzman-raphael: The only reason I reorder here is because I came across a bug between the `Connection` definition and how it was initialized in `dj.conn`. We can still preserve the order here if desired. But we should at least pass a `None` for port when initialized in `dj.conn`. After discussing with @ixcat and @ttngu207, the consensus is that the least impact to the users would be to keep the args ordered and add TLS as a last arg. Can proceed this way if you agree.

dimitri-yatsenko: By placing the `*`after password, we make the order of arguments insignificant since they will all have to be specified by name.

datajoint/connection.py
guzman-raphael: *See comment in `Connection` class init definition.
Actually, mentioning `ssl` here was my mistake, per MySQL [doc](https://dev.mysql.com/doc/refman/5.7/en/encrypted-connections.html), it actually only uses TLS. 

>  TLS is sometimes referred to as SSL (Secure Sockets Layer) but MySQL does not actually use the SSL protocol for encrypted connections because its encryption is weak. 

I can update `ssl` to `use_tls`, however, this will accept 3 values: `True`, `False`, or a dictionary of cert paths e.g. `{'ca': 'path/to/ca', 'capath': 'path/to/ca/dir', 'cert': 'path/to/client/pub/key', 'key': 'path/to/client/priv/key', 'cipher': 'cipher1,cipher2'}` per PyMySQL doc [here](https://pymysql.readthedocs.io/en/latest/modules/connections.html?highlight=ssl).

tests/test_ssl.py
guzman-raphael: Addressed

datajoint/connection.py
guzman-raphael: Addressed. Per discussion with the rest of the team, the preference seems to be to preserve the order and simply add option at the end to reduce impact on current users. Bug fix also applied.

datajoint/connection.py
guzman-raphael: Addressed. Utilizing `use_ssl` as it seems that this is still the standard for SSL/TLS connection option. We can document that it is in fact utilizing the more secure TLS. As interim documentation, the logic is as follows: 

`True` = Require TLS
`False` = Use only non-TLS
`'Unset'` = Best available with TLS as primary (default)
(python dict) = Implies required TLS and a straight pass-thru to PyMySQL.

datajoint/connection.py
```diff
@@ -57,14 +58,17 @@ class Connection:
...
-    def __init__(self, host, user, password, port=None, init_fun=None):
+    def __init__(self, host, user, password, port=None, init_fun=None, use_ssl='Unset'):
         if ':' in host:
             # the port in the hostname overrides the port argument
             host, port = host.split(':')
             port = int(port)
         elif port is None:
             port = config['database.port']
         self.conn_info = dict(host=host, port=port, user=user, passwd=password)
+        if not use_ssl == False:
```
dimitri-yatsenko:
```suggestion
        if use_ssl is not False:
```

datajoint/connection.py
```diff
@@ -57,14 +58,17 @@ class Connection:
...
+            self.conn_info['ssl'] = use_ssl if isinstance(use_ssl, dict) else {'ssl': {}}
+        self.conn_info['ssl_input'] = use_ssl
```
dimitri-yatsenko: change to TLS

datajoint/connection.py
```diff
@@ -17,7 +17,7 @@
 logger = logging.getLogger(__name__)
 
 
-def conn(host=None, user=None, password=None, init_fun=None, reset=False):
+def conn(host=None, user=None, password=None, init_fun=None, reset=False, use_ssl=None):
```
dimitri-yatsenko: change SSL to TLS throughout. SSL has been deprecated. It does not matter that others are not caught up yet.

datajoint/connection.py
guzman-raphael: Addressed

datajoint/connection.py
guzman-raphael: Addressed

datajoint/connection.py
```diff
@@ -17,7 +17,7 @@
 logger = logging.getLogger(__name__)
 
 
-def conn(host=None, user=None, password=None, init_fun=None, reset=False):
+def conn(host=None, user=None, password=None, init_fun=None, reset=False, use_tls=None):
     """
     Returns a persistent connection object to be shared by multiple modules.
```
dimitri-yatsenko: add `use_tls` to the docs string below.

guzman-raphael: Addressed

PR #624, opened 2019-06-12
Merge master and minor fix in docs

dimitri-yatsenko:


setup.py
```diff
@@ -29,4 +31,5 @@
     keywords='database organization',
     packages=find_packages(exclude=['contrib', 'docs', 'tests*']),
     install_requires=requirements,
+    python_requires='~={}.{}'.format(*min_py_version)
```
ixcat: does `~` work?
https://www.python.org/dev/peps/pep-0345/#requires-python
seems to indicate this should be `>=` or similar..

guzman-raphael: It is correct from my point of view. As it is, it will be interpreted that it will work for >=Python3.4 but will not automatically accept Python4, if that becomes a release. Here is a good reference:

> If your package is for Python 3.3 and up but you’re not willing to commit to Python 4 support yet, write:
python_requires='~=3.3',

[python_requires](https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires)

ixcat: Ah hmm.. 
Looks like the setup.py documentation/format strings differ from the 'package metadata ones';
The link I found was for the latter.. so right, looks good.

Issue #629, opened 2019-07-22
pyparsing 2.4.1 interop creates connection issues


ixcat:
The pyparsing 2.4.1 release seems to cause something in our code to hang for connections (at least; rest can't be tested w/o connection)

data points thus far:
- ubuntu 16.04 and system (pyparsing 2.0.3 via apt vs 2.4.1 in pip)
- datajoint-python travis builds
- personal test in ve (old 2.3.0 and 2.4.0 OK; 2.4.1 NOK)

dimitri-yatsenko:
This was a bug in `pyparsing.Or` introduced in `pyparsing 2.4.1`

ixcat:
looks like builds/unit tests are now ok after #631 : `https://travis-ci.org/datajoint/datajoint-python/builds/562698261`

@guzman-raphael @eywalker @dimitri-yatsenko who can cut a 0.11.3 release, or if someone wants to walk through I can do

thinking we close after this?

guzman-raphael:
@ixcat I can as long as everyone is in agreement, however, the one piece that I don't know yet is how to properly update the `CHANGELOG.md`. 

@eywalker anything special?

dimitri-yatsenko:
I have just updated both the CHANGELOG and the version number for both 0.11.2 and 0.12.rev5. Ready for release.

dimitri-yatsenko:
Fixed with #631

PR #630, opened 2019-07-22
requirements.txt: pyparsing<=2.4.0 interim fix for #629

ixcat:
Suggesting to apply this, issue datajoint 0.11.3 for master as solution for #629 
dev not impacted; leaving this for 'proper fix'

PR #631, opened 2019-07-23
fix pyparsing.Or bug introduced by pyparsing 2.4.1

dimitri-yatsenko:
Fix issue #629

### DIFF
--- a/datajoint/declare.py
+++ b/datajoint/declare.py
@@ -45,7 +45,7 @@ def build_foreign_key_parser():
 
 
 def build_attribute_parser():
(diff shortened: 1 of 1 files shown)


ixcat:
@dimitri-yatsenko - you mention in #630 this is backwards compatible, correct?

ixcat:
also, this is for master; need same for dev if not already present

guzman-raphael:
@ixcat I believe the purpose of #632 is to push this change to `dev` as well.

ixcat:
ah yes missed that :) think we are good here.. merging

PR #634, opened 2019-07-24
fix #633

### LINKED ISSUE #633: `:` in comments causes DDL parse issues in at least some cases
@shenshan reported an error with the following schema:

```@schema
class Cluster(dj.Imported):
    definition = """
    -> ProbeInsertion
    cluster_revision:               varchar(64)
    cluster_id:                     int
    ---
    -> ChannelGroup.Channel                         # peak channel for the cluster
    cluster_mean_waveform=null:     blob@ephys      # Mean unfiltered waveform of spikes in this cluster (but for neuropixels data will have been hardware filtered): nClusters*nSamples*nChannels
    cluster_template_waveform=null: blob@ephys      # Waveform that was used to detect those spikes in Kilosort, in whitened space (or the most representative such waveform if multiple templates were merged)
    cluster_depth:                  float           # Depth of mean cluster waveform on probe (µm). 0 means deepest site, positive means above this.
    cluster_waveform_duration:      blob@ephys      # trough to peak time (ms)
    cluster_amp:                    float           # Mean amplitude of each cluster (µV)
    cluster_phy_annotation=null:    tinyint         # 0 = noise, 1 = MUA, 2 = Good, 3 = Unsorted, other number indicates manual quality score (from 4 to 100)
    cluster_spike_times:            blob@ephys      # spike times of a particular cluster (seconds)
    cluster_spike_depth:            blob@ephys      # Depth along probe of each spike (µm; computed from waveform center of mass). 0 means deepest site, positive means above this
    cluster_spike_amps:             blob@ephys      # Amplitude of each spike (µV)
    """

```

error text:

```Unknown attribute type blob@ephys_external:Mean unfiltered waveform of spikes in this cluster (but for neuropixels data will have been hardware filtered)```

removing the comments allowed the schema to parse, drilling down on comments, the `:` in the definition of  `cluster_mean_waveform` seemed to be causing the issue; so it appears that the 'greedy match' at https://github.com/datajoint/datajoint-python/blob/95b138c9a19d7d15da60b17a3d7ee4f384826d4f/datajoint/declare.py#L52 could be the issue. Can issue patch but will take me a little bit to get up to speed on pyparsing & the parsing logic.. if there is someone else who can dive in this might yield a faster fix

### PR DESCRIPTION
**Pull request by dimitri-yatsenko**

### CODE REVIEW

File: datajoint/declare.py [hunk 1]
```diff
@@ -414,6 +414,9 @@ def compile_attribute(line, in_key, foreign_key_sql):
     match['comment'] = match['comment'].replace('"', '\\"')   # escape double quotes in comment
     category, type_match = match_type(match['type'])
 
+    if match['comment'].startswith(':'):
+        raise DataJointError('An attribute comment must not start with a colon in comment "{comment}"'.format(**match))
```
eywalker: why is it that a comment cannot start with `:`? Does this mean something like
```
id:  int    #:my number
```
is not valid?

ixcat: isn't the problem colon anywhere in the comment?

eywalker: Based on the new test case, it is OK to actually have `:` in the comment, but that's also why I'm not sure when it is that this gets triggered and why.

eywalker: Actually, I vaguely recall this having to do with the custom DataJoint datatype handling that uses tag of form `:external:` at the very beginning of the comment.

dimitri-yatsenko: That's right. We reserve comments that start with `:` to encode data types. We encode custom datatypes in the comment as `:custom-type: user comment`.

eywalker: OK got it and I recall agreeing to this strategy.

dimitri-yatsenko: Custom datatypes are encoded as `:custom-type: user comment` If it does not begin with a `:`, then it's fine.

dimitri-yatsenko: @ixcat, The actual fix was to fix the regexp that was previously greedily matching. The check above is fixing a separate problem from what you reported in #633

ixcat: Makes sense & Starting a comment with a colon would be an odd thing to do.. will change request / merge

Issue #627, opened 2019-07-02
Custom attribute data types


dimitri-yatsenko:
Release 0.12 adds support for many datatypes stored as blobs: all basic python types, numpy arrays, datetime.  This feature will add the ability to store custom user datatypes.

# Declaration syntax:

```python
@schema
class Person(dj.Manual):
    definition = """
    username : varchar(24)
    ---
    full_name : <full_name>
    cell_phone : <phone>
    photo = null : <photo>
    """
```

The custom datatypes are locally available identifiers pointing to objects of type `dj.AttributeType`. For security, types cannot be expressions or function calls. They can only be identifiers with dot notation allowed.

# AttributeType access objects
The datatypes must be defined as classes in the namespace accessible to the `schema` object similarly to the way foreign key resolution works.

The datatypes are subclasses of the abstract class `dj.AttributeType` and must define the property `datatype` and the methods `put()` and `get()`. DataJoint will processes inserted objects through the `put` method and fetched objects through the `get` method.

For example, the `Name` datatype could be declared as:
```python
class Phone(dj.AttributeType):
    """ 
    handles phones as 
    dict(country_code=1, area_code=813, number=5551234, extension=1234) 
    """
    def __init__(self, max_length=20):
        self.max_length = max_length

    @property
    def datatype(self):
        """ how the database will represent this datatype """
        return "varchar({})".format(self.max_length)

    def put(self, obj):
        """
        preprocess a phone storage
        :param obj: dict representing phone number
        """
        obj = "{country_code}-{area_code}-{number}.{extension}".format(**obj)
        assert len(obj) <= self.max_length, "phone number exceeds length"
        return obj

    def get(self, obj):
        country_code, area_code, number = obj.split('-')
        number, extension = number.split('.')
        return dict(country_code=country_code, area_code=area_code, number=number, extension=extension)

# create the datatype object
phone = Phone(max_length=30)
```

Custom datatypes can be used for data validation, for storing custom classes, etc.

Needless to say, correct datatypes must be defined when inserting and fetching the data. This applies also in schema reconstructed from the database using `schema.spawn_missing_classes` and `dj.create_virtual_module`. If the datatype objects are not available, the `insert` and `fetch` methods will raise an error when fetching the attribute but will allow fetching other attributes.

dimitri-yatsenko:
This is included in release 0.12 as an experimental feature requiring setting the environment variable
```
[...]

PR #639, opened 2019-08-04
Custom attribute type: implement #627

dimitri-yatsenko:
Implement user-defined attribute types as described in #627


dimitri-yatsenko:
Here are examples of use: https://nbviewer.jupyter.org/github/dimitri-yatsenko/db-programming-with-datajoint/blob/master/notebooks/Adapted-Types.ipynb

ixcat:
Was being a bit 'jesty' here since I'm working on docs for v0.12.. 

But: more generally should we *require* documentation upon adding a new feature?

aka the 'feature' is always 'code + user documentation'

this forces documentation to always be current, and also forces clarifying features since one has to be able to explain them coherently in human language. aka documentation is a 'human unit test'

dimitri-yatsenko:
Yes, we should update the documentation along with releases.  Attribute Adapters are an alpha feature for a specific lab and are not going to be officially supported until 0.12.1 at least. So no documentation yet.

PR #641, opened 2019-08-07
Add test cases for external blob migration 0.11 to 0.12

guzman-raphael:


tests/__init__.py
```diff
@@ -84,6 +111,81 @@ def setup_package():
...
+        CREATE DATABASE {};
+        """.format(db_name))
+
+    def parse_sql(filename):
+        data = open(filename, 'r').readlines()
+        stmts = []
+        DELIMITER = ';'
+        stmt = ''
+
+        for lineno, line in enumerate(data):
```
dimitri-yatsenko: `data` is unnecessary. You can loop directly over the text file even more efficiently.

tests/__init__.py
```diff
@@ -84,6 +111,81 @@ def setup_package():
...
+            if not line.strip():
```
dimitri-yatsenko: strip spaces from the line for the following checks too.

guzman-raphael: Hi Dimitri, shouldn't there be spaces between the command and arguments? If I also strip the spaces won't it cause the following `USE djtest_blob_migrate;\n` to become `USEdjtest_blob_migrate;`?

dimitri-yatsenko: I see. You are right. But for example, the next line `line.startswith('--')` will fail if there is a leading zero.

dimitri-yatsenko: Instead of adding up the strings you can be joining on `" "`.

tests/external-legacy-data/v0_11.sql
```diff
@@ -0,0 +1,138 @@
...
+
+DROP TABLE IF EXISTS `b`;
+/*!40101 SET @saved_cs_client     = @@character_set_client */;
+/*!40101 SET character_set_client = utf8 */;
+CREATE TABLE `b` (
+  `id` int(11) NOT NULL,
+  `blob_local` char(51) NOT NULL COMMENT ':external-local:uses files',
+  `blob_share` char(51) NOT NULL COMMENT ':external-shared:uses S3',
+  PRIMARY KEY (`id`),
+  KEY `blob_local` (`blob_local`),
```
dimitri-yatsenko: Interesting. These keys are redundant since the foreign key declaration below triggers their implicit creation.

guzman-raphael: Hey Dimitri, I see. I had not reviewed this script to closely as it was auto-generated by a `mysql dump ...` command. Is this required for us to merge on this feature? Trying to avoid changes on the generated file simply to streamline the process of creating/modifying this test.

dimitri-yatsenko: ok, no problem

Issue #648, opened 2019-09-03
handle ambiguous leading slashes in s3 locations


ixcat:
S3 buckets are strictly speaking, k:v stores, where the key is a string.

AWS S3 and min.io (possibly others) allow folder-based navigation in their web UI's based on '/' separation of key components.

However, min.io does not seem to allow leading '/' components, and the folder view for AWS results in confusing '//' paths. e.g. top level '/folder' becomes '//folder', with '/' being a viewable entry at top level.

Therefore, I propose always trimming leading slashes from S3 objects to allow minio-S3 usage, of top level bucket directories (e.g. `location=''`) for blob storage, and prevent the confusing case for aws-S3.

One potential downside could be that existing S3 configurations using leading slashes may regress, or not be usable by other non-blob S3 access patterns (e.g. filepath objects). These sorts of regressions could be fixed by users bulk moving 'moving' the objects in '//folderlike' to 'folderlike' via web gui.

dimitri-yatsenko:
I have not seen this. If double '/' appear anywhere, they should be eliminated. How do I reproduce this problem?

ixcat:
@dimitri-yatsenko , the steps to reproduce are:

- spin up minio
- set `'location': ''` in an s3 configuration

indeed removing all duplicate slashes might be even better..

ixcat:
Per PR in #649 will be treating as outside scope to 'normalize', but also will adjust paths to be generated without 'inherent', leading slashes, which work for top-level storage in both minio and s3. Using location='/' will yield the behavior described here as ambiguous.

PR #657, opened 2019-09-12
Issue 648

ixcat:
replacement for #649 to address #648


dimitri-yatsenko:
Issue #648 has been solved in #659 by switching to `pathlib` utilities from explicit string handling. I propose closing this PR.

ixcat:
sorry - reviewing the code - perhaps I'm missing - 
how does the new code prevent leading slashes?

```    
def _make_external_filepath(self, relative_filepath):
        """resolve the complete external path based on the relative path"""
        return PurePosixPath(self.spec['location'], relative_filepath)
```

with `'location' == '/` or == '' still yields '/':

```
>>> str(pathlib.PurePosixPath('/', 'relative/sub/path'))
'/relative/sub/path'
```

or am I missing something?

ixcat:
But yes, the refactoring in #659 would obviate this PR as a valid fix for #648; I just don't see how it fixes #648

dimitri-yatsenko:
```python
>>> str(Path('/', '/path')) 
'/path'
```

```python
>>> str(Path('', 'path')) 
'path'
```

If the user explicitly puts any invalid characters in the `location` path or in the filepaths, then of course, this will cause an error.

ixcat:
ok - but this has nothing to do with #648, which is talking about the fact that '/' is valid for AWS but not minio, and wanting to (or not) be consistent about what '/' as a 'location' means between the two? or, are you saying, we don't consider this within our scope?

ixcat:
The idea here was to force-remove '/' from both, since it doesn't work in minio, and yields 'odd'  navigation in s3, so allowing it (whether the backing supports it or not) provides little 'benefit' for users, and removing it makes it work for all cases

ixcat:
location `'/'` is more 'meaningful' than location `''`, or maybe that's my baggage. 

treating as outside scope would also make this irrelevant. probably should document though.

dimitri-yatsenko:
I thought the issue was the specifying the the store `location` to the empty string `""` would previously produce a leading slash. Now it doesn't. The fact that `minio` has different legal names that AWS is not something that DataJoint validates. We do not validate the path names. We simply pass them to the filesystem and s3 service without knowledge of it. If the user specifies and invalid name, then they will get an error, as expected.

ixcat:
ok makes sense - will treat as outside scope & document `location=''` as the best way to store things at top level.

PR #658, opened 2019-09-12
[WIP] docs-parts updates for v0.12

ixcat:


docs-parts/admin/5-blob-config_lang5.rst
```diff
@@ -1,3 +1,3 @@
 .. code-block:: python
 
-    >>> schema.external_table.clean_store('external-name')
+    >>> schema.external['external-raw'].clean_blobs()
```
dimitri-yatsenko: we are revising this implementation. A single `clean` method will clean all: blobs, attachments, and files.

ixcat: this was removed/combined with `lang4`, using 'delete' and the `delete_external_files` flag.

PR #673, opened 2019-10-18
Add DJ011 blob migration utility function, Test improvements, and minor updates

guzman-raphael:
- Add blob migration utility from DJ011 to DJ012 + added tests
- Provide test setup for secure `minio` tests + added test
- Bug Fix: Use of `delete_external_files=False` in `filepath` caused exception + added test
- Update `filepath` and `attach` to return path as string + added tests
- Hot Fix: Lock `networkx` version for now (#675)


datajoint/migrate.py
```diff
@@ -0,0 +1,157 @@
+import datajoint as dj
+from pathlib import Path
+import re
+from .utils import user_choice
+
+
+def migrate_dj011_external_blob_storage_to_dj012(migration_schema, store):
```
dimitri-yatsenko: add a doc string describing purpose and input arguments

guzman-raphael: Updated

datajoint/__init__.py
```diff
@@ -41,5 +41,6 @@
 from .attribute_adapter import AttributeAdapter
 from . import errors
 from .errors import DataJointError
+from .migrate import migrate_dj011_external_blob_storage_to_dj012
```
ixcat: not so important, not sure if we want to import by default..

guzman-raphael: Right, good point. This can be safely removed.

datajoint/migrate.py
```diff
@@ -0,0 +1,167 @@
...
+    Utility function to migrate external blob data from 0.11 to 0.12.
+    :param migration_schema: string of target schema to be migrated
+    :param store: string of target dj.config['store'] to be migrated
+    """
+    if not isinstance(migration_schema, str):
+        raise ValueError(
+                'Expected type {} for migration_schema, not {}.'.format(
+                str, type(migration_schema)))
+
+    do_migration = False
```
dimitri-yatsenko: this line is unnecessary.

PR #674, opened 2019-10-21
Add support for secure S3 connection. Fix #663

### LINKED ISSUE #663: The `secure` arg has to be configurable in dj.config['store']
https://github.com/datajoint/datajoint-python/blob/1e0e4b2f7500779588f5a8cde93a2cfd72e9e346/datajoint/s3.py#L15

In some cases, perhaps MinIO server's configuration requires secure connection

Configurable per "store"

eywalker:
Add a new optional configuration for S3 protocol storage called `secure` to control  whether S3 connection should be secure or not. Defaults to `False`, and is configured for each storage separately.

Fix #663

### DIFF
--- a/datajoint/s3.py
+++ b/datajoint/s3.py
@@ -13,7 +13,7 @@ class Folder:
     """
     A Folder instance manipulates a flat folder of objects within an S3-compatible object store
     """
-    def __init__(self, endpoint, bucket, access_key, secret_key, secure=False, **_):
+    def __init__(self, endpoint, bucket, access_key, secret_key, *, secure=False, **_):
         self.client = minio.Minio(endpoint, access_key=access_key, secret_key=secret_key, secure=secure)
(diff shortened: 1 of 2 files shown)


datajoint/s3.py
```diff
@@ -13,8 +13,8 @@ class Folder:
     """
     A Folder instance manipulates a flat folder of objects within an S3-compatible object store
     """
-    def __init__(self, endpoint, bucket, access_key, secret_key, **_):
-        self.client = minio.Minio(endpoint, access_key=access_key, secret_key=secret_key, secure=False)
+    def __init__(self, endpoint, bucket, access_key, secret_key, secure=False, **_):
```
dimitri-yatsenko:
```suggestion
    def __init__(self, endpoint, bucket, access_key, secret_key, *, secure=False, **_):
```

PR #679, opened 2019-10-25
corrections for using `enable_python_native_blobs` in README.

dimitri-yatsenko:


README.md
```diff
@@ -24,15 +24,29 @@ pip3 install --upgrade datajoint
...
+
+```python
+import datajoint as dj
+dj.config["enable_python_native_blobs"] = True
+```
+
+You can safely enable this setting if both of the following are true:
+
+  * All blobs in your current DataJoint databases contain only numerical arrays.
+  * You do not need to share blob data between Python and Matlab
```
eywalker: Could you make Matlab -> MATLAB here?

README.md
```diff
@@ -45,22 +59,18 @@ and Python for certain record types. However, this created a discrepancy
...
-`enable_python_native_blobs` flag was added as a temporary guard measure
-for the 0.12 release. This flag will trigger an exception if any of the
-ambiguous cases are encountered during inserts in order to allow testing
+appropriate native python type (dict, etc).  
+Furthermore, DataJoint for MATLAB does not yet support unpacking native Python datatypes.
+
+With `dj.config["enable_python_native_blobs"]` set to `False` (default), 
+any attempt to insert any datatype other than a numpy array will result in an exception.
+This is meant to get users to read this message in order to allow proper testing
 and migration of pre-0.12 pipelines to 0.11 in a safe manner.
```
eywalker: This should say `and migration of pre-012 pipelines to 0.12 in a safe manner`

README.md
```diff
@@ -24,58 +24,67 @@ pip3 install --upgrade datajoint
...
+Since this is a big change, we chose to disable full blob support by default as a temporary precaution, which will be removed in version 0.13.
+
+You can enable it by setting the `enable_python_native_blobs` flag in `dj.config`. 
+
+```python
+import datajoint as dj
+dj.config["enable_python_native_blobs"] = True
+```
+
+You can safely enable this setting if both of the following are true:
```
ixcat: I like the new list of improvements, etc at the intro -

However, since this is a migration note, we did discuss having some sort of 'quick skim list' at the top of the section for the people who don't want to read anything and just use the new feature, and to make hitting this error less intimidating for new users.

The new initial text is presented more positively, which I think is good, but pushes down the 'please set this setting if XYZ doesnt apply' block from the beginning.

README.md
```diff
@@ -24,58 +24,67 @@ pip3 install --upgrade datajoint
...
+
+  * All blobs in your current DataJoint databases contain only numerical arrays.
+  * You do not need to share blob data between Python and MATLAB.
```
ixcat: Why did the 'is a new pipeline'/'haven't used < 0.12' cases get removed?

Also: technically the 'squashed' data is 'a numerical array', and we haven't explained the details of squashing yet, so, devoid of details, making a statement based on an understanding of 'contain numerical arrays' is potentially dangerously ambiguous.

Put another way: the previous list outlined situations where the details of the disclaimer may need to be read, the new list outlines the cases where the data is correct, but we haven't described what those cases are cases yet..

datajoint/blob.py
```diff
@@ -73,7 +73,7 @@ def __init__(self, squeeze=False):
 
     def set_dj0(self):
         if not config.get('enable_python_native_blobs'):
-            raise DataJointError('v0.12+ python native blobs disabled. see also: https://github.com/datajoint/datajoint-python/blob/master/README.md')
+            raise DataJointError('v0.12+ python native blobs disabled. see also: https://github.com/datajoint/datajoint-python#python-native-blobs')
```
ixcat: While this is a more direct link, it also disambiguate the referral to the underlying README.md in the sources. Not sure what is better.

README.md
dimitri-yatsenko: Yes, I think it's important to explained what happened with a more accurate and complete, not only more positive explanation of what happened. The original explanation started mid-sentence without providing a context.

README.md
dimitri-yatsenko: Having a new pipeline is not a necessary or sufficient requirement. It was not clear in the original formulation if all or any of the following conditions were required. I made the list so that it's **all** of the following conditions.

datajoint/blob.py
dimitri-yatsenko: the link is meant t point to the specific section

README.md
ixcat: This is not what I'm getting at - change can be fine, but it changes the intent of the block.

Previous intent: 

list the limited subset of cases for people who will never need to understand any details whatsoever, specifically intended to allow 'easy and not scary enable' strictly for new users who might be intimidated by the exception.  Everyone else should read the full details.

(So the 'insufficient' you mention was by design, to force the 'fall through' case of reading full and complete details for everyone but new-python-only users) 

Current intent (I surmise, please forgive): 

Summary of details for all users, if desired can read further

Are we confirming the change in intent?

If so, fine. 

However, if so, I don't think 'contain numerical arrays' is 'scary' enough to force reading the full details for those that should - Much like the phrase 'the following are true', 'contain numerical arrays' is  completely and precise in referring to a specific situation, but casually can be confusing without extra clarification. Since 'numpy array' is often used to casually refer to anything in the numpy package, and any of these things, including structured arrays are incorrectly used interchangably with the phrase 'numerical array' casually, I think we need be a bit more explicit here.

So, 2 things are needed for this discussion:

1) Confirm change in intent (apologies in harping, but this was not explicitly reflected in feedback)

2) Agree/Disagree on extending 'contain numerical array' text. If we are changing the intent, I would propose 'contain purely numerical arrays (see below for details)' to help provoke full reading by those who need to.

datajoint/blob.py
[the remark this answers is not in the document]
dimitri-yatsenko: not sure how this is better

README.md
dimitri-yatsenko: we addressed this on Slack

PR #659, opened 2019-09-14
Store attachments as intact files

dimitri-yatsenko:
* Fix #637 
* Fix #648
* Fix #650

PR #693, opened 2019-11-08
Strip root from external S3 storage on insert

### LINKED ISSUE #692: DJ `filepath` inconsistent handling of read/write of `location` with leading slash on S3 protocol
Example store config:
        {
            "protocol": "s3",
            "endpoint": "s3.amazonaws.com",
            "access_key": "--------------------------",
            "secret_key": "--------------------------",
            "bucket": "report_bucket",
            "location": "/data/report",
            "stage": "F:/data/report"
        }

On S3, upon insert, the path got written will be something like `report_bucket//data/report/...` (keeping the leading slash from `location`). 
But when fetching, the leading slash got stripped and the path DJ would search for is actually something like `report_bucket/data/report/...`

guzman-raphael:
Fix #692

Currently, on `fetch` the root of the `location` from `s3` storage is stripped but not on `insert`. This change is to align both so that `insert` and `fetch` both strip the root for `s3` storage. This makes it compatible with all S3 providers e.g. Minio does not allow objects on root of `bucket` but AWS hosted objects are allowed. 

`file` storage, however, will still allow `insert` on root.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## Release notes
 
+### 0.12.2 -- Nov 8, 2019
+* Bugfix - Insert into external does not trim leading slash if defined in `dj.config['stores']['<store>']['location']` (#692)
+
 ### 0.12.1 -- Nov 2, 2019
 * Bugfix - AttributeAdapter converts into a string (#684)
 
--- a/datajoint/external.py
+++ b/datajoint/external.py
@@ -1,4 +1,4 @@
-from pathlib import Path, PurePosixPath
+from pathlib import Path, PurePosixPath, PureWindowsPath
 from collections import Mapping
 from tqdm import tqdm
 from .settings import config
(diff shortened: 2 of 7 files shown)


ixcat:
I didn't catch this at the time - 

Wasn't the previous resolution to simply 'not manipulate' location - e.g. if user provides leading slash, use leading slash, if not, dont? This was the previous policy in the resolution to https://github.com/datajoint/datajoint-python/issues/648

Previous PR against old code structure (https://github.com/datajoint/datajoint-python/pull/657) was implemented in an 'always-strip' policy, but was closed in favor of the "don't manipulate" policy which came in in https://github.com/datajoint/datajoint-python/pull/659

I'm OK either way as to the resulting behavior, but this seems like a reversion to one mode of behavior which was previously explicitly overturned ..

guzman-raphael:
@ixcat The main issue this PR aimed to solve was that although DataJoint 0.12.1 permitted users to store with external `location` set as `/lab`, records were un`fetch`able as the logic did not match. This is b/c `insert`s would place records in S3 as `bucket/schema//lab/object` but attempted to `fetch` as `bucket/schema/lab/object`.

Therefore, the logic for the `insert` only was modified to match. Additionally, this serves to solve the concerns in #648 for `s3` (preserving compatibility between AWS S3 and Minio) while still letting `file` stores store as they please on the root. Also, this satisfies Dimitri's concerns from #648 "If double '/' appear anywhere, they should be eliminated". 

Now this will require modification of S3 objects if they were already inserted with 0.12.1 under this use case, however, I would argue this was necessary in any case since the underlying bug did not allow these records to be fetched.

ixcat:
Thanks @guzman-raphael - this clears it up / makes sense 👍

PR #705, opened 2019-11-19
Fix #675 

### LINKED ISSUE #675: networkx 2.4 regression
We've had reports of ERD problems in some cases with NetworkX 2.4 -

These should be fixed, and ideally improved unit testing to test more ERD cases.

dimitri-yatsenko:


eywalker:
Does this now require networkx>=2.4? If so, please adjust the requirements

dimitri-yatsenko:
> Does this now require networkx>=2.4? If so, please adjust the requirements

No, it does not. This was their correct API before 2.4

tests/test_jobs.py
```diff
@@ -56,10 +54,9 @@ def test_restrictions():
     jobs.error('a', {'key': 'a2'}, 'error')
     jobs.error('b', {'key': 'b1'}, 'error')
 
-    assert_true(len(jobs & 'table_name = "a"') == 2, 'There should be two entries for table a')
-    assert_true(len(jobs & 'status = "error"') == 2, 'There should be two entries with error status')
-    assert_true(len(jobs & 'table_name = "a"' & 'status = "error"') == 1,
-                'There should be only one entries with error status in table a')
+    assert_true(len(jobs & {'table_name': "a"}) == 2)
```
eywalker: why did you get rid of the error message here?

dimitri-yatsenko: because it carried no useful information. The message is not useful without looking at the code. If do look at the code, then the assert condition is just as readable as the error message.

eywalker: Yeah fair

PR #706, opened 2019-11-19
Fix #698 and #699

### LINKED ISSUE #698: help does not work on tables
The builtin `help` function does not seem to work on tables.
`help(MyTable)` raises `DataJointError: Class UserTable is not properly declared (schema decorator not applied?)`, all other table functionality being intact.

### LINKED ISSUE #699: Append table definition to docstring
Thinking about improving the documentation of our own database, we thought it would be nice to add the table definition to its docstring. We do this using a decorator on the table definition somewhat like this:
```python
def docappend(cls):
    cls.__doc__ += cls.definition
    return cls

@docappend
@schema
class MyTable(dj.Imported):
    .....
```
To us, it seemed like a minor but useful feature to include in a future version of datajoint.

dimitri-yatsenko:
* Fix #698 -- Help now works for DataJoint classes
* Fix #699 -- table definition is appended to doc string

### DIFF
--- a/datajoint/blob.py
+++ b/datajoint/blob.py
@@ -253,6 +253,9 @@ class Blob:
     def read_sparse_array(self):
         raise DataJointError('datajoint-python does not yet support sparse arrays. Issue (#590)')
 
+    def read_scalar(selfs):
+        
+
     def read_decimal(self):
         return Decimal(self.read_string())
 
--- a/datajoint/blob.py
+++ b/datajoint/blob.py
@@ -152,10 +152,8 @@ class Blob:
             return self.pack_array(np.array(obj))
         if isinstance(obj, (bool, np.bool, np.bool_)):
             return self.pack_array(np.array(obj))
-        if isinstance(obj, float):
-            return self.pack_array(np.array(obj, dtype=np.float64))
-        if isinstance(obj, int):
-            return self.pack_array(np.array(obj, dtype=np.int64))
+        if isinstance(obj, (float, int, complex)):
+            return self.pack_array(np.array(obj))
         if isinstance(obj, (datetime.datetime, datetime.date, datetime.time)):
             return self.pack_datetime(obj)
         if isinstance(obj, Decimal):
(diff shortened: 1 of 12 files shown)


tests/test_relation.py
```diff
@@ -36,6 +36,12 @@ def setup_class(cls):
         cls.img = schema.Image()
         cls.trash = schema.UberTrash()
 
+    def test_class_help(self):
+        help(schema.TTest)
```
ixcat: would it make sense to also e.g.:
`assert schema.TTest.definition in schema.TTest.__doc__`

tests/test_relation.py
```diff
@@ -36,6 +36,12 @@ def setup_class(cls):
...
+
+    def test_instance_help(self):
```
ixcat: as before r.e. testing for definition in doc.

tests/test_relation.py
dimitri-yatsenko: done. Also moved into `test_declare`

tests/test_relation.py
dimitri-yatsenko: done. Also moved into `test_declare`

datajoint/table.py
```diff
@@ -45,14 +45,9 @@ def heading(self):
         """
         if self._heading is None:
             self._heading = Heading()  # instance-level heading
-        if not self._heading:  # lazy loading of heading
-            if self.connection is None:
-                raise DataJointError(
-                    'DataJoint class is missing a database connection. '
-                    'Missing schema decorator on the class? (e.g. @schema)')
```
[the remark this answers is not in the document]
dimitri-yatsenko: If the table is not decorated, with a `@schema`, then another error is already raised in `user_tables.py` when attempting to access the `database` property. Here we must be able to return a `None` connection object for the help to be properly generated. Help actually instantiates parent classes and this was one of the reasons help would raise an error.

guzman-raphael:
@dimitri-yatsenko Can you also update `docs-parts/intro/Releases_lang1.rst` to match `CHANGELOG.md`?

datajoint/blob.py
```diff
@@ -253,6 +253,9 @@ def pack_recarray(self, array):
     def read_sparse_array(self):
         raise DataJointError('datajoint-python does not yet support sparse arrays. Issue (#590)')
 
+    def read_scalar(selfs):
+
```
guzman-raphael: This looks like a typo... Does not raise/return anything?

PR #709, opened 2019-11-22
fix #690 -- blob packing/unpacking of native python bool, int, float, and complex.

### LINKED ISSUE #690: Python integers turn into Numpy integers when fetched.
In my table exist a non-primary column of type `longblob` that stores dictionaries.  Inside the dictionary some of the values are Python integers. However, when fetched, these Python integers turn into Numpy integers. It would be helpful if the types are kept the same at insert and fetch time.

DataJoint version = 0.12.0

Below is a toy example:
``` Python
@schema
class MyTable(dj.Manual):
    definition = """
    prim_key: varchar(32)
    ---
    attr: longblob
    
    """

MyTable().insert1(dict(prim_key=0, attr=dict(a=1)))

print(type(MyTable().fetch1()['attr']['a']))

# the output is <class 'numpy.int64'>
```

dimitri-yatsenko:
… and complex

### DIFF
--- a/datajoint/blob.py
+++ b/datajoint/blob.py
@@ -221,7 +221,7 @@ class Blob:
         if is_complex:
             array, imaginary = np.real(array), np.imag(array)
         type_id = (rev_class_id[array.dtype] if array.dtype.char != 'U'
-                else rev_class_id[np.dtype('O')])
+                   else rev_class_id[np.dtype('O')])
         if dtype_list[type_id] is None:
             raise DataJointError("Type %s is ambiguous or unknown" % array.dtype)
 
@@ -264,11 +264,13 @@ class Blob:
         raise DataJointError('datajoint-python does not yet support sparse arrays. Issue (#590)')
 
     def read_int(self):
-        return int.from_bytes(self.read_value('int64'), byteorder='little', signed=True)
+        return int.from_bytes(self.read_binary(self.read_value('uint16')), byteorder='little', signed=True)
 
     @staticmethod
     def pack_int(v):
-        return b"\x0a" + v.to_bytes(v.bit_length() // 8 + 1, byteorder='little', signed=True)
+        n_bytes = v.bit_length() // 8 + 1
+        assert n_bytes <= 0xFFFF, 'Integers are limited to 65535 bytes'
+        return b"\x0a" + np.uint16(n_bytes).tobytes() + v.to_bytes(n_bytes, byteorder='little', signed=True)
 
     def read_bool(self):
         return bool(self.read_value('bool'))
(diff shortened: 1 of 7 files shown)


datajoint/blob.py
```diff
@@ -251,6 +263,34 @@ def pack_recarray(self, array):
     def read_sparse_array(self):
         raise DataJointError('datajoint-python does not yet support sparse arrays. Issue (#590)')
 
+    def read_int(self):
+        return int(self.read_value('int64'))
+
+    @staticmethod
+    def pack_int(v):
+        return b"\x0a" + np.array(v, dtype='int64').tobytes()
```
guzman-raphael: Is there a reason why we did not utilize decimal packing here? Python `int` are essentially boundless (memory-dependent). I believe `decimal` packing would be a closer representation as the length would be encoded.

datajoint/blob.py
```diff
@@ -251,6 +263,34 @@ def pack_recarray(self, array):
...
+    @staticmethod
+    def pack_complex(v):
+        return b"\x0c" + np.array(v, dtype='complex128').tobytes()
+
+    def read_float(self):
+        return float(self.read_value('float64'))
+
+    @staticmethod
+    def pack_float(v):
+        return b"\x0d" + np.array(v, dtype='float64').tobytes()
```
guzman-raphael: Is there a reason why we did not utilize decimal packing here? Python `float` have a precision of 53 bits which means we would be storing unnecessary additional data.

datajoint/blob.py
```diff
@@ -251,6 +263,34 @@ def pack_recarray(self, array):
...
+    @staticmethod
+    def pack_bool(v):
+        return b"\x0b" + np.array(v, dtype='bool').tobytes()
+
+    def read_complex(self):
+        return complex(self.read_value('complex128'))
+
+    @staticmethod
+    def pack_complex(v):
+        return b"\x0c" + np.array(v, dtype='complex128').tobytes()
```
guzman-raphael: We could utilize `decimal` packing here for the same reasons as `float` below. Python seems to capture the first 53 bits for each the `real` part and the `complex` part.

dimitri-yatsenko: here Python is not doing anything special and just uses the standard IEEE 754 encoding.

datajoint/blob.py
dimitri-yatsenko: ok, added encoding of unbounded integers.

datajoint/blob.py
dimitri-yatsenko: modified to support unbounded `int`

guzman-raphael:
@dimitri-yatsenko Can you update `datajoint-python/docs-parts/intro/Releases_lang1.rst`?

PR #713, opened 2019-12-16
Update Log table

guzman-raphael:
- Update Log table to auto increment entries to prevent collisions and increase timestamp precision to millisecond.
- Add docs that describe how to run tests locally.

### DIFF
--- a/datajoint/declare.py
+++ b/datajoint/declare.py
@@ -12,7 +12,7 @@ from .utils import OrderedDict
 
 UUID_DATA_TYPE = 'binary(16)'
 MAX_TABLE_NAME_LENGTH = 64
-CONSTANT_LITERALS = {'CURRENT_TIMESTAMP', 'CURRENT_TIMESTAMP(3)'}  # SQL literals to be used without quotes (case insensitive)
+CONSTANT_LITERALS = {'CURRENT_TIMESTAMP'}  # SQL literals to be used without quotes (case insensitive)
 EXTERNAL_TABLE_ROOT = '~external'
 
 TYPE_PATTERN = {k: re.compile(v, re.I) for k, v in dict(
@@ -443,7 +443,8 @@ def compile_attribute(line, in_key, foreign_key_sql, context):
         match['default'] = 'DEFAULT NULL'  # nullable attributes default to null
     else:
         if match['default']:
-            quote = match['default'].upper() not in CONSTANT_LITERALS and match['default'][0] not in '"\''
+            quote = (match['default'].split('(')[0].upper() not in CONSTANT_LITERALS
+                        and match['default'][0] not in '"\'')
             match['default'] = 'NOT NULL DEFAULT ' + ('"%s"' if quote else "%s") % match['default']
         else:
             match['default'] = 'NOT NULL'
--- a/datajoint/declare.py
+++ b/datajoint/declare.py
@@ -12,7 +12,7 @@ from .utils import OrderedDict
 
 UUID_DATA_TYPE = 'binary(16)'
 MAX_TABLE_NAME_LENGTH = 64
-CONSTANT_LITERALS = {'CURRENT_TIMESTAMP'}  # SQL literals to be used without quotes (case insensitive)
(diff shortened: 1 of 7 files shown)


ixcat:
While I like this idea as found from log lock debug session, we probably need to consider the release managment side since we'll need to do some kind of migration for existing schemas (also: does this apply to matlab?)

guzman-raphael:
@ixcat Good news regarding MATLAB is that we have not implemented `~log` table yet. Regarding migration, yes this will come up. One alternative, if needed, is we could also use the following definition for `~log` table:
```
timestamp = CURRENT_TIMESTAMP(3) : timestamp(3)
---
version  :varchar(12)   # datajoint version
user     :varchar(255)  # user@host
host=""  :varchar(255)  # system hostname
event="" :varchar(255)  # custom message
```

This would reduce the chance of collisions without migration, however, the possibility would still exist.

datajoint/declare.py
```diff
@@ -12,7 +12,7 @@
 
 UUID_DATA_TYPE = 'binary(16)'
 MAX_TABLE_NAME_LENGTH = 64
-CONSTANT_LITERALS = {'CURRENT_TIMESTAMP'}  # SQL literals to be used without quotes (case insensitive)
+CONSTANT_LITERALS = {'CURRENT_TIMESTAMP', 'CURRENT_TIMESTAMP(3)'}  # SQL literals to be used without quotes (case insensitive)
```
dimitri-yatsenko: hm... we should find a way to support `CURRENT_TIMESTAMP` with arbitrary precision rather than hardcode `(3)`.

datajoint/table.py
```diff
@@ -683,8 +683,9 @@ def __init__(self, arg, database=None):
         self.database = database
         self._connection = arg
         self._definition = """    # event logging table for `{database}`
-        timestamp = CURRENT_TIMESTAMP : timestamp
+        id :  smallint auto_increment
```
dimitri-yatsenko: `smallint` may not be enough.

datajoint/declare.py
ixcat: does MySQL itself do this? or is this future proofing?

dimitri-yatsenko: Yes but our parser does not.

datajoint/table.py
dimitri-yatsenko: Can we keep the timestamp primary key with the higher precision or is collision probability still high?

guzman-raphael: Is there a reason why we shouldn't use an incremental primary key here? I would expect this to be the standard to prevent logging collisions. As far as an example use case: On our flask web deployments, I could see possible collisions near the millisecond level as the community grows. In the past I've used `bigint` for tables similar to this; just wasn't sure if we wanted to go there yet.

datajoint/declare.py
guzman-raphael: Do you mean as in a `my.cnf` setting? The default is second. If I declare a table and insert as follows:

``` MySQL
create table sample (dt timestamp(3));
insert into sample values (CURRENT_TIMESTAMP);
insert into sample values (NOW());
```

The result are only timestamps with second resolution.
``` MySQL
select dt from sample;
```
```
+-------------------------+
| dt                      |
+-------------------------+
| 2019-12-18 20:54:54.000 |
| 2019-12-18 20:54:56.000 |
+-------------------------+
2 rows in set (0.00 sec)
```

datajoint/table.py
dimitri-yatsenko: yes, it's fine to go with autoincrement or `serial` datatype. There is not that much logging going on in datajoint, but I agree that we may as well avoid any possibility of collisions.

datajoint/declare.py
dimitri-yatsenko: I think we should modify the parsing code to allow specifying function calls as default values. Currently our parser will reject `CURRENT_TIMESTAMP(2)` for example.

guzman-raphael: @dimitri-yatsenko Ah, I see your point. I have now updated the parsing to accommodate this.

datajoint/table.py
guzman-raphael: Had not come across `serial` but it looks ideal here, thanks! I have now updated the changes to leverage it.

PR #714, opened 2020-01-17
Handle config for DJ Hub

### LINKED ISSUE #715: DataJoint should facilitate work with database hosting services or "hubs"
In the event that DataJoint is utilized against a scalable deployment of database hosting services or `hubs`, requirements are as follows:

- No subdomain will be dedicated for databases instances.
- Allow host to be specified such that it queries a DJHub API target e.g. parse host config if matching pattern `hub://<DJHub_host>/<organization>/<project>`
- Requests will be unauthenticated (for now) since we assume that `projects` will be initially all public. Once there is interest in private, we will add authentication on requests to those specific `projects`. Similar to how the GitHub REST API behaves.
- Configuration should allow flexibility for backend in case we need to relocate servers to different hosts and/or modify ports.
- Requests should be secure and enforce TLS connections (Database connection treated separately).
- Allow room for expansion of API as it grows to provide more controls and access functions on DJHub data.
- Include helpful error handling
- Handle use case when a `project` has been relocated to a new endpoint in backend
- Specific request/response for initial API target was agreed as follows:
```
Host: <DJHub_host>
Route: /v0.000/get_database_fqdn
Method: POST
Body: {"org": "<organization>", "project": "<project>"}
Response: {"database.host": "<database_server>:<database_port>"}
```

Note: The above assumes best practice of backend hosting service with a suggested implementation.

guzman-raphael:
Fix #715
- Allow target host to be specified as one hosted by DJ Hub e.g. `hub://djhub.io/datajoint/university`.
- Backend `project` resources are retrieved via `GET` @ `/api/projects/<project_name>?org_name=<org_name>`
- Add logic if server becomes unavailable. Such a case would re-query DJ Hub API for latest endpoint (once per server-side change).
- Add tests


datajoint/hub.py
```diff
@@ -0,0 +1,43 @@
...
+session.mount('http://', HTTPAdapter(max_retries=3))
+
+
+def get_host(host):
+    hub_path = re.findall('/[:._\-a-zA-Z0-9]+', host)
+    if re.match(HUB_PROTOCOL, host) and len(hub_path) > 2:
+        try:
+            resp = session.post('{}{}{}{}'.format(
+                REQUEST_PROTOCOL, hub_path[0][1:], API_ROUTE, API_TARGETS['GET_DB_FQDN']),
+                data={'org': hub_path[1][1:], 'project': hub_path[2][1:]}, timeout=10)
```
ixcat: Can we change these to org_name and project_name?
Thinking that's cleaner w/r/t attribute naming on the backend.

datajoint/hub.py
```diff
@@ -0,0 +1,43 @@
+import re
+import requests
+from requests.adapters import HTTPAdapter
+from datajoint.errors import DataJointError
+
+HUB_PROTOCOL = 'hub://'
+REQUEST_PROTOCOL = 'https://'
+API_ROUTE = '/api'
+API_TARGETS = dict(GET_DB_FQDN='/v0.000/get_database_fqdn')
```
ixcat: Let's change to /projectdatabaseinstance/get to make it more 'resty'

guzman-raphael: @ixcat Actually, if the aim is to make this more 'RESTy', we should likely modify this from a `POST` to a `GET` since we are performing a 'read-only' operation. Usually, the method dictates the operation. If parameters need to be supplied, they can be passed as query params e.g. a `GET` on `/v0.000/projectdatabaseinstance?org_name=datajoint&project_name=travis`. If you agree, I can make the updates.

datajoint/hub.py
guzman-raphael: Makes sense. I will include this into the next commit.

datajoint/hub.py
ixcat: Yes - this is true. Some background on why this was implemented this way is that we had problems in IBL for 'subsets' in using a pure GET/'HTTP Verb' approach since it wasn't possible to construct >1x values to filter a query request for e.g. both 'project1' and 'project2' against the 'root' node (GET parameters get squashed - e.g. 2x 'project=X' will only pass one value to the handler) and this doesn't flatten to a clean URL either.

In those cases we set a single value in `_json` post variable which was then decoded to perform the N-valued restriction This doesn't directly apply here since we are targeting only one object at a time, but that usage might come in handy later (e.g. 'show me the databases for this subset of projects').. not sure what's best as an overall approach. A middle ground might be to offer fallback to a post queries for multi-valued filtering and allow GET filtering for the normal case. One difference from IBL worth noting for the discussion is that this service will likely grow to be read-write, or at least some other similar API will grow to be read-write, so we'll need to find a suitable mechanism for read-write cases either way.

Pinging @eywalker for feedback

PR #725, opened 2020-01-27
Fix #716

### LINKED ISSUE #716: SSL KeyError on failed connection
Simon at NTNU reported `KeyErrors` related to the 

> the ‘ssl_input’ line

near where `dj.conn()` is called.

Issue is likely related to use of `pop` as it is implemented to remove keys for connection scenarios then add them back once established. For Simon's case, connection likely was interrupted on client-side prior to key was added back into `conn_info` in connection object. Such would explain KeyError in succeeding attempts.

Solution should be to not use `pop` and specify the keys directly to be included with connection scenarios.

guzman-raphael:


datajoint/connection.py
```diff
@@ -145,17 +144,17 @@ def connect(self):
...
                 self._conn = client.connect(
                     init_command=self.init_fun,
                     sql_mode="NO_ZERO_DATE,NO_ZERO_IN_DATE,ERROR_FOR_DIVISION_BY_ZERO,"
                              "STRICT_ALL_TABLES,NO_ENGINE_SUBSTITUTION",
                     charset=config['connection.charset'],
-                    **self.conn_info)
-        self.conn_info['ssl_input'] = ssl_input
+                    **{k: v for k, v in self.conn_info.items()
+                    if (k != 'ssl_input') and (
+                    self.conn_info['ssl_input'] is not None or k != 'ssl')})
```
dimitri-yatsenko: I would find the following form more readable
```python
if not(k == 'ssl_input' or k == 'ssl' and self.conn_info['ssl_input'] is None)
```

PR #730, opened 2020-02-02
Plugin Infrastructure

### LINKED ISSUE #715: DataJoint should facilitate work with database hosting services or "hubs"

### LINKED ISSUE #729: Plugin Infrastructure
DataJoint should allow extensibility by means of 'plugins'.
Users should be able to develop hooks that can be leveraged by `datajoint-python` once pip installed.

Details forthcoming...

guzman-raphael:
- Allow pip installations that register as `datajoint.plugin` i.e. a DataJoint plugin package, to be detected by `datajoint-python` and loaded when appropriate.
- In the event that users have multiple plugins installed that conflict, allow user to specify which plugin should be imported via `datajoint.config['plugin']`. Omission of configuration will result in no restriction applied. Example config:
``` json
{
     "plugin" : {
            "schema": ["djpyplugin_schema_ephys"]
     }
}
```
- Add `setuptools_certificate` as dependency (This is for packaging 'core' and 'plugin' applications to determine 'trustworthiness')
- Add DataJoint 'trusted' development public key (private key copied to LastPass target)
- Expose `schema.py` to plugins
- Expose `attribute_adapter.py` to plugins
- Expose `connection.py` to plugins
- Abstract `host_input` configuration and invoked connection such that they are appropriate for plugin hooks.
- Add `PluginWarning` as cause for `DataJointError`'s thrown when unverified plugins are imported. Otherwise suppress cause.
- Allow `adapted_types` to be loaded from other Python modules (So that they may be developed as plugins)
- Add `cryptography` as dependency (This is necessary for MySQL8 TLS connections)
- Add/Update tests

Fix #715
Fix #729

**Setuptools Certificate Plugin**: source=https://github.com/guzman-raphael/setuptools_certificate , pypi=https://pypi.org/project/setuptools-certificate/
**DataJoint Hub Plugin**: source=https://github.com/guzman-raphael/djpyplugin_connection_hub , pypi=https://pypi.org/project/raphael-connection-hub/


datajoint/autopopulate.py
```diff
@@ -48,7 +48,7 @@ def parent_gen(self):
             try:
                 self._key_source = next(parents)
             except StopIteration:
-                raise DataJointError('A relation must have primary dependencies for auto-populate to work') from None
+                raise DataJointError('A relation must have primary dependencies for auto-populate to work')
```
dimitri-yatsenko:
```suggestion
                raise DataJointError('A table must have primary dependencies for auto-populate to work')
```
This remained from a time when we used theory-based terminology.

Issue #740, opened 2020-03-04
add mechanism for programmatic use of dj.kill


ixcat:
Programmatic use of dj.kill() may be useful in some cases, for example to allow job runners to cleanup old/stuck jobs.

dimitri-yatsenko:
fixed in #741

PR #741, opened 2020-03-04
datajoint/admin.py: add unsafe dj.kill option, expand display to include host

ixcat:
Minimal extension of existing dj.kill() to support #740 -

other more extensive reimplementation might be possible (e.g. making a full 'table' of information_schema.processlist) but would require more development effort.

Unsafe dj.kill() added to facillitate usage by automatic scripts
(e.g. job runners which need to terminate stuck/old jobs forcefully)

Also, add HOST field to display to improve usability in multi-host environments

Changes:

  - add 'safemode' argument to dj.kill

    Value defaults to Null which implies reading the value from current
    dj.config['safemode'] option, falling back to True if this is not present.

  - add safemode=False dj.kill operation mode

    This will apply the given restriction to the processlist query
    and kill all matching processes. Number of killed processes will
    be returned if no errors are encountered.

### DIFF
--- a/datajoint/admin.py
+++ b/datajoint/admin.py
@@ -21,11 +21,13 @@ def set_password(new_password=None, connection=None, update_config=None):   # pr
         config.save_local(verbose=True)
 
 
-def kill(restriction=None, connection=None):  # pragma: no cover
+def kill(restriction=None, connection=None, safemode=None):  # pragma: no cover
     """
     view and kill database connections.
     :param restriction: restriction to be applied to processlist
     :param connection: a datajoint.Connection object. Default calls datajoint.conn()
+    :param safemode: use interactive menu or programmatically terminate processes
+      default: current dj.config safemode value, itself defaulting to True
 
     Restrictions are specified as strings and can involve any of the attributes of
     information_schema.processlist: ID, USER, HOST, DB, COMMAND, TIME, STATE, INFO.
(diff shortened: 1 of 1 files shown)


datajoint/admin.py
```diff
@@ -41,13 +43,23 @@ def kill(restriction=None, connection=None):  # pragma: no cover
     query = 'SELECT * FROM information_schema.processlist WHERE id <> CONNECTION_ID()' + (
         "" if restriction is None else ' AND (%s)' % restriction)
 
+    safemode = config.get('safemode', True) if safemode is None else True
```
dimitri-yatsenko: If `safemode` is not None, then its value is ignored

datajoint/admin.py
```diff
@@ -21,11 +21,13 @@ def set_password(new_password=None, connection=None, update_config=None):   # pr
         config.save_local(verbose=True)
 
 
-def kill(restriction=None, connection=None):  # pragma: no cover
+def kill(restriction=None, connection=None, safemode=None):  # pragma: no cover
     """
     view and kill database connections.
     :param restriction: restriction to be applied to processlist
     :param connection: a datajoint.Connection object. Default calls datajoint.conn()
+    :param safemode: use interactive menu or programmatically terminate processes
```
dimitri-yatsenko: a better name would be `interactive` than `safemode`

datajoint/admin.py
ixcat: `    safemode = config.get('safemode', True) if safemode is None else safemode` ?

datajoint/admin.py
ixcat: agree - though this is consistent with 'delete'... not sure what's best

datajoint/admin.py
ixcat: ^ went ahead with this update

datajoint/admin.py
dimitri-yatsenko: `delete` does not have a `safemode` argument. It has `verbose`. 

I think this can be implemented more intuitively as a separate method `kill_quick` or something.

ixcat: ah yes - got the argument confused with the configuration - either way, we use a term 'safemode' to denote 'prompting' elswhere in the codebase.

separating to 'kill quick' would also 'do' & be consistent elsewhere - will adjust.

ixcat: updated.

Issue #386, opened 2017-10-25
Incorrect transpilation of complex aggregation


dimitri-yatsenko:
The following query 
```python
experiment.Session() & (
        (experiment.Scan() * stimulus.Clip()).aggr(stimulus.Trial(), n='count(*)') 
        & 'n>=2')
```
throws the error
```
---------------------------------------------------------------------------
InternalError                             Traceback (most recent call last)
/usr/local/lib/python3.6/dist-packages/IPython/core/formatters.py in __call__(self, obj)
    691                 type_pprinters=self.type_printers,
    692                 deferred_pprinters=self.deferred_printers)
--> 693             printer.pretty(obj)
    694             printer.flush()
    695             return stream.getvalue()

/usr/local/lib/python3.6/dist-packages/IPython/lib/pretty.py in pretty(self, obj)
    378                             if callable(meth):
    379                                 return meth(obj, self, cycle)
--> 380             return _default_pprint(obj, self, cycle)
    381         finally:
    382             self.end_group()

/usr/local/lib/python3.6/dist-packages/IPython/lib/pretty.py in _default_pprint(obj, p, cycle)
    493     if _safe_getattr(klass, '__repr__', None) is not object.__repr__:
    494         # A user-provided repr. Find newlines and replace them with p.break_()
--> 495         _repr_pprint(obj, p, cycle)
    496         return
    497     p.begin_group(1, '<')

/usr/local/lib/python3.6/dist-packages/IPython/lib/pretty.py in _repr_pprint(obj, p, cycle)
    691     """A pprint that just redirects to the normal repr function."""
    692     # Find newlines and replace them with p.break_()
--> 693     output = repr(obj)
    694     for idx,output_line in enumerate(output.splitlines()):
    695         if idx:

~/dev/datajoint-python/datajoint/relational_operand.py in __repr__(self)
    394 
    395     def __repr__(self):
--> 396         return super().__repr__() if config['loglevel'].lower() == 'debug' else self.preview()
    397 
    398     def preview(self, limit=None, width=None):

~/dev/datajoint-python/datajoint/relational_operand.py in preview(self, limit, width)
    406         if width is None:
    407             width = config['display.width']
--> 408         tuples = rel.fetch(limit=limit+1)
    409         has_more = len(tuples) > limit
    410         tuples = tuples[:limit]

~/dev/datajoint-python/datajoint/fetch.py in __call__(self, *attrs, **kwargs)
    186 
[...]

Issue #558, opened 2019-01-15
Projection removes antijoin constraints


fabiansinz:
I have a collection stimuli that I wanted to extend. The `imagenet.Album` tables have a field called `collection_id`. In the first line, I identify new images that are not yet in the existing set of stimuli. Note that the `stimulus.StaticImage.ImageNet` tables does not have `collection_id`. 

In the second line I create a selection of new stimuli and directly give them a new `collection_id`. However, when I do that, datajoint forgets about the antijoin constraints above. When I remove the `collection_id=3` call the antijoin constraints are applied correctly. 

```
candidates = stimulus.StaticImage.ImageNet()  - imagenet.Album.Single() - imagenet.Album.Oracle()
selection = candidates.proj('image_class', 'image_id', collection_id='3')```

**Comment by eywalker** [comment]
Here is a minimally reproducing case:

```python
schema = dj.schema('test')

@schema
class A(dj.Lookup):
    definition = """
    id: int
    """
    contents = zip(range(10))
    
@schema
class B(dj.Lookup):
    definition = """
    id: int
    id2: int
    """
    contents = zip(range(5), range(5, 10))
```

```python
>>> len(A-B)
5   # correct behavior
>>> len((A - B).proj(id2='3'))
10  # this is a bug!
```

eywalker:
After studying the code, here is an even simpler (and worse) example of failure caused by the same issue:

```python
@schema
class X(dj.Lookup):
    definition = """
    id: int
    """
    contents = zip(range(10))
 
d = dict(id=3, id2=5)
```

Given these,

```python
>>> len(X & d)
1  # as expected
>>> len((X & d).proj(id2='3')
0 # incorrect result!
```

dimitri-yatsenko:
These are two different and somewhat unusual cases.

eywalker:
The underlying cause is the same - projecting in an attribute that actually exists in the restricting dictionary or expression. It does point out logical error in the handling of projection on a restricted expression

dimitri-yatsenko:
Yes, this is clearly an urgent bug but adding more subqueries is not a good solution.  We are working on updating query processing.

eywalker:
Whatever the form will be, we will put some form of a workaround by the end of this week. Hopefully we don’t have to put in the subquery workaround, although it will be temporary anyway.

dimitri-yatsenko:
The fix here is to prohibit creating new attributes with the name of an existing attribute. This will now raise a `DataJointError`

Issue #484, opened 2018-08-20
aggr error


dimitri-yatsenko:
```python
dj.U().aggr(ta3.Segment.aggr(nda2.Trace, n='count(*)'), m='avg(n)')
```

---------------------------------------------------------------------------
InternalError                             Traceback (most recent call last)
/usr/local/lib/python3.6/dist-packages/IPython/core/formatters.py in __call__(self, obj)
    700                 type_pprinters=self.type_printers,
    701                 deferred_pprinters=self.deferred_printers)
--> 702             printer.pretty(obj)
    703             printer.flush()
    704             return stream.getvalue()

/usr/local/lib/python3.6/dist-packages/IPython/lib/pretty.py in pretty(self, obj)
    398                         if cls is not object \
    399                                 and callable(cls.__dict__.get('__repr__')):
--> 400                             return _repr_pprint(obj, self, cycle)
    401 
    402             return _default_pprint(obj, self, cycle)

/usr/local/lib/python3.6/dist-packages/IPython/lib/pretty.py in _repr_pprint(obj, p, cycle)
    693     """A pprint that just redirects to the normal repr function."""
    694     # Find newlines and replace them with p.break_()
--> 695     output = repr(obj)
    696     for idx,output_line in enumerate(output.splitlines()):
    697         if idx:

/usr/local/lib/python3.6/dist-packages/datajoint/relational_operand.py in __repr__(self)
    346 
    347     def __repr__(self):
--> 348         return super().__repr__() if config['loglevel'].lower() == 'debug' else self.preview()
    349 
    350     def preview(self, limit=None, width=None):

/usr/local/lib/python3.6/dist-packages/datajoint/relational_operand.py in preview(self, limit, width)
    371                 for f in columns) for tup in tuples) +
    372             ('\n   ...\n' if has_more else '\n') +
--> 373             (' (%d tuples)\n' % len(rel) if config['display.show_tuple_count'] else ''))
    374 
    375     def _repr_html_(self):

/usr/local/lib/python3.6/dist-packages/datajoint/relational_operand.py in __len__(self)
    474             ' FROM {from_}{where}'.format(
    475                 from_=self.from_clause,
--> 476                 where=self.where_clause)).fetchone()[0]
    477 
    478     def __bool__(self):

/usr/local/lib/python3.6/dist-packages/datajoint/connection.py in query(self, query, args, as_dict, suppress_warnings)
    133                     # suppress all warnings arising from underlying SQL library
[...]

Issue #450, opened 2018-02-20
Slow query due to subquery wrapping


eywalker:
When performing something like the following:

```python
(rel & 'val = x').fetch('KEY')
```

where `val` is an attribute **not** in the primary key, then under the hood, DJ will have to wrap the relation into a subquery because you cannot restrict by a field that is not "selected" (as far I know). However, when the query contains large blob fields, this wrapping with a subquery proves to be extremely slow. Interestingly, it would be much faster if we would simply fetch all non blob attributes and then throw away undesired attributes because there is no subquery wrapping. This is precisely what the `preview` and `repr` does and prove to be an efficient query such that often preview works with no issue but doing `fetch(dj.key)` hangs up.

dimitri-yatsenko:
Related to #349

PR #754, opened 2020-04-11
Aggr fix

dimitri-yatsenko:
Fix issues #386, #449, #450, #484, #558, #761, #785


guzman-raphael:
@dimitri-yatsenko Were you able to patch the bugs we encountered during the demo session we had with the team? The notes I have were:
1. `AssertionError` triggered when performing query `(Enroll - Grade) + (Enroll - Grade)*StudentMajor`
2. Must give a name to every alias in a subquery (perhaps related to above?). This seems to be missed and must be added.
3. `Assertion Error` length bug on query `Course.aggr(Grade*LetterGRade, avg(points))`. Perhaps related to 1.

dimitri-yatsenko:
All issues have been addressed.

datajoint/settings.py
```diff
@@ -45,7 +45,7 @@
     'display.width': 14,
     'display.show_tuple_count': True,
     'database.use_tls': None,
-    'enable_python_native_blobs': False,  # python-native/dj0 encoding support
+    'enable_python_native_blobs': True,  # python-native/dj0 encoding support
```
ixcat, line 48: ^ thinking this needs an adjustment to the README.md (maybe moving the old notes somewhere or removing but leaving some URL for the deleted text)

dimitri-yatsenko: Agreed. Will update before release.

datajoint/expression.py
```diff
@@ -348,11 +161,240 @@ def restrict(self, restriction):
...
+    def __neg__(self):
+        if isinstance(self, Not):
+            return self.restriction
+        return Not(self)
+
+    def __mul__(self, other):
+        """ join of query expressions `self` and `other` """
+        return self.join(other)
+
+    def __matmul__(self, other):
```
guzman-raphael, line 225: This operator presents an issue on the MATLAB side since it is [not an allowed operator](https://www.mathworks.com/help/matlab/matlab_oop/implementing-operators-for-your-class.html) to overload. In the spirit of preserving consistency between clients, we should offer a means which feels similar when introducing a new operator. It doesn't appear as if we have a precedent (intentionally) for this.

dimitri-yatsenko: Each language will necessarily implement DataJoint differently. The Python implementation will use the best available operators. Matlab may find others.

guzman-raphael: Ah yes; missed the implementation of `.join` with appropriate kwargs. That is the *actual* join operation with the `*` and `@` as appropriate aliases for the client language. In that case, we can mark this as resolved here, however, this will be problematic to address in MATLAB in a backward-compatible way... Proper implementation between the two DataJoint clients should have been:
- Join w/ semantic check: `*` for Python, `.*` for MATLAB
- Join w/o semantic check: `@` for Python, `*` for MATLAB

Perhaps the short-term solution would be that once we introduce the operation in MATLAB, we can simply choose not to select an operator alias but have the user structure queries with a specific order in mind.

This just bring further need to bring about `dj.expression(...)` so that we can explicitly apply a standard order to parsing operations so that user experiences a consistent way to generate queries between clients. Will file separate issues on these points.

dimitri-yatsenko:
Closing since all these changes are included in #839

PR #763, opened 2020-04-17
add method update1 to dj.Table

dimitri-yatsenko:
Implement `Table.update1` that works similarly to `insert1`, updating specified secondary attributes.


dimitri-yatsenko:
@guzman-raphael Your review comment has been addressed.

PR #774, opened 2020-04-30
fix when fetch_format='frame'

gucky92:
When setting the fetch_format to 'frame' in the dj.config, using attributes in fetch gives an error.

This was easily fixed by setting format='array', when fetch is called again within "if attrs:"


gucky92:
I have added a test

Issue #666, opened 2019-10-01
pandas insert should use `index=False` to prevent extra `index` field generation


ixcat:

ixcat:
conversation snippet:

```
eywalker:spiral_calendar_pad:  we'd actually want to do to_records() after we reset index, I'd say
index=False simply drops the index field if I recall correctly and that we don't want
so something like rows.reset_index().to_records()
maybe there is also a smart option in to_records() to do this in one step
chris: would reset_index ‘do’ for user-created DataFrames? (or conversely, for DJ fetched ones?)
new messages
eywalker:spiral_calendar_pad:  I'd say so. Most people don't even have any index
and for ones returned by DJ fetch, the DataFrame has the primary keys as multi-index, and this certainly needs to go through reset_index()
```

dimitri-yatsenko:
Will you please describe the scenario in a bit more detail? When is the bothersome extra index field generated?

ixcat:
the below illustrates what this hopes to work but results in a KeyError with 'unknown column index', this should be fixed in some manner which does not break `Dest.insert(Source.fetch(format='frame'))`

```$ cat data_frame_prob.py 
#! /usr/bin/env python

import datajoint as dj
import numpy as np
import pandas as pd

from code import interact

dj.config['enable_python_native_blobs'] = True


schema = dj.schema('test_pd_insert')


@schema
class TestPdInsert(dj.Manual):
    definition = """
    desc:    char(16)
    ---
    data:    longblob
    """



def test_dataframe_index_strip_fails():

    try:
        TestPdInsert.insert(
            pd.DataFrame(data=[('error', np.zeros(10, dtype=np.int64))],
                         columns=['desc', 'data']))
    except KeyError as e:
        if '`index` is not in the table heading' in repr(e):
            print('fails as expected')
            pass


if __name__ == '__main__':
   test_dataframe_index_strip_fails()
   interact('test_dataframe_index_strip_fails', local=locals())
```

ixcat:
todo: reconfirm, report back

ixcat:
confirmed still valid. will adjust per comments above, ensure pandas test coverage for this case

ixcat:
Currently running with:
```$ git diff 98c0a81e2
diff --git a/datajoint/table.py b/datajoint/table.py
index e9bacd4..be79832 100644
--- a/datajoint/table.py
+++ b/datajoint/table.py
@@ -193,7 +193,9 @@ class Table(QueryExpression):
[...]

PR #776, opened 2020-04-30
datajoint/table.py: smarter dataframe conversion (#666)

ixcat:
to address #666

needs further discussion/confirmation with those more familiar with pandas


ixcat:
@dimitri-yatsenko / @eywalker / @guzman-raphael -
Would be good to get some feedback here before un-WIP, my familiarity with pandas is a bit lacking. See also example test script `issue666.txt` in #666 comments.

ixcat:
Thanks for the feedback @gucky92 
Just to make sure I've got it - these both refer to passing the 'drop' argument to reset_index, correct? e.g. you're proposing either:

* `rows.reset_index(drop=not isinstance(rows.index, pandas.MultiIndex)).to_records(index=False)`
or:
* `rows.reset_index(drop=set(self.primary_key) != set(rows.index.names)).to_records(index=False)`

correct?

gucky92:
@ixcat Yes, that is what i meant.

eywalker:
Simpler and more in alignment with what we actually want to check is whether there is any "named" index/indices. Given this, a good check would be 
```
rows.reset_index(drop=len(rows.index.names)==1 and not rows.index.names[0]).to_records(index=False)
```
The default case of an unnamed index will have `rows.index.names` with single entry of `None` and can be safely dropped.

datajoint/table.py
```diff
@@ -193,7 +193,9 @@ def insert(self, rows, replace=False, skip_duplicates=False, ignore_extra_fields
         """
 
         if isinstance(rows, pandas.DataFrame):
-            rows = rows.to_records()
+            rows = rows.reset_index(
```
eywalker, line 196: Suggestion is to check for actual empty named index to drop:
```python
rows.reset_index(drop=len(rows.index.names)==1 and not rows.index.names[0]).to_records(index=False)
```

ixcat:
back to 'WIP' until unit test

eywalker:
Ah sorry I removed WIP as this was set as Draft and figured that suffice to indicate the WIP status.

PR #781, opened 2020-05-11
Fix S3 remove_object and add external location checks

### LINKED ISSUE #686: Filepath: delete_external_files=True does not remove from S3 external store
Record is removed from `schema.external['remote']` but the object in S3 is preserved.

### LINKED ISSUE #748: DataJoint should not create S3 buckets and storage root paths.
Currently, DataJoint automatically creates buckets (in S3) and location paths (in S3 and filesystem) when inserting into external storage if they don't exist. This affects data types `blob@store`, `attach@store`, and `filepath@store`.  This can lead to inadvertent creation of store location due to mispellings, etc.

The behavior should be modified to prevent DataJoint from automatic creation of S3 buckets and storage root paths (location) and raise an error. These paths must be created deliberately outside DataJoint.

guzman-raphael:
- Fix bug with `s3` remove object caller (Fix #686)
- Add explicit `s3` bucket and `file` storage location existence checks (Fix #748)
- Serve `minio` on port 80
- Restructure test buckets to be compatible with DNS-like NGINX rules
- Update tests

Note: `file` store directory check here is a temporary fix until external storage can be properly redesigned to be compatible with `plugin` framework.


datajoint/errors.py
```diff
@@ -81,6 +81,18 @@ class MissingExternalFile(DataJointError):
     """
 
 
+class DirectoryInaccessible(DataJointError):
```
dimitri-yatsenko, line 84: hm... this seems like a non-datajoint-specific error. Should we just raise `FileNotFoundError` instead of creating a datajoint-specific exception?

datajoint/external.py
```diff
@@ -46,6 +46,9 @@ def __init__(self, connection, store=None, database=None):
         if not self.is_declared:
             self.declare()
         self._s3 = None
+        if self.spec['protocol'] == 'file' and not Path(self.spec['location']).is_dir():
+            raise DirectoryInaccessible('Inaccessible local directory %s' %
```
dimitri-yatsenko, line 50:
```suggestion
            raise FileNotFoundError('Inaccessible local directory %s' %
```

Issue #779, opened 2020-05-01
dj.kill order_by extension


ixcat:
https://github.com/datajoint/datajoint-python/issues/668 / https://github.com/datajoint/datajoint-python/pull/775 did not implement list-based order_by as in fetch. Should be adjusted accordingly

dimitri-yatsenko:
closed with #783

PR #775, opened 2020-04-30
datajoint/admin.py: add order_by argument to dj.kill (#668)

ixcat:
see #668

Issue #668, opened 2019-10-02
output of dj.kill has no particular order


ecobost:
For instance, 
```python3
In [22]: dj.kill()
  ID USER         STATE         TIME  INFO
+--+ +----------+ +-----------+ +--+
1241912 ewang                      2261  None
1266899 root                      32096  None
1241936 ewang                      7988  None
1267543 jiakun                     6088  None
1267496 zhiwei                     7736  None
1238807 atlab                         0  None
1238804 atlab                         0  None
1238800 atlab                         0  None
1267558 jiakun                      174  None
1265835 eywalker                  70459  None
1265570 manolis                   54718  None
1267661 ewang                      1434  None
...
```
It would be nice if it was ordered either by id or by user.

ixcat:
normally `dj.kill()` only can see the current user so this is less of an issue - for this installation, process privilege was added to allow users to see all queries. The actual query is in `datajoint/admin.py` and is currently: `select * from information_schema.processlist where id <> connection_id()`. Not sure what or how mysql chooses to sort this internally. Agree this could be sorted. Not sure what would be best. Any feedback on that e.g. what sorts of scenarios is this useful?

To note: the function will take an optional string-style restriction, so e.g. `dj.kill('user like "ecobost"')` or `dj.kill('time > 3600')` may be useful to narrow things down.

ecobost:
`ORDER BY connection_id` should be ok, don't think there's a right answer

dimitri-yatsenko:
Let's just add an `order_By` keyword argument similar to the one in fetch. I see how this can be useful

eywalker:
Fixed by #775

PR #783, opened 2020-05-15
datajoint/admin.py: dj.kill order_by extension 1/N (#779)

ixcat:
adjustment per #779 to address deficiencies in #668 / #775

### DIFF
--- a/datajoint/admin.py
+++ b/datajoint/admin.py
@@ -40,8 +40,7 @@ def kill(restriction=None, connection=None, order_by=None):  # pragma: no cover
         connection = conn()
 
     if order_by is not None and not isinstance(order_by, str):
-        # partial order_by parsing - we don't do key manip, etc. like fetch
-        order_by = ','.join(order_by)
+        order_by = ','.join(order_by)  # c.f. Fetch.fetch() order_by
 
     query = 'SELECT * FROM information_schema.processlist WHERE id <> CONNECTION_ID()' + (
         "" if restriction is None else ' AND (%s)' % restriction) + (
--- a/datajoint/admin.py
+++ b/datajoint/admin.py
@@ -40,7 +40,7 @@ def kill(restriction=None, connection=None, order_by=None):  # pragma: no cover
         connection = conn()
 
     if order_by is not None and not isinstance(order_by, str):
-        order_by = ','.join(order_by)  # cf. Fetch.fetch() order_by
+        order_by = ','.join(order_by)
 
     query = 'SELECT * FROM information_schema.processlist WHERE id <> CONNECTION_ID()' + (
         "" if restriction is None else ' AND (%s)' % restriction) + (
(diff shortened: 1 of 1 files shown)


datajoint/admin.py
```diff
@@ -39,6 +39,10 @@ def kill(restriction=None, connection=None, order_by=None):  # pragma: no cover
     if connection is None:
         connection = conn()
 
+    if order_by is not None and not isinstance(order_by, str):
+        # partial order_by parsing - we don't do key manip, etc. like fetch
```
dimitri-yatsenko, line 43: This comment is unnecessary (no pertinent information). There is no parsing as such. You could say "compare with order_by in fetch" instead

ixcat: it's pertinent in that the order by functions differently and this explains why it was done this way - not sure of a better place to document this kind of trivia and the alternative is omission. if omission is preferred, wil proceed.

ixcat: ok, moved to single line and minimalized

dimitri-yatsenko: What does "partial order_by parsing" mean? What's missing?

dimitri-yatsenko: What's "key manip"?

ixcat: clearly, need to 'cf. Fetch.fetch()` :) 

fetch.fetch does:

```
# expand "KEY" or "KEY DESC"
order_by = list(_flatten_attribute_list(self._expression.primary_key, order_by))
```

which doesn't really apply so much here but technically could

since processlist isn't a datajoint table, there is no primary key, and so using/implementing this logic & associated 'KEY' expansion that happens in _flatten_attribute_list would be superflouous.

however, since direction was to 'make it like fetch', as implemented, this update fails the spec.

dimitri-yatsenko: But the reader of this code does not know this. This logic can be understood simply without knowing anything about fetch.  The comment wording makes it sound like one needs to understand fetch in order to understand this code.

dimitri-yatsenko: This code does not do "key manip", so there is no need to bring it up.

ixcat: fair point, but the reader should also understand 'how order_by works in datajoint', and 'why this code deviates' which comprises 2 mechanisms: dj.kill shorthand and also 'real' fetch.fetch.

anyway, its amended.

ixcat: partially this was a defensive comment against the potential 'but this doesn't actually work like fetch' statement. this discussion replaces the need for the comment to transmit that understanding at least for the current devteam.

dimitri-yatsenko: This `order_by` does not have to do much with `fetch` and it's not helpful to get the reader to understand fetch for understanding `kill`. `kill` does not have anything to do with queries and fetch logic. I would prefer to comment here -- it only distracts.

ixcat: makes sense - which (in retrospect) is why i initially had left out the multiple order by option to begin with. question here was to what extent does it need to be 'like real fetch' since initial partial implementation was deemed 'too partial' in it's omission.

so - should I zap the updated short form comment as well?

dimitri-yatsenko: Thanks. yes, these `admin` utilities are rather general and have little to do with the rest of datajoint, so there is no need explain datajoint.

PR #811, opened 2020-10-09
make dtype=object for adapted attributes. 

dimitri-yatsenko:


guzman-raphael (changes requested):
Since we are modifying the default in tests from `localhost` to `fakeservices.datajoint.io`, would you add a note in `README.md` relating to the tests that for users that run locally, they will need to add an entry in their appropriate `hosts` file mapping their localhost (likely `127.0.0.1`) to `fakeservices.datajoint.io`?

guzman-raphael (approved):

PR #819, opened 2020-10-23
Fix MySQL8 case-sensitivity issue

### LINKED ISSUE #818: MySQL8: Issue with case-sensitivity in dicts returned from information_schema
When setting `as_dict=True`, returned values are `.upper()` for `information_schema` tables in MySQL 8.0.21. Since dicts are case-sensitive, lower-case key will thrown `KeyError`. See:
```python
# MySQL 8.0.21
>>> res = dj.conn().query('SELECT constraint_name FROM information_schema.key_column_usage limit 3', as_dict=True).fetchall()
>>> res
[{'CONSTRAINT_NAME': 'PRIMARY'}, {'CONSTRAINT_NAME': 'PRIMARY'}, {'CONSTRAINT_NAME': 'PRIMARY'}]
>>> res[0]['constraint_name']
  File "<stdin>", line 1, in <module>
KeyError: 'constraint_name'
```
Solution could be to `.lower()` all returned keys after query specifically from `information_schema.*`.

Also, this is a good note on the warnings thrown from `MySQL8` in docker i.e. `mbind: Operation not permitted`
https://github.com/docker-library/mysql/issues/303

guzman-raphael:
Fix #818

### DIFF
--- a/datajoint/dependencies.py
+++ b/datajoint/dependencies.py
@@ -43,7 +43,7 @@ class Dependencies(nx.DiGraph):
             self.add_node(n, primary_key=pk)
 
         # load foreign keys
-        keys = self._conn.query("""
+        keys = [{k.lower(): v for k, v in elem.items()} for elem in self._conn.query("""
         SELECT constraint_name,
             concat('`', table_schema, '`.`', table_name, '`') as referencing_table,
             concat('`', referenced_table_schema, '`.`',  referenced_table_name, '`') as referenced_table,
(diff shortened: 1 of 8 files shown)


datajoint/dependencies.py
```diff
@@ -43,15 +43,15 @@ def load(self):
             self.add_node(n, primary_key=pk)
 
         # load foreign keys
-        keys = self._conn.query("""
+        keys = [{k.lower(): v for k, v in elem.items()} for elem in self._conn.query("""
```
dimitri-yatsenko, line 46: change list to generator expression for lazy evaluation.

datajoint/external.py
```diff
@@ -276,11 +276,11 @@ def references(self):
         """
         :return: generator of referencing table names and their referencing columns
         """
-        return self.connection.query("""
+        return [{k.lower(): v for k, v in elem.items()} for elem in self.connection.query("""
```
dimitri-yatsenko, line 279: replace list comprehension with a generator expression just as the docstring specifies.

dimitri-yatsenko (changes requested):

datajoint/external.py
guzman-raphael: Good point, I have updated it to be a proper generator

datajoint/dependencies.py
guzman-raphael: Good point, I have updated it to be a proper generator

guzman-raphael:
@ixcat I have updated `datajoint.admin` to now expect `.lower()` keys returned to match the rest of the codebase from `information_schema.*` as we discussed in our Dev meeting earlier today.

ixcat (approved):

dimitri-yatsenko (approved):

PR #823, opened 2020-10-26
Sync master into Release013

dimitri-yatsenko:


guzman-raphael (changes requested):
Assuming we will be following this with a `0.12.7` release, would you update `CHANGELOG.md` and `docs-parts/intro/Releases_lang1.rst` to reflect bugfixes and new features?

guzman-raphael (approved):

PR #826, opened 2020-10-27
Master -> Pre

guzman-raphael:
Note: `pre` is our stable branch where users may safely pull from. It should match with the latest release (or pre-release) in [PyPi](https://pypi.org/project/datajoint/).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -6,13 +6,13 @@ import sys
 min_py_version = (3, 5)
 
 if sys.version_info <  min_py_version:
-    sys.exit('DataJoint is only supported on Python {}.{} or higher'.format(*min_py_version))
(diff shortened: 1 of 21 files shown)


dimitri-yatsenko (approved):

dimitri-yatsenko (approved):

setup.py
```diff
@@ -6,13 +6,13 @@
 min_py_version = (3, 5)
 
 if sys.version_info <  min_py_version:
-    sys.exit('DataJoint is only supported on Python {}.{} or higher'.format(*min_py_version))
+    sys.exit('DataJoint is only supported for Python {}.{} or higher'.format(*min_py_version))
```
ixcat, line 9: this is actually less idiomatic usage.. don't want to go all grammar crazy with a paragraph about this though, both make sense

PR #836, opened 2020-12-03
Query cache

dimitri-yatsenko:
Cache queries when when `conn.set_cache('name')` is set.


dimitri-yatsenko:
closing because changes were merged into #839

Issue #838, opened 2020-12-04
Request for a `schema.list_tables()`


ttngu207:
Similar to `dj.list_schemas()`, would be great if we can list all tables under a particular schema that are topologically sorted - with `schema.list_tables()`

Synicix:
#844 
I added basic table fetcher for now since that is what I needed for DJGUI.

I am not entirely sure what you mean by topologically sorted, as in a dict with the table names in hierarchy of which one depends on each other? Either way I probably work on that later unless it is urgently needed.

dimitri-yatsenko:
Fixed with #844

guzman-raphael:
Hey @dimitri-yatsenko. We are keeping this open as it is currently not applying a topological sort.

dimitri-yatsenko:
I see. #833 provides `connection.dependencies.topological_sort()` to provide this functionality.

PR #844, opened 2020-12-09
Added basic list_tables function for schema class

Synicix:
I just added what I needed for DJGUI for now, as such it returns the raw names to be parsed by the API on the other side.

I can add additional features as people request them like restricting by lookup tables are returning it in a different format.

### DIFF
--- a/datajoint/schemas.py
+++ b/datajoint/schemas.py
@@ -290,7 +290,8 @@ class Schema:
 
     def list_tables(self):
         """
-        Return a list of all tables in the schema except tables with ~ in first character such as ~logs
+        Return a list of all tables in the schema except tables with ~ in first character such
+        as ~logs and ~job
         :return: A list of table names in their raw datajoint naming convection form
         """
 
--- a/datajoint/schemas.py
+++ b/datajoint/schemas.py
@@ -294,8 +294,9 @@ class Schema:
         :return: A list of table names in their raw datajoint naming convection form
         """
 
-        query_result = self.connection.query('SELECT table_name FROM information_schema.tables WHERE table_schema = \'' + self.database + '\'').fetchall()
-        return [table_name_tuple[0] for table_name_tuple in query_result if '~' != table_name_tuple[0][0]]
+        return [table_name for (table_name,) in self.connection.query("""
+            SELECT table_name FROM information_schema.tables
+            WHERE table_schema = %s and table_name NOT LIKE '~%%'""", args=(self.database))]
         
 
 class VirtualModule(types.ModuleType):
--- a/datajoint/schemas.py
+++ b/datajoint/schemas.py
@@ -298,7 +298,7 @@ class Schema:
         return [table_name for (table_name,) in self.connection.query("""
             SELECT table_name FROM information_schema.tables
             WHERE table_schema = %s and table_name NOT LIKE '~%%'""", args=(self.database))]
-        
+
 
 class VirtualModule(types.ModuleType):
     """
--- a/datajoint/schemas.py
+++ b/datajoint/schemas.py
@@ -288,6 +288,15 @@ class Schema:
             with open(python_filename, 'wt') as f:
                 f.write(python_code)
 
+    def list_tables(self):
+        """
+        Return a list of all tables in the schema except tables with ~ in first character such as ~logs
+        :return: A list of table names in their raw datajoint naming convection form
+        """
+
+        query_result = self.connection.query('SELECT table_name FROM information_schema.tables WHERE table_schema = \'' + self.database + '\'').fetchall()
+        return [table_name_tuple[0] for table_name_tuple in query_result if '~' != table_name_tuple[0][0]]
+        
 
 class VirtualModule(types.ModuleType):
     """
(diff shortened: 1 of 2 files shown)


guzman-raphael (changes requested):
Thanks @Synicix for the PR. In addition to the provided inline feedback, I would suggest to add at least one test to validate the feature. Otherwise, our coverage will go down. It could simply be to create a table and then verify that it is returned when invoking the class method.

Ugh, looks like there was a recent release (yesterday!) of `minio` python package that broke some S3 features. Such is the nature of development... Could you also track down the nature of the failure and see if it is easy to patch? You should be able to inspect the tests in GitHub Actions or locally to get a good idea of what is happening. Worst case scenario, we can version pin it to the last release. You can also file an issue regarding it.

datajoint/schemas.py
```diff
@@ -288,6 +288,15 @@ def repl(s):
             with open(python_filename, 'wt') as f:
                 f.write(python_code)
 
+    def list_tables(self):
+        """
+        Return a list of all tables in the schema except tables with ~ in first character such as ~logs
+        :return: A list of table names in their raw datajoint naming convection form
+        """
+
+        query_result = self.connection.query('SELECT table_name FROM information_schema.tables WHERE table_schema = \'' + self.database + '\'').fetchall()
```
dimitri-yatsenko, line 297: Take a look at the SHOW TABLES query above and refactor the code to re-use list_tables there as well.

dimitri-yatsenko (changes requested):

datajoint/schemas.py
```diff
@@ -288,6 +288,15 @@ def repl(s):
             with open(python_filename, 'wt') as f:
                 f.write(python_code)
 
+    def list_tables(self):
+        """
+        Return a list of all tables in the schema except tables with ~ in first character such as ~logs
```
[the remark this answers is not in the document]
Synicix: Man people must be really using small screen or low resolution...

datajoint/schemas.py
Synicix: I will not do this here under this pull request. This should be seperate thing.

Synicix:
Need to add tests, please wait for that commmit before reviewing

Synicix:
Will look into the minio errors in a different pull request

Synicix:
#847 In progress, probably will fix that first then have this update with that test fix to pass the requirements.

tests/test_schema.py
```diff
@@ -105,6 +106,9 @@ class Unit(dj.Part):
 
     test_schema.drop()
 
+def test_list_tables():
+    assert(['#a', '#argmax_test', '#data_a', '#data_b', '#i_j', '#j_i', '#l', '#t_test_update', '__b', '__b__c', '__d', '__e', '__e__f', 'f', 'reserved_word'] == schema_simple.list_tables())
```
guzman-raphael, line 110: Can you also break this up based on line character limit of 95? In VSCode, I usually use a ruler to remind me throughout the code. As a tip, here is some of the values in my `settings.json` that I like using.
```json
{
    "python.linting.pycodestyleEnabled": true,
    "python.linting.pylintEnabled": true,
    "python.linting.enabled": true,
    "editor.rulers": [
        95
    ],
    "python.linting.pycodestyleArgs": ["--max-line-length=95"],
}
```

guzman-raphael (changes requested):
See inline comment.

datajoint/schemas.py
```diff
@@ -288,6 +288,17 @@ def repl(s):
...
 
+    def list_tables(self):
+        """
+        Return a list of all tables in the schema except tables with ~ in first character such
+        as ~logs and ~job
+        :return: A list of table names in their raw datajoint naming convection form
+        """
+
+        return [table_name for (table_name,) in self.connection.query("""
+            SELECT table_name FROM information_schema.tables
```
dimitri-yatsenko, line 299: While selecting from the `information_schema` is a MySQL-specific command, the more general and simpler SQL command for getting the table list is `SHOW TABLES`. Unless absolutely necessary, please use the more general SQL.

dimitri-yatsenko (changes requested):

datajoint/schemas.py
guzman-raphael: @dimitri-yatsenko Wait... we should be careful here. Is it not the other way around? When using `SHOW TABLES` on postgres the following is the result:
```
postgres=# SHOW TABLES;
ERROR:  unrecognized configuration parameter "tables"
```
As opposed to how it is has been implemented here which properly returns a list of tables.

dimitri-yatsenko: Good point. Then we need to correct it in the other places. I had assumed that `information_schema` was MySQL specific. We have other places in the code using "SHOW TABLES".  Since we are adding this as the verified way to do it, let's also refactor the other places to call `list_tables` instead of issuing their own query.

guzman-raphael: @dimitri-yatsenko I agree that we should make those updates, though I think it best to treat as a separate issue/PR so we don't block adding in this simple feature for use in DJ GUI. Method signature won't change except perhaps in adding more options/args which can easily be done in a backward-compatible way.

Synicix:
#838   Partially implmented for now

guzman-raphael (approved):

dimitri-yatsenko (approved):

dimitri-yatsenko:
fix #838

PR #847, opened 2020-12-10
Fixing Minio newest package version incompability

Synicix:
For now I changed to except all errors to make it work. I need to dig more into how minio does error handling to shrink the error scope to be more specific.

### DIFF
--- a/datajoint/s3.py
+++ b/datajoint/s3.py
@@ -38,6 +38,7 @@ class Folder:
         try:
             return self.client.get_object(self.bucket, str(name)).data
         except Exception as e:
+            print(e)
             raise errors.MissingExternalFile('Missing s3 key %s' % name) from None
 
     def fget(self, name, local_filepath):
@@ -69,6 +70,7 @@ class Folder:
         try:
             return self.client.stat_object(self.bucket, str(name)).size
         except Exception as e:
+            print(e)
             raise errors.MissingExternalFile from None
 
     def remove_object(self, name):
--- a/datajoint/s3.py
+++ b/datajoint/s3.py
@@ -78,7 +78,7 @@ class Folder:
                 raise errors.MissingExternalFile from None
             else:
                 raise e
-            
+  
     def remove_object(self, name):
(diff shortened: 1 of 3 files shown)


guzman-raphael:
@Synicix Looks like from [here](https://github.com/minio/minio-py/blob/beb3d531bee8ed3b778ee673ce296f1f4e5bc2dd/minio/error.py#L67) you can check `minio.error.S3Error._code`. It appears that the code translation is done based on [this](https://github.com/minio/minio-py/blob/beb3d531bee8ed3b778ee673ce296f1f4e5bc2dd/minio/api.py#L329-L360) mapping.

datajoint/s3.py
```diff
@@ -59,17 +62,23 @@ def exists(self, name):
...
             return self.client.stat_object(self.bucket, str(name)).size
-        except minio.error.NoSuchKey:
-            raise errors.MissingExternalFile from None
-
+        except minio.error.S3Error as e:
+            if e.code == 'NoSuchKey':
+                raise errors.MissingExternalFile from None
+            else:
+                raise e
+
```
guzman-raphael, line 81: Just took a quick look and looks like there is a simple whitespace issue.
```suggestion

```
To run the syntax tests locally, see [here](https://github.com/datajoint/datajoint-python/blob/143ba7b55fd292a0737e72d058670bcb5c6db064/local-docker-compose.yml#L89). To run the style tests locally, see [here](https://github.com/datajoint/datajoint-python/blob/143ba7b55fd292a0737e72d058670bcb5c6db064/local-docker-compose.yml#L90). We should probably add a comment/note in `local-docker-compose.yml` to be more clear about this.

guzman-raphael (changes requested):

datajoint/s3.py
Synicix: Sigh I thought I removed it all....

dimitri-yatsenko (approved):

guzman-raphael (approved):
Looks good! :rocket:

PR #850, opened 2020-12-24
Pre-release 0.13dev2 candidate

dimitri-yatsenko:
Pre-release 0.13dev2 candidate


guzman-raphael (changes requested):
@dimitri-yatsenko Can you update the version to `0.13.dev2` so that I can issue a new pre-release after merging?

guzman-raphael (approved):

PR #859, opened 2021-01-21
0.13 enable_python_native_blobs -> True (#761)

ixcat:
fix #761

details / points:
- [X] default `enable_python_native_blobs changed` to `True`
- [X] readme updated (slight question if this should be removed/moved to docs - my thinking is after #765)
- [X] unit tests adjusted where applicable (slight judgement call)
- [X] upgrading to =True upgrade on inserting to jobs, etc. per #700 not adjusted

PR #861, opened 2021-01-21
minimum python -> 3.6 (#829)

### LINKED ISSUE #829: Remove Python 3.5 support
As it has reached [End-of-Life](https://www.python.org/dev/peps/pep-0478/#release-schedule), we can safely remove support and update docs.

Noting hear as there were some challenges that Python 3.5 presented specifically when building lightweight conda-based images for development, testing, and deployments.

ixcat:
fix #829 

Bump minimum required python from 3.5 to 3.6, remove 3.5 / add 3.9 to
unit test actions.

Currently, this is a configuration-only change, removal of OrderedDict,
etc. or other >3.5.x feature migration not included.


ixcat:
@guzman-raphael - since this touched the GH actions somewhat blindly, would be good to get your signoff / adjustments.

guzman-raphael:
@ixcat The issue here is b/c we have not yet added proper support for python3.9 in our images and also have not migrated the images used in tests here to utilize the new, optimized image hierarchy. The right thing to do would be to pull from https://github.com/datajoint/djtest-docker and add `3.9` to its build/push process.

ixcat:
makes sense - is that your preferred course of action, or to back out the 3.9 additions here and just leave the 3.5 removals?

guzman-raphael:
@ixcat I think for now, simply removing Python3.5 is fine here. We do need to add Python3.9 to the process but it comes with some time for development to propagate, verify things. In any case, it should be low risk as I am not aware of any feature we are leveraging that is currently throwing a deprecation warning.

ixcat:
think this is gtg @guzman-raphael

CHANGELOG.md
```diff
@@ -1,8 +1,11 @@
 ## Release notes
 
+### Current
+=======
```
guzman-raphael, line 4:
```suggestion
```
Not sure if you intentionally meant for this to go in but the `0.13.0 -- TBD` block is serving as the 'current' placeholder.

guzman-raphael (changes requested):
Thanks for the update, @ixcat. We'll also need to coordinate a new PR to our [conda forge feedstock](https://github.com/conda-forge/datajoint-feedstock) as well for this to be fully applied across distributions. We can handle that change though once we are closer to `0.13.0` release though.

CHANGELOG.md
ixcat: yes, this is a side-effect from 'develop against master', wasn't sure of the best approach. 

updated/adjusted/conflict resolved. would be good to figure out the best approach to auto-prevent conflicts here if possible, though these are quick to resolve anyway.

guzman-raphael (changes requested):
@ixcat Since I merged #859, this now also has a conflict. Could you resolve this? This is expected when you modify the same file across different PR's (as why it is only the `CHANGELOG.md`).

Speaking of which, please go ahead and update `docs-parts/intro/Releases_lang1.rst` document as well.

ixcat:
@guzman-raphael resolved merge conflicts

ixcat:
sorry, @guzman-raphael docs-parts/intro/Releases_lang1.rst also updated now.

guzman-raphael (approved):

PR #862, opened 2021-01-22
add docs-parts/computation/06-distributed-computing_jobs_by_key.rst  (#804)

### LINKED ISSUE #804: python-specific documentation for schema.jobs + hash of keys
Python component for: https://github.com/datajoint/datajoint-docs/issues/246

ixcat:
fix #804

also, set other docs-parts in this directory to use `python` code block instead of `text`


docs-parts/computation/06-distributed-computing_jobs_by_key.rst
```diff
@@ -0,0 +1,21 @@
+
+This can be done by using `dj.hash.key_hash` to convert the key as follows:
+
+.. code-block:: python
+
+    In [4]: schema.jobs & {'key_hash' : dj.hash.key_hash({'id': 2})}
```
dimitri-yatsenko, line 6: Ok. So we are now making `key_hash` part of the external API. Perhaps it can be brought to the package level `dj.key_hash`

dimitri-yatsenko (changes requested):

docs-parts/computation/06-distributed-computing_jobs_by_key.rst
dimitri-yatsenko: How common is this operation? Why would someone need to look for a job by the primary key of the `make` call?

ixcat: .. this crossed my mind. should i include this in the PR? possibly a good utility anyway for 'compressing' downstream keys to save key attributes/size, etc

dimitri-yatsenko: You are right. Yes, let's make this a top-level function `dj.key_hash`. It can be used for many things as you said @ixcat.

ixcat: @dimitri-yatsenko - ok sounds good - thoughts on a) full move to top-level, b) just importing a reference, or c) moving to top-level and keeping a (possibly temporary) refrence in dj.hash? Leaning towards 'b' to keep the toplevel init cleaner / hash related code in one place.

ixcat: implemented as 'b' (include reference at toplevel). adjustments welcome.

r.e. common - when there are populate jobs running against multiple tables, it can be tedious to manually navigate through the bulk error list to find the related exception for a given key; also, often the jobs investigation process is based on finding out 'what happened' to a missing result/key - e.g. users will see a missing key & wonder what happened in populate - this shortcuts needing to fetch jobs and loop through results in an ad-hoc loop, cross check/estimate against job run timestamps, source machine etc - can just say `schema.jobs & dj.key_hash(missing_key)`

dimitri-yatsenko (approved):

docs-parts/computation/06-distributed-computing_jobs_by_key.rst
guzman-raphael, line 6: Makes sense. Based on the discussion, let's update this doc to:
```suggestion
    In [4]: schema.jobs & {'key_hash' : dj.key_hash({'id': 2})}                
```

guzman-raphael (approved):
@ixcat Looks pretty good. Thanks!

Only adding a suggestion that perhaps our docs should reflect `dj.key_hash` based on discussion.

guzman-raphael (changes requested):
@ixcat Oh whoops, need the merge conflicts resolved actually...

docs-parts/computation/06-distributed-computing_jobs_by_key.rst
ixcat: thanks, marking feedback as resolved by 25ae98f

ixcat: treating as resolved per last comment / 3945725

ixcat:
think we are good to go here @guzman-raphael

guzman-raphael (approved):

PR #871, opened 2021-02-12
Replace tinyint(1) with bool in describe

guzman-raphael:


dimitri-yatsenko:
As we discussed off-line, `tinyint(1)` cannot be used as the alias for `bool`.  We approach this by explicitly saving the declared type.

guzman-raphael:
Pausing this for now and will revisit soon.

PR #872, opened 2021-02-13
Cleanup `schema.activate` logic

dimitri-yatsenko:
Fixed an issue when schemas were created with `create_schema=False`

### DIFF
--- a/datajoint/schemas.py
+++ b/datajoint/schemas.py
@@ -299,7 +299,7 @@ class Schema:
         if self.database is None:
             raise DataJointError("Schema must be activated first.")
         return self.database is not None and (
-            self.connection.query("SHOW DATABASES LIKE '{database}'".format(
+            self.connection.query("SELECT schema_name FROM information_schema.schemata WHERE schema_name LIKE '{database}'".format(
                 database=self.database)).rowcount > 0)
 
     @property
--- a/datajoint/schemas.py
+++ b/datajoint/schemas.py
@@ -406,4 +406,7 @@ def list_schemas(connection=None):
     :param connection: a dj.Connection object
     :return: list of all accessible schemas on the server
     """
-    return [r[0] for r in (connection or conn()).query('SHOW SCHEMAS') if r[0] not in {'information_schema'}]
+    return [r[0] for r in (connection or conn()).query(
+        'SELECT schema_name '
+        'FROM information_schema.schemata '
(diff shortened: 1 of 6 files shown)


datajoint/schemas.py
```diff
@@ -273,43 +276,46 @@ def drop(self, force=False):
...
         """
         :return: true if the associated schema exists on the server
         """
-        self._assert_activation()
-        cur = self.connection.query("SHOW DATABASES LIKE '{database}'".format(database=self.database))
-        return cur.rowcount > 0
+        if self.database is None:
+            raise DataJointError("Schema must be activated first.")
+        return self.database is not None and (
+            self.connection.query("SHOW DATABASES LIKE '{database}'".format(
```
guzman-raphael, line 302: We should avoid querying using syntax unique to MySQL.

```suggestion
            self.connection.query("SELECT schema_name FROM information_schema.schemata WHERE schema_name LIKE '{database}'".format(
```

guzman-raphael (changes requested):
Looks good. Only have a minor update regarding MySQL-specific language.

datajoint/schemas.py
dimitri-yatsenko: perhaps here instead of `LIKE` we should just use `=`

guzman-raphael: Yea wasn't sure if I missed something on first pass but looked like there were no wildcard operators used so should be good.

datajoint/schemas.py
```diff
@@ -273,43 +276,46 @@ def drop(self, force=False):
...
         """
         :return: true if the associated schema exists on the server
         """
-        self._assert_activation()
-        cur = self.connection.query("SHOW DATABASES LIKE '{database}'".format(database=self.database))
-        return cur.rowcount > 0
+        if self.database is None:
+            raise DataJointError("Schema must be activated first.")
+        return self.database is not None and (
+            self.connection.query("SELECT schema_name FROM information_schema.schemata WHERE schema_name LIKE '{database}'".format(
```
guzman-raphael, line 302:
```suggestion
            self.connection.query(
                "SELECT schema_name "
                "FROM information_schema.schemata "
                "WHERE schema_name = '{database}'".format(
```

guzman-raphael (changes requested):
I've updated the suggestion. If you commit it we should be ready to merge.

guzman-raphael (approved):

Issue #876, opened 2021-03-05
Index key error when using fetch1('KEY')


shenshan:
### Description
When using fetch1('KEY') in datajoint 0.12.8, it gives this error when using fetch1('KEY'):
![image](https://user-images.githubusercontent.com/2539832/110170128-91dea080-7dae-11eb-9052-3a27f802b212.png)

dimitri-yatsenko:
This occurred when `dj.config['fetch_format']` was set to "`frame`" to return Pandas dataframes.  The `fetch_format` applies only to results of `fetch` not `fetch1`.  But `fetch1` calls `fetch` and did not handle the dataframe results correctly. The fix was to set `format="array"` when calling `fetch` from `fetch1`.  Fixed in #857.

Issue #857, opened 2021-01-15
0.13.dev2 duplicate-attribute join issue


ixcat:
### Description

Testing 0.13.dev2 query engine yielded an unexpected exception when restricting on tables with duplicate fields

### Reproducibility
```
>>> print(lab.Subject().definition)
   subject_id          : int   # institution 6 digit animal ID
   ---
   -> [nullable] Person        # person responsible for the animal
   cage_number         : int   # institution 6 digit animal ID
   date_of_birth       : date  # format: yyyy-mm-dd
   sex                 : enum('M','F','Unknown')
   -> [nullable] AnimalSource  # where was the animal ordered from
>>> print(lab.WaterRestriction().definition)
   -> Subject
   ---
   water_restriction_number    : varchar(16)   # WR number
   cage_number                 : int
   wr_start_date               : date
   wr_start_weight             : Decimal(6,3)
>>> lab.WaterRestriction() * lab.Subject().proj()
...
pymysql.err.OperationalError: (1052, "Column 'cage_number' in field
 list is ambiguous")

```
### Expected Behavior

This previously allowed joining on the projected lab.Subject

PR #878, opened 2021-03-11
Fix join error in the new query parser (#857)

dimitri-yatsenko:
Fix issue #857. 
Fix issue #876.

### DIFF
--- a/tests/test_fetch.py
+++ b/tests/test_fetch.py
@@ -241,5 +241,9 @@ class TestFetch:
         list1 = sorted(key, key=itemgetter('subject_id'))
         for l1, l2 in zip(list1, list2):
             assert_dict_equal(l1, l2, 'Primary key is not returned correctly')
+
+        # test pandas with fetch1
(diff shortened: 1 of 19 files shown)


tests/test_fetch.py
```diff
@@ -222,24 +222,33 @@ def test_nullable_numbers(self):
...
+    def test_key_fetch1(self):
+        """test KEY fetch1 - issue #976"""
+        with dj.config(fetch_format="array"):
+            k1 = (self.subject & 'subject_id=10').fetch1('KEY')
+        with dj.config(fetch_format="frame"):
+            k2 = (self.subject & 'subject_id=10').fetch1('KEY')
+        assert_equal(k1, k2)
+
+    def test_same_secondary_attribute(self):
+        print(schema.Child * schema.Parent().proj())
```
guzman-raphael, line 254: We need a better check here than just a simple print. Perhaps something that confirms the correct value is fetched.

guzman-raphael (approved):

Issue #895, opened 2021-03-25
explicit calls to dj.connection.Connection fail without providing 'host_input'


ixcat:
### Description

explicit calls to dj.connection.Connection fail without providing 'host_input'
(seems to be a 0.13 regression)

### Expected Behavior

call dj.connection.Connection without host_input argument

### Additional Research and Context

dj.connection.Connection sets default host_input to None; this value is then passed to get_host_hook
where it is tested in the statement: `if '://' in connection_obj.conn_info['host_input']:` which raises: `TypeError: argument of type 'NoneType' is not iterable`

side note: seems like some duplicate & inconsistent logic in processing dj.config values between Connection, conn and get_host_hook; might make sense to refactor get_host_hook to simply take a raw host string which returns all needed parts (casually, among the 3: host, port). This would also remove the need for host_input in Connection.

also probably means that explicit dj.Connection tests need to be added

guzman-raphael:
@ixcat The logic behind `conn` and `Connection` is quite old and clunky. For this, it is not advisable to call `Connection` directly. There is specific pre-processing in `conn` that is bypassed by calling `Connection` directly. We should optimize this to be more streamlined in the future but for now, I would say that we should discourage use of `dj.Connection` directly and perhaps not make it available by default. If you have a specific use-case on the need for using `dj.Connection` over `dj.conn` I would like to understand it a bit more.

ixcat:
We use multiple connections in several places, so relying on dj.conn() is a no-go.
Above refactor description addresses all concerns & supports both methods AFAICT

ixcat:
Didn't consider plugin usage in refactor proposal; that said I think the overall approach should be applicable & would simplify logic overall (haven't xchecked). A stopgap would be simply defaulting host_input in Connection to the empty string or adjusting the '://' test to support NoneType input (former seems preferable; Connection seems to be the only consumer passing None to the hook function so no need to layer hacks on hacks to support Connection)

saumilpatel:
Yes, we have to be able to make multiple connections, need to fix this asap and also provide a mechanism to install older version via pip.

PR #906, opened 2021-04-20
datajoint/connection.py: default host_input to '' in Connection as stopgap for #895

ixcat:
fix (temporary) for #895 

still think this can be refactored to remove need for host_input (or around making full host_input string the default argument);
for now, defaulting host_input to '' will allow the test on line 25 in get_host_hook to work when host_input is not provided to Connection. Didn't change arg default in method signature to keep the current API the same until some later refactor though that could be reasonable/better as a stopgap.

### DIFF
--- a/datajoint/connection.py
+++ b/datajoint/connection.py
@@ -172,7 +172,7 @@ class Connection:
         if use_tls is not False:
             self.conn_info['ssl'] = use_tls if isinstance(use_tls, dict) else {'ssl': {}}
         self.conn_info['ssl_input'] = use_tls
-        self.conn_info['host_input'] = host_input
+        self.conn_info['host_input'] = host_input if host_input else ''
         self.init_fun = init_fun
         print("Connecting {user}@{host}:{port}".format(**self.conn_info))
         self._conn = None


datajoint/connection.py
```diff
@@ -172,7 +172,7 @@ def __init__(self, host, user, password, port=None, init_fun=None, use_tls=None,
         if use_tls is not False:
             self.conn_info['ssl'] = use_tls if isinstance(use_tls, dict) else {'ssl': {}}
         self.conn_info['ssl_input'] = use_tls
-        self.conn_info['host_input'] = host_input
+        self.conn_info['host_input'] = host_input if host_input else ''
```
guzman-raphael, line 175: Yes, this is very much a hack that won't work for connection plugin users if they call `Connection` explicitly. Looks like `get_host_hook` was simply misplaced in `conn` rather than `Connection` in the original implementation. Should be pretty straightforward to patch. Could you:
- remove `dj.Connection` argument `host_input`
- move `get_host_hook` be called first in `dj.Connection`

guzman-raphael (changes requested):
Thanks for the PR @ixcat. Could you also update the changelogs?

datajoint/connection.py
ixcat: a) not sure how get_host_hook would be called in proposed refactor - simply with empty string?

b) assuming 'a' as yes, the `connect_host_hook` call later depends on conn_info containing 'host_input' item, and it also tries to 'iterate' on this string (if '://' in ...), so unless I'm misunderstanding things, this would just move the problem elsewhere

feel free to counter-PR if that's easier, mainly breaking my 'you break it, you buy it' rule by putting 'something' in to get the ball rolling for just the regression aspect.

guzman-raphael: a. Basically, in the first line of the `dj.Connection` constructor, you would have:
   ```python
   host_input, host = (host, get_host_hook(host_input))
   ```
b. This would not be applicable as `host_input` would always be satisfied now provided `host` is a string (as assumption that has already been there for `dj.Connection`).

Could you also add a test for this too?

guzman-raphael: a. Basically, in the first line of the `dj.Connection` constructor, you would have:
   ```python
   host_input, host = (host, get_host_hook(host_input))
   ```
b. This would not be applicable as `host_input` would always be satisfied now provided `host` is a string (as assumption that has already been there for `dj.Connection`).

Could you also add a test for this too?

guzman-raphael: Let me know if this makes sense or if you prefer I open another one.

ixcat: a) the proposed call to get_host_hook(host_input) will now have no input variable - where does the value for 'host_input' come from if in the proposed call if it is removed from __init__ signature? or do you mean `host_input, host = (host, host)` ? 
b) still confused due to 'a')
c) r.e. tests: this is a hack fix to a regression that I didn't introduce, so don't think this is in-scope here. You break, you buy (tm).

guzman-raphael: Oh sorry, that was a typo. Fixed it above.
OK, I'll open a PR then.

ixcat:
superseded by https://github.com/datajoint/datajoint-python/pull/907

PR #907, opened 2021-04-20
Fix issue with explicit calls to `dj.Connection`

### LINKED ISSUE #895: explicit calls to dj.connection.Connection fail without providing 'host_input'

guzman-raphael:
Fix #895

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## Release notes
 
+### 0.13.2 -- TBD
+* Bugfix - Explicit calls to `dj.Connection` throw error due to missing `host_input` (#895) PR #907
+
 ### 0.13.1 -- Apr 16, 2021
 * Add `None` as an alias for `IS NULL` comparison in `dict` restrictions (#824) PR #893
 * Drop support for MySQL 5.6 since it has reached EOL PR #893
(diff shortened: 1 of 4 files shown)


ixcat (approved):

ixcat:
Thanks @guzman-raphael - now that I see it in person, makes much more sense :) also, test was not that bad, thought you meant more tests around the *hook functions

guzman-raphael:
@ixcat No worries man. Always a challenge to translate some thoughts clearly in a remote capacity. :sweat_smile:

dimitri-yatsenko (approved):

PR #910, opened 2021-04-23
Rename `setuptools_certificate` dependency to `otumat`

guzman-raphael:
This rename was necessary and allows us to achieve the following:
- Prepares the `otumat` dependency to be `conda-forge` compatible.
- Prepares the `otumat` dependency to be more generic so that other utilities can be introduced (such as usage tracking)

Will follow this with a new release so that we can catch up our conda distro properly. It is currently at `0.12.9` due to several issues.


guzman-raphael:
@dimitri-yatsenko This PR is ready for final review.

PR #913, opened 2021-04-28
Add usage tracking dependency

guzman-raphael:


jverswijver:
Decided to move in a different direction, discussed this with @guzman-raphael

PR #939, opened 2021-08-05
Replace use of numpy aliases of built-in types with built-in type

### LINKED ISSUE #938: Remove deprecated use of numpy aliases of built-in types
Using datajoint 0.13.2 and numpy 1.20.3 or the latest 1.21.1, I get the following deprecation warning repeatedly:
```
/home/rly/miniconda3/envs/nwb_datajoint/lib/python3.8/site-packages/datajoint/table.py:618: DeprecationWarning: `np.float` is a deprecated alias for the builtin `float`. To silence this warning, use `float` by itself. Doing this will not modify any behavior and is safe. If you specifically wanted the numpy scalar type, use `np.float64` here.
Deprecated in NumPy 1.20; for more details and guidance: https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations
  if value is None or (attr.numeric and (value == '' or np.isnan(np.float(value)))):
```

See numpy deprecation info here: https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations

There are several uses of `np.float`, `np.bool`, `np.object` in the codebase that raise the same deprecation warning above.

rly:
Fix #938.

Numpy 1.20 deprecated the use of numpy aliases of built-in types. Replacing the aliases with the built-in types does not change behavior but will silence the deprecation warning. 
https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,8 @@
 ## Release notes
 
-### 0.13.3 -- May 28, 2021
+### 0.13.3 -- TBD
 * Bugfix - Dependencies not properly loaded on populate. (#902) PR #919
+* Bugfix - Replace use of numpy aliases of built-in types with built-in type. (#938) PR #939
 
 ### 0.13.2 -- May 7, 2021
 * Update `setuptools_certificate` dependency to new name `otumat`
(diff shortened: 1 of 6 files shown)


guzman-raphael:
@rly Thanks for the PR! Looks like the tests failed because the LetsEncrypt cert has expired. Would you mind updating all mentions of `datajoint/nginx:v0.0.16` to `datajoint/nginx:v0.0.17`? The latest image should have the updated cert.

guzman-raphael:
@rly Awesome. Change looks good. If you would just update the changelog we should be then ready for a merge. Could you add an entry to the changelog under `0.13.3` in the following 2 places? The date used there seems to be incorrect so you can simply put `TBD` for now.
- https://github.com/datajoint/datajoint-python/blob/master/CHANGELOG.md
- https://github.com/datajoint/datajoint-python/blob/master/docs-parts/intro/Releases_lang1.rst

guzman-raphael (approved):
Excellent, thanks for the PR! :handshake:

Issue #911, opened 2021-04-25
File access via fsspec


chrisroat:
### Problem

Accessing various filesystems (s3, gcsfs, azure, gdrive, dropbox) with the current implementation requires new modules for each system, providing the file-like methods/attributes.  

### Requirements

Switch file-storage backend to use [fsspec](https://filesystem-spec.readthedocs.io/en/latest/?badge=latest).

### Justification

This change would allow users to automatically use any fsspec-supported backend by loading the corresponding module, i.e. GCS vs [gcsfs](https://github.com/dask/gcsfs).  Of course, the backend must support the operations needed by datajoint.

### Alternative Considerations

As stated in the problem, a custom module is currently necessary for a new file system.

### Related Errors

N/A

### Screenshots

N/A

### Additional Research and Context

The  fsspec project is providing a common, pythonic interface across many kinds of local and remote filesystems.  Several project have announced recently starting to use it -- for a major example, see the [pandas 1.1 release](https://pandas.pydata.org/docs/whatsnew/v1.1.0.html#fsspec-now-used-for-filesystem-handling).

dimitri-yatsenko:
Excellent idea.

yambottle:
This issue is closed because there has been no activity for more than 365 days, please reopen if it's still applicable.

PR #946, opened 2021-08-22
Add fsspec as alternate file protocol

chrisroat:
In principle, fsspec can completely subsume the current protocols (file and s3) and can provide many more.  However, since it is a new integration and getting credentialling correct can be difficult, I implemented it as an alternate protocol which people can migrate to.

Some s3fs and gcsfs dependencies (installed via fsspec[s3,gcs]) do not have wheels for alpine, so they need to be built during startup.   To correctly run the tests, the following extra install of dev packages is needed:

apk add musl-dev

Addresses #911

### DIFF
--- a/datajoint/external.py
+++ b/datajoint/external.py
@@ -1,5 +1,6 @@
 from pathlib import Path, PurePosixPath, PureWindowsPath
 from collections import Mapping
+from fsspec.core import url_to_fs
 from tqdm import tqdm
 from .settings import config
 from .errors import DataJointError, MissingExternalFile
@@ -44,6 +45,8 @@ class ExternalTable(Table):
         if self.spec['protocol'] == 'file' and not Path(self.spec['location']).is_dir():
             raise FileNotFoundError('Inaccessible local directory %s' %
                                     self.spec['location']) from None
+        if self.spec['protocol'] == 'fsspec':
+            self._fs, _ = url_to_fs(self.spec['location'])
 
     @property
     def definition(self):
@@ -84,6 +87,11 @@ class ExternalTable(Table):
         # Preserve root
         elif self.spec['protocol'] == 'file':
             return PurePosixPath(Path(self.spec['location']), relative_filepath)
+        elif self.spec['protocol'] == 'fsspec':
+            # The pathlib library strips the double slashes used in uris.  For now,
+            # just assume linux slash directory separator.
+            # TODO: Make this work on all systems, including Windows.
+            return self.spec['location'] + "/" + str(relative_filepath)
         else:
             assert False
 
@@ -97,6 +105,10 @@ class ExternalTable(Table):
             self.s3.fput(local_path, external_path, metadata)
         elif self.spec['protocol'] == 'file':
             safe_copy(local_path, external_path, overwrite=True)
+        elif self.spec['protocol'] == 'fsspec':
+            parent = self._fs._parent(external_path)
+            self._fs.makedirs(parent)
+            self._fs.put_file(local_path, external_path)
         else:
             assert False
 
@@ -105,6 +117,8 @@ class ExternalTable(Table):
             self.s3.fget(external_path, download_path)
         elif self.spec['protocol'] == 'file':
             safe_copy(external_path, download_path)
+        elif self.spec['protocol'] == 'fsspec':
+            self._fs.get_file(external_path, download_path)
         else:
             assert False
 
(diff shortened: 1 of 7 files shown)


chrisroat:
The CI failures are due to built wheels for dependencies of s3fs and gcsfs.  I believe this will require updates to the datajoint/pydev image to include additional dev packages.

chrisroat:
As per my comments above, I believe these failures need `apk add musl-dev` in `datajoint/pydev` in order to succeed.  If you'd like to move forward with this, we can work on making that update and resolving conflicts.

guzman-raphael:
@chrisroat Thanks for the PR and good stuff here man! :rocket:

We are currently playing a bit of some catch-up on other regression issues and new team members (they've recently patched 2 issues you opened!) but we haven't forgotten about this PR. We may be a bit slow in our response/review while adjusting to this new cycle so we appreciate your patience with this.

Regarding the issues of the docker images here, we are also currently in the process of upgrading it to a new image stack so we will likely not continue using `datajoint/pydev` in favor of `datajoint/djtest`. A PR for that update will soon be coming addressing that along with fixing the broken `datajoint/datajoint` image. Regarding the external storage support for `fsspec`, is this purely a Python dependency or are other system dependencies involved?

chrisroat:
Yes, fsspec is python only (AFAIK).  It actually doesn't seem to have any dependencies, as it is really just a specification.

There can be extra packages, depending on which underlying filesystem is being used (i.e. gcsfs for GCS, s3fs for S3) -- and those could have additional dependencies.  Datajoint could include some of those  packages/dependencies, or leave it up to the user to install what they need.  

I'd personally lean toward having the user install what they need, but the fact that datajoint currently brings in S3 dependencies may have already set a different precedent.

chrisroat:
Is there still interest in this?

horsto:
I want to use my dropbox business account as external drive location to my datajoint schema. Since dropbox is not S3, this option is currently not available - but, as I understand, this PR + some additions (https://github.com/fsspec/dropboxdrivefs), would enable it? I'd be grateful for any tips.

dimitri-yatsenko:
Ok, let's test and merge this and test with Dropbox. Thank you for finding this @horsto and thank you for your patience @chrisroat.

guzman-raphael:
@chrisroat DataJoint core team has opened some bandwidth and we are finally coming back to review the outstanding PRs that are in the queue. Thinking back on your contribution, I really like the idea of conforming to a single interface to simplify how we add support for new external storage options.

On that note, why did you choose `fsspec` as opposed to FUSE-based approaches? Is there an additional benefit to using the specification as you proposed?

Some additional context, we've now had much more experience using [s3fs-fuse](https://github.com/s3fs-fuse/s3fs-fuse) and it is demonstrating amazing performance! We've been mounting S3 BLOBs (>150GB) and accessing the files as if they were local from Python with read/write speeds consistent with online speedtests (900 Mbps in our HQ).

FUSE projects have been around for quite some time and the above in particular has: caching features, strong community acceptance, compatibility for STS, and more. Perhaps the *common* interface is simply treating objects like a normal file and letting FUSE do the rest? :shrug: 

Curious on your thoughts.

guzman-raphael:
[rclone](https://rclone.org/) seems like a very promising one as well.

chrisroat:
You can use fsspec with fuse-mounted filesystems, so I don't think they are competitors in that way.  See the docs:
https://filesystem-spec.readthedocs.io/en/latest/features.html#mount-anything-with-fuse

I think fsspec is heavily used in the pydata ecosystem (i.e. pandas, dask), so it is a natural to support it.  You'll benefit from improvements and interoperability.

Regarding rclone, I note that someone even tried to make an fsspec implementation for it (it doesn't look supported, though):
https://pypi.org/project/fsspec-rclone/

dimitri-yatsenko:
ok, let's integrate it before we add new file management

dimitri-yatsenko:
Hi @chrisroat, Thank you for your patience. We are working on merging this. We just need to redo the tests in `pytest`. If you are unavailable for this, we will re-open a new PR where we will fix this.

chrisroat:
Sorry.  I haven’t been working in this space for some time.

ethho:
@dimitri-yatsenko Okay, I'll take care of migration to pytest.

dimitri-yatsenko:
@chrisroat Thank you for your contribution. We will close your PR and @ethho will reopen a new one with your changes.

PR #947, opened 2021-08-23
Support pandas series and dataframe in blobs

chrisroat:
Data used in pandas objects may fit well into a Part table.  Or the user prefers to store pandas objects as opaque blobs.

### DIFF
--- a/datajoint/blob.py
+++ b/datajoint/blob.py
@@ -3,6 +3,7 @@
 compatibility with Matlab-based serialization implemented by mYm.
 """
 
+import io
 import zlib
 from itertools import repeat
 import collections
@@ -10,6 +11,7 @@ from decimal import Decimal
 import datetime
 import uuid
 import numpy as np
+import pandas as pd
 from .errors import DataJointError
 from .settings import config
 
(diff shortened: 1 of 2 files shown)


datajoint/blob.py
```diff
@@ -344,6 +349,16 @@ def pack_set(self, t):
         return b"\3" + len_u64(t) + b"".join(
             len_u64(it) + it for it in (self.pack_blob(i) for i in t))
 
+    def read_pandas(self):
+        return pd.read_pickle(io.BytesIO(self.read_blob()))
+
+    def pack_pandas(self, t):
+        buffer = io.BytesIO()
+        # t.to_pickle(buffer)  # bug in pandas 1.0.3, use pickle instead:
+        import pickle
```
dimitri-yatsenko, line 358: Thank you for this solution. I am hesitant to use pickle as a long-term serialization. Let me review a bit to ensure that this can be used as a long term strategy: the data need to be de-serializable 20 years from now.

chrisroat: Sure thing.  20 years is a long time horizon!

As another consideration, I am also making some changes in a fork to support dask output.  It's a little invasive, so I don't think it would be useful to datajoint.  I'm planning to use parquet for dataframes, with the pyarrow engine.  

I'd be happy to change this PR to parquet/pyarrow, or any other format you deem appropriate.

dimitri-yatsenko: We are also trying to keep the serialization framework language agnostic. Pickle is explicitly Python specific.

chrisroat: Yes, that's right.  Parquet could a better solution.  I'm not invested in any particular format, so defer to your expertise and constraints.

Issue #952, opened 2021-08-27
remove_object failure should be exception, not error return value


chrisroat:
Should this `return` be a `raise`?
https://github.com/datajoint/datajoint-python/blob/5b81247fdb5861ebef098a5d4cbfe26098143eda/datajoint/s3.py#L87

guzman-raphael:
@chrisroat Thanks for reviewing this! It does appear that it would make more sense for this to raise an error. I'll mark it as a bug for now.

PR #954, opened 2021-09-02
Raise instead of return

normyp:
Fixes https://github.com/datajoint/datajoint-python/issues/952

### DIFF
--- a/datajoint/s3.py
+++ b/datajoint/s3.py
@@ -84,4 +84,4 @@ class Folder:
         try:
             self.client.remove_object(self.bucket, str(name))
         except minio.ResponseError:
-            return errors.DataJointError('Failed to delete %s from s3 storage' % name)
+            raise errors.DataJointError('Failed to delete %s from s3 storage' % name)


guzman-raphael:
@normyp Thanks for the PR! The change here looks pretty straightforward. It would be good to add some integration tests to provide coverage over this since it was missed before. We know that tests can be a bit tricky to write so we can have someone from our team collaborate with you on this.

@jverswijver Would you help us add some tests here and update the changelog too by issuing a PR to @normyp?

jverswijver:
Yes @guzman-raphael, I am working on a test that ensures that the error thrown is always raised. When that is ready I will pull request into @normyp's fork.

jverswijver:
@normyp Thank you for your pull request! After further review the solution to this issue and corresponding unit test was much more complicated than we anticipated. You can see our solution at this pull request #955. I have revised my commit history to include your contribution which you can verify [here](https://github.com/jverswijver/datajoint-python/commits/external_remove_bug). I will be closing this pull request as it is superseded by PR #955.

PR #956, opened 2021-09-09
Fix logic bug in external file clean up

### LINKED ISSUE #953: ExternalTable.delete should not remove row on error
### Description

When deleting from an `External` table with `delete_external_files=True`, the rows are deleted from the table followed by deleting files from the filesystem.  This behavior means an error in file deletion will leave excess data present with no record of it in the database.  For large pieces of data, this could be undesired.  And since an error is not raised, many clients won't realize there is a return value and check it (mosts/all tests do not even check this).

https://github.com/datajoint/datajoint-python/blob/5b81247fdb5861ebef098a5d4cbfe26098143eda/datajoint/external.py#L339-L350

### Expected Behavior

Instead of errors being returned, I suggest:
- an error be raised if there is a problem
- the rows should not be deleted from the external table

zitrosolrac:
fix #953


datajoint/external.py
```diff
@@ -338,7 +338,7 @@ def delete(self, *, delete_external_files=None, limit=None, display_progress=Tru
             error_list = []
             for uuid, external_path in items:
                 try:
-                    count = (self & {'hash': uuid}).delete_quick(get_count=True)  # optimize
+                    count = len(self & {'hash': uuid})  # optimize
                 except Exception:
                     pass   # if delete failed, do not remove the external file
```
dimitri-yatsenko, line 343: The comment is stale. No delete has been attempted.

dimitri-yatsenko, line 343: What exception is expected from the `len` query? Why is this helpful?

datajoint/external.py
```diff
@@ -338,7 +338,7 @@ def delete(self, *, delete_external_files=None, limit=None, display_progress=Tru
             error_list = []
             for uuid, external_path in items:
                 try:
-                    count = (self & {'hash': uuid}).delete_quick(get_count=True)  # optimize
+                    count = len(self & {'hash': uuid})  # optimize
```
dimitri-yatsenko, line 341: Why do we need to get the count? Why do we need an extra query rather than deleting directly?

datajoint/external.py
```diff
@@ -347,6 +347,8 @@ def delete(self, *, delete_external_files=None, limit=None, display_progress=Tru
                         self._remove_external_file(external_path)
```
dimitri-yatsenko, line 347: Why is it better to remove the external file before deleting from the external table. What if the delete from the external table fails?

dimitri-yatsenko (changes requested):

datajoint/external.py
```diff
@@ -338,7 +338,7 @@ def delete(self, *, delete_external_files=None, limit=None, display_progress=Tru
             error_list = []
             for uuid, external_path in items:
                 try:
```
guzman-raphael, line 340: Looks like lines [340-345](https://github.com/zitrosolrac/datajoint-python/blob/84e4aa4aac92a3cf135230cf54a5593e1851ab37/datajoint/external.py#L340-L345) are unnecessary here.

datajoint/external.py
```diff
@@ -338,7 +338,7 @@ def delete(self, *, delete_external_files=None, limit=None, display_progress=Tru
             error_list = []
             for uuid, external_path in items:
```
guzman-raphael, line 339: Thinking it might be safer to perform a `LOCK` on the external table before starting the loop to ensure that no references are created while the remote file is being deleted. Just have to be sure that we `UNLOCK` so that we don't cause issues for others. This will reinforce that this is an admin operation and should be done carefully.

https://dev.mysql.com/doc/refman/5.7/en/lock-tables.html

guzman-raphael (changes requested):
@zitrosolrac Thanks for the PR! :handshake: 

Need a few more refinements to the delete algorithm, test, and a bit more clean up.
Also, make sure to remove the binary files that were accidentally committed.

guzman-raphael:
@zitrosolrac oh forgot one thing. Can you please update the changelog and `doc-parts` release notes as well.

guzman-raphael (changes requested):
@zitrosolrac Nice work man! Just some minor things left. :clap:

guzman-raphael (approved):
@zitrosolrac :tada: WooHoo! :tada: Nice job on a tricky one :+1:

datajoint/external.py
```diff
@@ -337,18 +340,19 @@ def delete(self, *, delete_external_files=None, limit=None, display_progress=Tru
...
+                        (self & {'hash': uuid}).delete_quick()
+                    except Exception:
+                        pass   # if delete failed, do not remove the external file
                     else:
-                        (self & {'hash': uuid}).delete_quick(get_count=True)
+                        try:
+                            self._remove_external_file(external_path)
+                        except Exception as error:
+                            # adding row back into table after failed delete
+                            self.insert1(row[0])
```
dimitri-yatsenko, line 355: set `skip_duplicates=True` to address potential race conditions.

dimitri-yatsenko: I think it's okay not to re-insert if the delete fails. The procedure will return the undeleted files to the user for special handling. Without implementing an actual transaction management process, we can allow some orphaned files in the external storage and provide an additional cleanup utility.

guzman-raphael: @dimitri-yatsenko I feel this is actually necessary. There is basically 2 primary concerns:
- **The user has only one chance to catch the errors**. Meaning that running the external delete/clean up multiple times will yield no change though it has only been removed from the external tracking table. This gives a false sense that the system has been 'cleaned-up' but it has not been removed from the actual store (this is the main complaint from the user in issue #953). We can provide an additional utility to resolve this but that would be expensive to run as it would have to 'crawl' the entire store to find objects that don't exist within the external tracking table. By simply inserting it back when there is an exception, we are allowing the error to be more visible but most importantly reproducible.
- `schema.external[store].delete()` is not currently documented that it returns a list of the errors. Users are most likely unaware of this feature and therefore aren't using it properly. @zitrosolrac Could you open an issue on this in our [datajoint-docs](https://github.com/datajoint/datajoint-docs) and reference it in #953? (Filing for now since our new team members haven't been oriented to the docs setup yet).

guzman-raphael:
```suggestion
                            self.insert1(row[0], skip_duplicates=True)
```

guzman-raphael (changes requested):

guzman-raphael (approved):

datajoint/external.py
zitrosolrac: @guzman-raphael Got it! I'll open an issue on this in the datajoint-docs and add the necessary reference.

dimitri-yatsenko (approved):

PR #959, opened 2021-09-10
Update test_s3 to work for undockerized testing

### LINKED ISSUE #958: Fix s3_test to work for users running the test suite not in a docker container
### Description
When users run the test_s3.py as part of out test suite they will get an error when they do not have minio command line client installed, additionally the alias for the minio login is set within the docker-compose file which will lead to the test breaking when non docker users run the test.

### Expected Behavior
Users should be able to run the test suite as long as they have minio command line client installed.

### Current workaround
- users install minio command line client and add it to their path so they can use it in the terminal
- run this command to add the alias for minio before running tests `mc alias set myminio/ http://fakeservices.datajoint.io datajoint datajoint`

### Possible solution
- Remove minio client alias from docker-compose and do it in test_s3.py
- Update readme to include minio client install

jverswijver:
Fixes #958

### DIFF
--- a/LNX-docker-compose.yml
+++ b/LNX-docker-compose.yml
@@ -79,7 +79,6 @@ services:
         set -e
         wget -P /home/dja/.local/bin/ https://dl.min.io/client/mc/release/linux-amd64/mc
         chmod +x /home/dja/.local/bin/mc
-        mc alias set myminio/ http://minio:9000 datajoint datajoint
         pip install --user -r test_requirements.txt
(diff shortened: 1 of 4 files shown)


jverswijver:
I believe this is related to the errors you were getting this afternoon @dimitri-yatsenko when running tests

guzman-raphael (approved):
@jverswijver Nice one :clap:

dimitri-yatsenko (approved):

PR #960, opened 2021-09-15
Remove MC dependency.

jverswijver:
I have found a way to run the test_S3.py test_remove_object exception without using minio client MC.

We should no longer get these errors https://github.com/datajoint/datajoint-python/runs/3571645577?check_suite_focus=true#step:6:2203 

I have also updated the readme to remove instructions to install minio client and updated the docker compose files to remove the minio client install as well.

### DIFF
--- a/tests/test_s3.py
+++ b/tests/test_s3.py
@@ -60,13 +60,6 @@ class TestS3:
     def test_remove_object_exception():
         # https://github.com/datajoint/datajoint-python/issues/952
 
-        # Initialize minioClient with an endpoint and access/secret keys.
-        minio_client = Minio(
-            'minio:9000',
-            access_key='jeffjeff',
-            secret_key='jeffjeff',
-            secure=False)
-
         # Insert some test data and remove it so that the external table is populated
         test = [1, [1, 2, 3]]
         SimpleRemote.insert1(test)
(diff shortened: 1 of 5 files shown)


jverswijver:
No longer gets an error in the github actions run: https://github.com/datajoint/datajoint-python/pull/960/checks?check_run_id=3610904141#step:6:1979

tests/test_s3.py
```diff
@@ -124,20 +75,16 @@ def test_remove_object_exception():
         # Save the old external table minio client
         old_client = schema.external['share'].s3.client
 
-        # Apply our new minio client to the external table that has permissions restrictions
+        # Apply our new minio client which has a user that does not exist
         schema.external['share'].s3.client = minio_client
```
guzman-raphael, line 79: Since `minio_client` is now only being called once, I wouldn't assign it to a variable but simply instantiate your client directly here. A bit easier to follow the logic in your test that way.

guzman-raphael (changes requested):
@jverswijver Thanks for optimizing away the `mc` dependency. :+1: 
Requesting only a minor test change to make your logic a bit easier to follow.

guzman-raphael (approved):

PR #963, opened 2021-09-17
Fetch order by bug

### LINKED ISSUE #914: Bug in fetching with `order_by` specified on attributes that are not part of the returned attribute list
### Description
DataJoint encountered error in fetching with `order_by` specified on attributes that are not part of the returned attribute list. (error stack pasted below)

This line yields the error stack below:

```
scan.ScanInfo.Field.fetch('KEY', order_by='field_z')
```

It seems like fetching just the 'KEY' part does the projection before the ordering, so order_by=field_z couldn't find field_z.
Fetch everything works just fine:

```
scan.ScanInfo.Field.fetch(order_by='field_z')
```

Error stack:
```
  File "C:\ProgramData\Miniconda3\envs\workflow-imaging\lib\site-packages\IPython\core\interactiveshell.py", line 3437, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-118-2c9725558855>", line 1, in <module>
    scan.ScanInfo.Field.fetch('KEY', order_by='field_z')
  File "C:\ProgramData\Miniconda3\envs\workflow-imaging\lib\site-packages\datajoint\fetch.py", line 182, in __call__
    format='array')
  File "C:\ProgramData\Miniconda3\envs\workflow-imaging\lib\site-packages\datajoint\fetch.py", line 193, in __call__
    as_dict=as_dict, limit=limit, offset=offset, order_by=order_by)
  File "C:\ProgramData\Miniconda3\envs\workflow-imaging\lib\site-packages\datajoint\expression.py", line 519, in cursor
    return self.connection.query(sql, as_dict=as_dict)
  File "C:\ProgramData\Miniconda3\envs\workflow-imaging\lib\site-packages\datajoint\connection.py", line 302, in query
    self._execute_query(cursor, query, args, suppress_warnings)
  File "C:\ProgramData\Miniconda3\envs\workflow-imaging\lib\site-packages\datajoint\connection.py", line 268, in _execute_query
    raise translate_query_error(err, query)
  File "C:\ProgramData\Miniconda3\envs\workflow-imaging\lib\site-packages\datajoint\connection.py", line 266, in _execute_query
    cursor.execute(query, args)
  File "C:\ProgramData\Miniconda3\envs\workflow-imaging\lib\site-packages\pymysql\cursors.py", line 170, in execute
    result = self._query(query)
  File "C:\ProgramData\Miniconda3\envs\workflow-imaging\lib\site-packages\pymysql\cursors.py", line 328, in _query
    conn.query(q)
  File "C:\ProgramData\Miniconda3\envs\workflow-imaging\lib\site-packages\pymysql\connections.py", line 517, in query
    self._affected_rows = self._read_query_result(unbuffered=unbuffered)
  File "C:\ProgramData\Miniconda3\envs\workflow-imaging\lib\site-packages\pymysql\connections.py", line 732, in _read_query_result
    result.read()
  File "C:\ProgramData\Miniconda3\envs\workflow-imaging\lib\site-packages\pymysql\connections.py", line 1075, in read
    first_packet = self.connection._read_packet()
  File "C:\ProgramData\Miniconda3\envs\workflow-imaging\lib\site-packages\pymysql\connections.py", line 684, in _read_packet
    packet.check_error()
  File "C:\ProgramData\Miniconda3\envs\workflow-imaging\lib\site-packages\pymysql\protocol.py", line 220, in check_error
    err.raise_mysql_exception(self._data)
  File "C:\ProgramData\Miniconda3\envs\workflow-imaging\lib\site-packages\pymysql\err.py", line 109, in raise_mysql_exception
    raise errorclass(errno, errval)
pymysql.err.InternalError: (3065, "Expression #1 of ORDER BY clause is not in SELECT list, references column 'neuro_imaging._scan_info__field.field_z' which is not in SELECT list; this is incompatible with DISTINCT")
```

### Reproducibility
Include:
- OS (WIN)
- Python Version 3.7
- MySQL Version 5.7
- MySQL Deployment Strategy (local-native)
- DataJoint Version (0.13.1)

jverswijver:
Added test based off of linked issue. It seems enforcing DISTINCT when we are making sql queries with primary keys was not useful, I have removed it for now.

fix #914

### DIFF
--- a/datajoint/expression.py
+++ b/datajoint/expression.py
@@ -106,9 +106,7 @@ class QueryExpression:
         Make the SQL SELECT statement.
         :param fields: used to explicitly set the select attributes
         """
-        distinct = self.heading.names == self.primary_key
-        return 'SELECT {distinct}{fields} FROM {from_}{where}'.format(
-            distinct="DISTINCT " if distinct else "",
+        return 'SELECT {fields} FROM {from_}{where}'.format(
             fields=self.heading.as_sql(fields or self.heading.names),
             from_=self.from_clause(), where=self.where_clause())
 
--- a/tests/test_fetch.py
+++ b/tests/test_fetch.py
@@ -296,4 +296,3 @@ class TestFetch:
         # This command is confirmed to work in v0.12.9 but not in v0.13.2
 
         Parent().fetch('KEY', order_by='name')
-
--- a/tests/test_fetch.py
+++ b/tests/test_fetch.py
@@ -287,3 +287,7 @@ class TestFetch:
 
         # reset cache directory state (will fail if purge was unsuccessful)
         os.rmdir(os.path.expanduser('~/dj_query_cache'))
+    
+    def test_fetch_group_by(self):
+        # nosetests -vs --tests=tests.test_fetch:TestFetch.test_fetch_group_by --nologcapture
+        schema.Parent.fetch('KEY', order_by='name')
--- a/tests/test_fetch.py
+++ b/tests/test_fetch.py
@@ -6,6 +6,7 @@ import decimal
 import pandas
 import warnings
 from . import schema
+from .schema import Parent
 import datajoint as dj
 import os
 
@@ -287,7 +288,12 @@ class TestFetch:
 
         # reset cache directory state (will fail if purge was unsuccessful)
         os.rmdir(os.path.expanduser('~/dj_query_cache'))
-    
+
     def test_fetch_group_by(self):
         # nosetests -vs --tests=tests.test_fetch:TestFetch.test_fetch_group_by --nologcapture
-        schema.Parent.fetch('KEY', order_by='name')
+        # https://github.com/datajoint/datajoint-python/issues/914
+
+        # This command is confirmed to work in v0.12.9 but not in v0.13.2
+
+        Parent().fetch('KEY', order_by='name')
+
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -20,7 +20,7 @@ from datajoint.utils import parse_sql
 __author__ = 'Edgar Walker, Fabian Sinz, Dimitri Yatsenko, Raphael Guzman'
 
 # turn on verbose logging
-logging.basicConfig(level=logging.DEBUG)
+# logging.basicConfig(level=logging.DEBUG)
 
 __all__ = ['__author__', 'PREFIX', 'CONN_INFO']
 


tests/test_fetch.py
```diff
@@ -287,3 +288,8 @@ def test_query_caching(self):
 
         # reset cache directory state (will fail if purge was unsuccessful)
         os.rmdir(os.path.expanduser('~/dj_query_cache'))
+
+    def test_fetch_group_by(self):
+        # https://github.com/datajoint/datajoint-python/issues/914
+
+        Parent().fetch('KEY', order_by='name')
```
guzman-raphael, line 295: Would you add an `assert` statement here to be specific on what we are expecting from this query used in this example?

datajoint/expression.py
```diff
@@ -106,9 +106,7 @@ def make_sql(self, fields=None):
         Make the SQL SELECT statement.
         :param fields: used to explicitly set the select attributes
         """
-        distinct = self.heading.names == self.primary_key
-        return 'SELECT {distinct}{fields} FROM {from_}{where}'.format(
-            distinct="DISTINCT " if distinct else "",
+        return 'SELECT {fields} FROM {from_}{where}'.format(
```
guzman-raphael, line 109: Can you also explore all usage of `DISTINCT`. Noticed that there is another similar usage in the `Aggregation` class. Did you mean to remove its use here but maintain its use there?

guzman-raphael (changes requested):
Thanks for the PR @jverswijver! Primarily just have some questions about other similar usage of `DISTINCT` in source.

tests/test_fetch.py
jverswijver: addressed this in commit ce16d0e

datajoint/expression.py
jverswijver: I looked through it a bit and saw that it also uses `DISTINCT`, but it might be best to just leave it until a user reports a specific issue that is related to the aggregation usage of distinct. It would be a bad idea to remove it without a specific reason/issue.

tests/test_fetch.py
```diff
@@ -287,3 +288,12 @@ def test_query_caching(self):
 
         # reset cache directory state (will fail if purge was unsuccessful)
         os.rmdir(os.path.expanduser('~/dj_query_cache'))
+
+    def test_fetch_group_by(self):
+        # https://github.com/datajoint/datajoint-python/issues/914
+
+        try:
+            Parent().fetch('KEY', order_by='name')
+            assert True
```
dimitri-yatsenko, line 297: This has no effect.

tests/test_fetch.py
```diff
@@ -287,3 +288,12 @@ def test_query_caching(self):
 
         # reset cache directory state (will fail if purge was unsuccessful)
         os.rmdir(os.path.expanduser('~/dj_query_cache'))
+
+    def test_fetch_group_by(self):
+        # https://github.com/datajoint/datajoint-python/issues/914
+
+        try:
```
dimitri-yatsenko, line 295: This `try` ... `except` does not appear to be necessary. Just letting the exception raise will have the same effect.

dimitri-yatsenko (changes requested):

dimitri-yatsenko:
`DISTINCT` becomes important in queries of the following types:
```python
dj.U('attr') & q
```
where `attr` is not the primary key of `q`.

Another case is unions. 
```python
q + r
```

If we don't have test cases that require `DISTINCT`, let's add them

jverswijver:
A few updates:

- github actions are failing because an api we use for code coverage is down for maintenance but the code up to commit 13b77d4 passes the test suite on my machine.
- I have added an additional test that ensures that the results of dj.U operations are unique.
- I have added the _distinct property to the QueryExpression class so we can be explicit when we want to use `DISTINCT` in our query formulation.
- I have updated the U class to use this new _distinct property to ensure that the results produced are unique like they were before when we blanket applied `DISTINCT`.

tests/test_fetch.py
```diff
@@ -287,3 +288,41 @@ def test_query_caching(self):
...
+
+    def test_fetch_group_by(self):
+        # https://github.com/datajoint/datajoint-python/issues/914
+
+        expected_data = [{'parent_id': 1}]
+        fetched_data = Parent().fetch('KEY', order_by='name')
+        print(fetched_data)
+        assert fetched_data == expected_data
+
+    def test_dj_U_DISTINCT(self):
```
guzman-raphael, line 300: Following python's `snake_case` convention for functional naming.

```suggestion
    def test_dj_u_distinct(self):
```

Otherwise, a great test. :+1:

datajoint/expression.py
```diff
@@ -101,14 +104,13 @@ def where_clause(self):
         return '' if not self.restriction else ' WHERE(%s)' % ')AND('.join(
             str(s) for s in self.restriction)
 
-    def make_sql(self, fields=None):
+    def make_sql(self, fields=None, distinct=True):
```
guzman-raphael, line 107: Hmm, don't think this is serving any purpose now, right?

```suggestion
    def make_sql(self, fields=None):
```

guzman-raphael (changes requested):
@jverswijver Nice work in determining the root cause for this one! :muscle:
Only providing some minor feedback here.

guzman-raphael (approved):
@jverswijver Nice job man! :chart_with_upwards_trend:

dimitri-yatsenko (changes requested):

PR #966, opened 2021-09-23
Fix __len__() for left join QueryExpressions

### LINKED ISSUE #951: Inconsistent lengths between QueryExpression table and fetched object
### Description
The view of a QueryExpression table and the `len()` method for that table report different length values from the fetched object. 
### Reproducibility
Include:
- OS MACOS 11.5.2
- Python Version 3.9.6
- MySQL Version mysql-docker from datajoint
- MySQL Deployment Strategy local-docker
- DataJoint Version 1.13.2
- Minimum number of steps to reliably reproduce the issue

```python
import datajoint as dj
import numpy as np
import datetime

dj.conn()

dj.Schema("dj_test").drop(force=True)
schema = dj.Schema("dj_test")


@schema
class Parent(dj.Manual):
    definition = """
    # Parent table
    id: int
    ---
    txt = "NA" : varchar(32)
    """


@schema
class Child(dj.Computed):
    definition = """
    # Child table
    -> Parent
    start_time: datetime(6)
    ---
    timestamps: longblob
    """

    def make(self, key):
        t = np.random.poisson(1000, 10)
        t.sort()
        self.insert1(
            {
                "start_time": datetime.datetime.now(),
                "timestamps": t,
                **key,
            }
        )
        print(f"ingested w/ {key}")


if __name__ == "__main__":
    Parent.insert1({"id": 1})
    Child.populate()
    Parent.insert([{"id": 2}, {"id": 3}, {"id": 4}])

    q = Parent.join(Child & "id != 1", left=True)
    qf = q.fetch()

    print("\nQuery Table:\n", q)
    print("Fetched from Query Table:\n", qf)
    print("\nLength of Query Table: ", len(q))
    print("Length of Fetched obj: ", len(qf))
```

Output:

```
Connecting root@localhost:3307
ingested w/ {'id': 1}

Query Table:
 *id    *start_time    txt     timestamps
+----+ +------------+ +-----+ +--------+
2      None           NA      =BLOB=    
3      None           NA      =BLOB=    
4      None           NA      =BLOB=    
 (Total: 0)

Fetched from Query Table:
 [(2, None, 'NA', None) (3, None, 'NA', None) (4, None, 'NA', None)]

Length of Query Table:  0
Length of Fetched obj:  3
```

### Expected Behavior
I expect the lengths of the QueryExpression and length displayed in the view to be the same as the length of the fetched object.

jverswijver:
Left joins in SQL have the possibility of producing entire columns of `NULL` values which leads to some difficulty when we want to know the length of a query expression that has used a left join due to the way `COUNT()` behaves.

When `COUNT()` gets any column passed to it that is entirely `NULL` it returns 0, it does not care if other columns within the parenthesis are populated with something. It is due to this that our current implementation of `__len__` does not produce correct results for certain left joins.

I have resolved this by adding a special count case for when the query expression contains a left join.

Fix #951

### DIFF
--- a/datajoint/expression.py
+++ b/datajoint/expression.py
@@ -439,6 +439,14 @@ class QueryExpression:
 
     def __len__(self):
         """:return: number of elements in the result set e.g. ``len(q1)``."""
+        try:
+            if self._left[0]:
+                return self.connection.query(
+                    'SELECT distinct count(*) FROM {from_}{where}'.format(
+                        from_=self.from_clause(),
+                        where=self.where_clause())).fetchone()[0]
+        except IndexError:
+            pass
         return self.connection.query(
             'SELECT count(DISTINCT {fields}) FROM {from_}{where}'.format(
                 fields=self.heading.as_sql(self.primary_key, include_aliases=False),
--- a/tests/test_aggr_regressions.py
+++ b/tests/test_aggr_regressions.py
@@ -6,7 +6,8 @@ import itertools
 from nose.tools import assert_equal, raises
 import datajoint as dj
 from . import PREFIX, CONN_INFO
-
+import datetime
+import numpy as np
 schema = dj.Schema(PREFIX + '_aggr_regress', connection=dj.conn(**CONN_INFO))
 
 # --------------- ISSUE 386 -------------------
@@ -94,6 +95,38 @@ class X(dj.Lookup):
     contents = zip(range(10))
 
 
+@schema
+class Parent(dj.Manual):
+    definition = """
+    # Parent table
+    id: int
+    ---
+    txt = "NA" : varchar(32)
+    """
+
+
+@schema
+class Child(dj.Computed):
+    definition = """
+    # Child table
+    -> Parent
+    start_time: datetime(6)
+    ---
+    timestamps: longblob
+    """
+
+    def make(self, key):
+        t = np.random.poisson(1000, 10)
+        t.sort()
+        self.insert1(
+            {
+                "start_time": datetime.datetime.now(),
+                "timestamps": t,
+                **key,
+            }
+        )
+        print(f"ingested w/ {key}")
+
 def test_issue558_part1():
     q = (A-B).proj(id2='3')
     assert_equal(len(A - B), len(q))
@@ -103,3 +136,11 @@ def test_issue558_part2():
     d = dict(id=3, id2=5)
     assert_equal(len(X & d), len((X & d).proj(id2='3')))
 
+
+def test_left_join_len():
+    Parent.insert1({"id": 1})
+    Child.populate()
+    Parent.insert([{"id": 2}, {"id": 3}, {"id": 4}])
+    q = Parent.join(Child & "id != 1", left=True)
+    qf = q.fetch()
+    assert len(q) == len(qf)


datajoint/expression.py
```diff
@@ -439,6 +439,14 @@ def tail(self, limit=25, **fetch_kwargs):
 
     def __len__(self):
         """:return: number of elements in the result set e.g. ``len(q1)``."""
+        try:
+            if self._left[0]:
```
dimitri-yatsenko, line 443: This does not appear optimally factored code since the two queries only differ by their fields.

dimitri-yatsenko (changes requested):

datajoint/expression.py
```diff
@@ -439,6 +439,14 @@ def tail(self, limit=25, **fetch_kwargs):
...
+                return self.connection.query(
+                    'SELECT distinct count(*) FROM {from_}{where}'.format(
+                        from_=self.from_clause(),
+                        where=self.where_clause())).fetchone()[0]
+        except IndexError:
```
dimitri-yatsenko, line 448: The exception does not seem like the most expressive way to check whether there are joins or not.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 * Bugfix - Replace use of numpy aliases of built-in types with built-in type. (#938) PR #939
 * Bugfix - `ExternalTable.delete` should not remove row on error (#953) PR #956
 * Bugfix - Fix error handling of remove_object function in `s3.py` (#952) PR #955
+* Bugfix - Fix __len__() for left join QueryExpressions (#951) PR #966
```
guzman-raphael, line 8:
```suggestion
* Bugfix - Fix count for left-joined `QueryExpressions` (#951) PR #966
```

docs-parts/intro/Releases_lang1.rst
```diff
@@ -4,6 +4,7 @@
 * Bugfix - Replace use of numpy aliases of built-in types with built-in type. (#938) PR #939
 * Bugfix - `ExternalTable.delete` should not remove row on error (#953) PR #956
 * Bugfix - Fix error handling of remove_object function in `s3.py` (#952) PR #955
+* Bugfix - Fix __len__() for left join QueryExpressions (#951) PR #966
```
guzman-raphael, line 7:
```suggestion
* Bugfix - Fix count for left-joined ``QueryExpressions`` (#951) PR #966
```

datajoint/expression.py
guzman-raphael: I agree. @jverswijver There looks like there could be a simple condition we can introduce here with the combination of a ternary + string formatting operation. Let's tag up on this together since that will likely be a bit easier to demonstrate.

tests/test_aggr_regressions.py
```diff
@@ -94,6 +95,38 @@ class X(dj.Lookup):
...
+        t.sort()
+        self.insert1(
+            {
+                "start_time": datetime.datetime.now(),
+                "timestamps": t,
+                **key,
+            }
+        )
+        print(f"ingested w/ {key}")
+
```
guzman-raphael, line 129: @jverswijver Think we can possibly repurpose the following tables for what these were used for:
- `tests/schema_uuid.py:Topic`
- `tests/schema_uuid.py:Item`

If not, then it would just be best to place these table a bit higher up so they could be reused in the future.

guzman-raphael (changes requested):
@jverswijver Almost there man but it is coming together. Let's tag up on this one individually for a bit to get it over the finish line.  :raccoon: :checkered_flag:

datajoint/expression.py
```diff
@@ -441,17 +441,11 @@ def tail(self, limit=25, **fetch_kwargs):
...
-                    'SELECT distinct count(*) FROM {from_}{where}'.format(
-                        from_=self.from_clause(),
-                        where=self.where_clause())).fetchone()[0]
-        except IndexError:
-            pass
         return self.connection.query(
-            'SELECT count(DISTINCT {fields}) FROM {from_}{where}'.format(
-                fields=self.heading.as_sql(self.primary_key, include_aliases=False),
+            'SELECT {select_} FROM {from_}{where}'.format(
+                select_=('DISTINCT count(*)' if any(self._left)
```
dimitri-yatsenko, line 446: `DISTINCT count` does not seem like the right syntax.

tests/test_aggr_regressions.py
```diff
@@ -6,7 +6,10 @@
 from nose.tools import assert_equal, raises
 import datajoint as dj
 from . import PREFIX, CONN_INFO
-
+import datetime
+import numpy as np
+import uuid
```
guzman-raphael, line 11: These appear to no longer be necessary.
```suggestion
```

guzman-raphael (approved):

datajoint/expression.py
jverswijver: COUNT(*) returns the number of rows in a specified table, and it preserves duplicate rows. In order to get a proper count of the rows on a left join `DISTINCT` is needed. With a regular count you would put `DISTINCT` inside the parenthesis but COUNT(*) is special and you need to put it outside.

guzman-raphael: Thanks @dimitri-yatsenko. That is right, `DISTINCT` has no effect here and can be removed.

dimitri-yatsenko: As we discussed, `DISTINCT COUNT(*)` does not distinct entries.

jverswijver: I have removed `DISTINCT COUNT(*)` in commit 15146b1

PR #971, opened 2021-11-01
Keyword arguments for populate()

Hendriela:
This pull request implements optional keyword arguments in `populate()`, which are passed down to each `make()` call.

This implementation is tested and works with our database. It allows for custom arguments in `make()`, while not affecting normal `make()` methods without arguments.

It is for example useful for bool flags whether to save intermediate results during a pipeline step.

### DIFF
--- a/datajoint/autopopulate.py
+++ b/datajoint/autopopulate.py
@@ -96,7 +96,7 @@ class AutoPopulate:
 
     def populate(self, *restrictions, suppress_errors=False, return_exception_objects=False,
                  reserve_jobs=False, order="original", limit=None, max_calls=None,
-                 display_progress=False, **kwargs):
+                 display_progress=False, make_kwargs=None):
         """
         rel.populate() calls rel.make(key) for every primary key in self.key_source
         for which there is not already a tuple in rel.
@@ -108,7 +108,7 @@ class AutoPopulate:
         :param display_progress: if True, report progress_bar
         :param limit: if not None, checks at most that many keys
         :param max_calls: if not None, populates at max that many keys
-        :param kwargs: additional custom keyword arguments that will be passed down to each make() call
+        :param kwargs: optional dict containing arguments that will be passed down to each make() call
         """
         if self.connection.in_transaction:
             raise DataJointError('Populate cannot be called during a transaction.')
@@ -151,8 +151,8 @@ class AutoPopulate:
                     call_count += 1
                     self.__class__._allow_insert = True
                     try:
-                        if kwargs:
-                            make(dict(key), **kwargs)
+                        if make_kwargs is not None:
+                            make(dict(key), **make_kwargs)
                         else:
                             make(dict(key))
                     except (KeyboardInterrupt, SystemExit, Exception) as error:
--- a/datajoint/autopopulate.py
+++ b/datajoint/autopopulate.py
@@ -8,10 +8,35 @@ from tqdm import tqdm
 from .expression import QueryExpression, AndList
 from .errors import DataJointError, LostConnectionError
 import signal
+import multiprocessing as mp
 
 # noinspection PyExceptionInherit,PyCallingNonCallable
 
 logger = logging.getLogger(__name__)
+
+
+# --- helper functions for multiprocessing --
+
+def _initialize_populate(table, jobs, populate_kwargs):
+    """
+    Initialize the process for mulitprocessing.
+    Saves the unpickled copy of the table to the current process and reconnects.
+    """
+    process = mp.current_process()
+    process.table = table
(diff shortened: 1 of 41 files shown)


dimitri-yatsenko:
The idea of workflows is that they must be reproducible and that input data must be captured (data provenance). Addition custom inputs into `make` would not be captured. Perhaps we could allow additional arguments but provide a mechanism for storing them in the database as well for complete provenance. Or we could add this feature but document the principles clearly and ensure that users understand that it's on them to ensure reproducibility and provenance.

datajoint/autopopulate.py
```diff
@@ -150,7 +151,10 @@ def handler(signum, frame):
                     call_count += 1
                     self.__class__._allow_insert = True
                     try:
-                        make(dict(key))
+                        if kwargs:
+                            make(dict(key), **kwargs)
+                        else:
+                            make(dict(key))
```
dimitri-yatsenko, line 157:
```suggestion
                        make(dict(key), **kwargs)
```

datajoint/autopopulate.py
```diff
@@ -96,7 +96,7 @@ def _jobs_to_do(self, restrictions):
 
     def populate(self, *restrictions, suppress_errors=False, return_exception_objects=False,
                  reserve_jobs=False, order="original", limit=None, max_calls=None,
-                 display_progress=False):
+                 display_progress=False, **kwargs):
```
dimitri-yatsenko, line 99: rather than using `**kwargs`, use an explicit named argument, something like `make_kwargs` containing the dict of all the custom parameters. This is because `populate` already has a lot of keyword arguments. If someone misspells one of the regular keyword arguments, then rather than raising an error, populate will simply pass the misspelled argument into `**kwargs`.

Hendriela:
> 
> 
> The idea of workflows is that they must be reproducible and that input data must be captured (data provenance). Additional custom inputs into `make` are not captured in the database.
> 
> Perhaps we could allow additional arguments but provide a mechanism for storing them in the database as well for complete provenance. Or we could add this feature but document the principles clearly and ensure that users understand that it's on them to ensure reproducibility and provenance.

That is true. This change would not be suitable to pass down arguments that affect the analysis itself. 

We use it, as I said, for controlling peripheral processes like plotting intermediate results, or exporting the results in addition to inserting them in the table. Because these things do not influence the pipeline, but should be more flexible than storing them together with other parameters or hard-code them, I think it is fair if they are not captured in the database.

I agree that if users are provided with this option, workflow reproducibility can easily be (consciously or accidentally) broken. They would have to be clearly aware of the intended use of this feature (similar to the dangers of `update()`).

datajoint/autopopulate.py
Hendriela: I implemented your suggestion in the next commit #2d62835. Feel free to close the pull request if this feature does not align with your DataJoint philosophy 👍

dimitri-yatsenko:
> I agree that if users are provided with this option, workflow reproducibility can easily be (consciously or accidentally) broken. They would have to be clearly aware of the intended use of this feature (similar to the dangers of `update()`).

Agreed. We have seen the need for this solutions as well and with proper documentation, it makes sense. We would need to document and add test cases before including in a release.

datajoint/autopopulate.py
```diff
@@ -150,7 +151,10 @@ def handler(signum, frame):
                     call_count += 1
                     self.__class__._allow_insert = True
                     try:
-                        make(dict(key))
+                        if make_kwargs is not None:
+                            make(dict(key), **make_kwargs)
+                        else:
+                            make(dict(key))
```
dimitri-yatsenko, line 157:
```suggestion
                        make(dict(key), **(make_kwargs or {}))
```

dimitri-yatsenko:
@Hendriela Will you please make integrate the suggestions? I will then merge your PR.

A-Baji:
Sorry @Hendriela!, I did not mean to commit that to your fork.

guzman-raphael (changes requested):
@Hendriela @A-Baji Thanks for the contributions! :handshake: 

These updates look great now and the feature seems quite useful. Would you just update the changelog and fix the minor style test issue? You can place the release notes within the upcoming `0.13.3` release.

Unfortunately, we keep the changelog in 2 places right now located in the following paths:
- `./CHANGELOG.md`
- `./docs-parts/intro/Releases_lang1.rst`

dimitri-yatsenko (approved):

guzman-raphael (approved):
@Hendriela @A-Baji

:tada: Nice work! :tada: 
Great to see collaboration in the community.

Issue #927, opened 2021-05-27
`.descendants()` failed with master-part tables where the descendants only contain the part-table


ttngu207:
### Description
`table.descendants()` failed for the case where there are master-part tables downstream, where only the part-table is a descendant, but not the master.

### Reproducibility
Include:
- OS (WIN)
- Python 3.7
- MySQL Version 5.7
- MySQL Deployment Strategy (local-native)
- DataJoint Version 0.13.2

#### Reproducible case

Given the following schema design

```python
import datajoint as dj
schema = dj.Schema('test')

@schema
class A(dj.Manual):
    definition = """
    a: int
    """


@schema
class B(dj.Manual):
    definition = """
    -> A
    b: int
    """


@schema
class Master(dj.Manual):
    definition = """
    table_master: int
    """

    class Part(dj.Part):
        definition = """
        -> master
        -> B
        """
```

calling `A.descendants()` will result in the error below:

```
Traceback (most recent call last):
  File "C:\Users\thinh\Documents\TN-Vathes\MAP\map-ephys\venv\lib\site-packages\IPython\core\interactiveshell.py", line 3343, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-3-a68a3ab7b68c>", line 1, in <module>
    TableA.descendants()
  File "C:\Users\thinh\Documents\TN-Vathes\MAP\map-ephys\venv\lib\site-packages\datajoint\table.py", line 181, in descendants
    for node in self.connection.dependencies.descendants(self.full_table_name) if not node.isdigit()]
  File "C:\Users\thinh\Documents\TN-Vathes\MAP\map-ephys\venv\lib\site-packages\datajoint\dependencies.py", line 149, in descendants
    nx.algorithms.dag.topological_sort(nodes)))
  File "C:\Users\thinh\Documents\TN-Vathes\MAP\map-ephys\venv\lib\site-packages\datajoint\dependencies.py", line 30, in unite_master_parts
    raise DataJointError("Found a part table {name} without its master table.".format(name=name))
datajoint.errors.DataJointError: Found a part table `test`.`master_table__part_table` without its master table.
```

guzman-raphael:
@ttngu207 Thanks for the report and it is unfortunate to hear of your issue. Have you by chance confirmed if this is a regression issue or not? As in, does this work as expected in `0.12.x`?

ttngu207:
[...]

PR #985, opened 2022-01-19
bugfix in `unite_master_parts` - handles dj.Lookup tables

ttngu207:


dimitri-yatsenko:
Looks great. Just update CHANGELOG.mg,  docs-parts/intro/Releases_lang1.rst, and reference the bug from this PR. I will merge then.

ttngu207:
@dimitri-yatsenko there isn't an issue filed for this bug, I just ran into it and issued a PR to fix. Should I file a new issue?

dimitri-yatsenko:
no need for an issue. But do add a note in the CHANGELOG describing what was fixed.

dimitri-yatsenko:
I suggest a higher-level description in CHANGELOG:
> Bugfix in listing part tables of lookup tables.

dimitri-yatsenko (approved):

dimitri-yatsenko:
Does this solve #927?

ttngu207:
@dimitri-yatsenko, I don't believe so, it is likely a different underlying cause for #927, perhaps related to diagram not stepping into the part tables and search for ancestors of each of the part, recursively.
I could be off here, haven't dig deep into 927

PR #991, opened 2022-02-09
Fix error when altering Part table that uses "master" keyword

### LINKED ISSUE #936: alter support for part tables
### Description

Alter on part tables yields 'datajoint.errors.DataJointError: Foreign key reference master could not be resolved'

### Reproducibility

Via support ticket 151 ; reproduced on linux / dj 0.13.3 / py 3.8 using:

```
import datajoint as dj

schema = dj.schema('test_fooparts')

@schema
class Foo(dj.Manual):
    definition = '''
    foo_id: int
    '''

    class SubFooOne(dj.Part):
        definition = '''
        -> master
        subfoo_id: int
        '''
```
followed by:
```
>>> Foo.SubFooOne().alter()
  File "<stdin>", line 1, in <module>
  File "/src/djc/datajoint-python/datajoint/table.py", line 104, in alter
    sql, external_stores = alter(self.definition, old_definition, context)
  File "/src/djc/datajoint-python/datajoint/declare.py", line 360, in alter
    table_comment, primary_key, attribute_sql, foreign_key_sql, index_sql, external_stores = prepare_declare(
  File "/src/djc/datajoint-python/datajoint/declare.py", line 248, in prepare_declare
    compile_foreign_key(line, context, attributes,
  File "/src/djc/datajoint-python/datajoint/declare.py", line 143, in compile_foreign_key
    raise DataJointError('Foreign key reference %s could not be resolved' % result.ref_table)
datajoint.errors.DataJointError: Foreign key reference master could not be resolved
```

### Expected Behavior

Unclear if master/part support is included in alter yet (alter is experimental); filing for reference to track capability

rly:
Fix #936.

The core of the issue appears to be that the "master" keyword in the Part table definition does not match the name of a table in `context`. So in this PR, I update `alter` only for Part tables so that "master" refers to the master table in `context`. There may be a more elegant way to fix this issue, but this seems to do the trick.


rly:
This solution does not handle the case of a part table using the "master" keyword in a more complex way, though.
```python
@schema
class Probe(dj.Manual):
    definition = """
    id: int
    """

    class Shank(dj.Part):
        definition = """
        -> master
        """

    class Electrode(dj.Part):
        definition = """
        -> master.Shank
        """
```

datajoint/user_tables.py
```diff
@@ -186,3 +187,16 @@ def drop(self, force=False):
...
+        """
+        Alter the table definition from self.definition
+        """
+        # map "master" keyword to master table in context
+        if context is None:
+            frame = inspect.currentframe().f_back
+            context = dict(frame.f_globals, **frame.f_locals)
+            del frame
+        if self.master:
+            context['master'] = self.master
```
rly, line 201: If the same `context` is shared across tables (it's not clear to me if that is the case), then altering `context` in this way would be problematic. Then the "master" keyword could accidentally be applied to a non-Part table.

tests/test_alter.py
```diff
@@ -39,3 +39,26 @@ def test_alter():
...
+        definition1 = """
+        -> AlterMaster
+        """
+
+def test_alter_part_master_keyword():
+    original = schema.connection.query("SHOW CREATE TABLE " + AlterMaster.AlterPart.full_table_name).fetchone()[1]
+    AlterMaster.AlterPart.definition = AlterMaster.AlterPart.definition1
+    AlterMaster.AlterPart.alter(prompt=False)
+    altered = schema.connection.query("SHOW CREATE TABLE " + AlterMaster.AlterPart.full_table_name).fetchone()[1]
+    assert_equal(original, altered)
```
guzman-raphael, line 64: A good principle to guide you through fixes and new features is [test-driven development](https://en.wikipedia.org/wiki/Test-driven_development). We use this quite a bit throughout the DataJoint codebases.

Here I would say that your test violates this since it still passes even before a fix is introduced. We should think first of a breaking test case based on the reference issue and work backwards from there to find the solution.

Also, we typically keep the table declarations together near the top as opposed to weaving them between tests. Taking this into account and your [comment](https://github.com/datajoint/datajoint-python/pull/991#issuecomment-1033694920) of another edge-case, we can create one like so:

```suggestion
@schema
class Parent(dj.Manual):
    definition = """
    parent_id: int
    """

    class Child(dj.Part):
        definition = """
        -> Parent
        """
        definition_new = """
        -> master
        ---
        child_id=null: int
        """

    class Grandchild(dj.Part):
        definition = """
        -> master.Child
        """
        definition_new = """
        -> master.Child
        ---
        grandchild_id=null: int
        """
# --- other tests ---
def test_alter_part():
    # https://github.com/datajoint/datajoint-python/issues/936
    def verify_alter(table, attribute_sql):
        definition_original = schema.connection.query(
            f"SHOW CREATE TABLE {table.full_table_name}"
        ).fetchone()[1]
        table.definition = table.definition_new
        table.alter(prompt=False)
        definition_new = schema.connection.query(
            f"SHOW CREATE TABLE {table.full_table_name}"
        ).fetchone()[1]
        assert (
            definition_new.replace(f"{attribute_sql},\n  ", "") == definition_original
        )

    verify_alter(table=Parent.Child, attribute_sql="`child_id` int DEFAULT NULL")
    verify_alter(
        table=Parent.Grandchild, attribute_sql="`grandchild_id` int DEFAULT NULL"
    )
```

datajoint/user_tables.py
```diff
@@ -186,3 +187,16 @@ def drop(self, force=False):
...
+        super().alter(prompt, context)
```
guzman-raphael, line 202: Nice find! I agree this is probably the best place to address this type of fix. However, you are right that there is a more elegant way to solve this problem.

The primary issue with your approach here is the additional overhead in inspecting the memory again (we've previously loaded this in `context`). Additionally, we add several more lines and it still doesn't satisfy further indexing into `master`.

Consider the following which should help address the above.
```suggestion
    def alter(self, prompt=True, context=None):
        super().alter(prompt=prompt, context=context or self.declaration_context)
```

datajoint/user_tables.py
```diff
@@ -2,6 +2,7 @@
 Hosts the table tiers, user relations should be derived from.
 """
 
+import inspect
```
guzman-raphael, line 5:
```suggestion
```

docs-parts/intro/Releases_lang1.rst
```diff
@@ -15,6 +15,7 @@
 * Bugfix - Fix assertion error when performing a union into a join (#930) PR #967
 * Bugfix - Fix regression issue with `DISTINCT` clause and `GROUP_BY` (#914) PR #963
 * Update `~jobs.error_stack` from blob to mediumblob to allow error stacks >64kB in jobs (#984) PR #986
+* Bugfix - Fix error when altering a part table that uses the "master" keyword (#936) PR #991
```
guzman-raphael, line 18: Would you repull from `upstream` and fix any merge conflicts? Once you do, you will notice that this doc doesn't exist anymore and we only need to update this in one place (`CHANGELOG.md`).

guzman-raphael (changes requested):
@rly Thank you for contributing to the fix! :muscle:

We appreciate your patience with our review but we've finally opened up some bandwidth to work through our PR queue.

Suggesting an alternative approach.

datajoint/user_tables.py
rly: Thanks!

tests/test_alter.py
rly: Great, thanks! I made this change.

rly:
Thanks for taking a look at this PR. This issue comes up in the Frank Lab often.

Syntax, integration, and style checks pass. Thanks for the comprehensive developer documentation!

guzman-raphael (approved):
@rly Np. :beers: 

Most of all, thanks to contributors like you for your help! LGTM

PR #995, opened 2022-02-15
Fix issue with reading 32 bit blobs

A-Baji:


datajoint/blob.py
```diff
@@ -431,8 +433,16 @@ def read_zero_terminated_string(self):
...
         return data
 
-    def read_value(self, dtype='uint64', count=1):
-        data = np.frombuffer(self._blob, dtype=dtype, count=count, offset=self._pos)
+    def read_value(self, dtype=None, count=1):
+        if dtype is None:
+            dtype = 'uint32' if self.is_32_bit else 'uint64'
+        try:
+            data = np.frombuffer(self._blob, dtype=dtype, count=count, offset=self._pos)
+        except ValueError:
```
dimitri-yatsenko, line 441: This may not be a reliable mechanisms for determining the decoding. The serialization should be explicit with respect to how the encoding is performed.

guzman-raphael: @dimitri-yatsenko Yes, we discussed that but the issue is that including the precision into the serialization scheme (e.g. `dj064` or `mym32`) is an update that isn't backward compatible. It would work point forward but wouldn't help us for prior blobs which is precisely where there is likely to be an issue.

dimitri-yatsenko: The Python implementation is deterministic. The serialization does not depend on whether the platform is 32- or 64-bit. 

The problem is that the above solution will not catch all errors. Some cases will not raise the `ValueError`.  We should devise the solution to catch those cases.

dimitri-yatsenko: The real problem is the `mym` implementation depended on the platform. We should explicitly catch this case.

guzman-raphael: We are trying to find an approach with a 'fall-back' mechanism that can capture the vast majority of cases without introducing performance overhead for normal unpacking. This was chosen since one of our customers has an enormous amount of data that would need to be migrated otherwise.

dimitri-yatsenko: As far as I understand, for all numeric arrays, the datatype is fully determined. `mym` underspecified the integer length and it defaulted to the platform default. That was a problem with the matlab implementation only. Correct?  If so, we can have a more targeted fix.

dimitri-yatsenko:
We should not rely on the value error to determine the encoding of array dimensions. This issue is limited to a specific case of old blobs created by a version of mYm that datajoint itself never used.  Therefore, please make this flag explicit rather than automatic. When working with these outdated blobs, the user will need to set 
```python
dj.blob.use_32bit_dims = True
```

datajoint/blob.py
```diff
@@ -43,6 +43,7 @@
 }
 
 bypass_serialization = False  # runtime setting to bypass blob (en|de)code
+use_32bit_dims = False # runtime setting to read data as 32-bit
```
dimitri-yatsenko, line 46:
```suggestion
# runtime setting to read integers as 32 bits to read blobs created by the 32-bit 
# version of the mYm library for MATLAB 
use_32bit_dims = False 
```

A-Baji:
Is there a better way of wording the change log that you would prefer?

CHANGELOG.md
```diff
@@ -1,7 +1,7 @@
 ## Release notes
 
 ### 0.13.4 -- TBA
-* Bugfid - Fix error when fetching data that was inserted as 32-bit
+* Bugfix - Added a module flag for reading 32-bit data
```
dimitri-yatsenko:
```suggestion
* Bugfix - Allow reading blobs produced by legacy 32-bit compiled mYm library for matlab.
```

docs-parts/intro/Releases_lang1.rst
```diff
@@ -1,6 +1,6 @@
 0.13.4 -- TBA
 ----------------------
-* Bugfix - Fix error when fetching data that was inserted as 32-bit
+* Bugfix - Added a module flag for reading 32-bit data
```
dimitri-yatsenko:
```suggestion
* Bugfix - Allow reading blobs produced by legacy 32-bit compiled mYm library for matlab. PR #995
```

guzman-raphael (approved):
@A-Baji :clap: Fanstastic job on a tricky one! :clap:

dimitri-yatsenko (approved):

PR #998, opened 2022-03-02
Black formatting

A-Baji:


guzman-raphael (approved):
@A-Baji Beautiful! Looking quite slick now with `black`. :sunglasses:

PR #1002, opened 2022-03-22
Import ABC from collections.abc for Python 3.10 compatibility

tirkarthi:
Import ABC from `collections` was deprecated and removed in Python 3.10. Use `collections.abc` .


guzman-raphael:
@tirkarthi Thanks for the updates! Looks like this may be failing just b/c it is using old certs. Would you update it with the latest from `master`?

PR #1019, opened 2022-05-03
Optimize read_datetime

jverswijver:
Switching from floor division to string processing results in ~35% speedup in execution time for read_datetime after profiling with cProfile.

### DIFF
--- a/datajoint/blob.py
+++ b/datajoint/blob.py
@@ -487,17 +487,24 @@ class Blob:
     def read_datetime(self):
         """deserialize datetime.date, .time, or .datetime"""
         date, time = self.read_value("int32"), self.read_value("int64")
+        date_str = str(date)
+        time_str = str(time)
+
         date = (
-            datetime.date(year=date // 10000, month=(date // 100) % 100, day=date % 100)
+            datetime.date(
+                year=int(date_str[:4]) if date_str[:4] != "" else 0,
+                month=int(date_str[4:6]) if date_str[4:6] != "" else 0,
+                day=int(date_str[-2:] if date_str[-2:] != "" else 0),
+            )
             if date >= 0
             else None
         )
         time = (
             datetime.time(
-                hour=(time // 10000000000) % 100,
-                minute=(time // 100000000) % 100,
-                second=(time // 1000000) % 100,
-                microsecond=time % 1000000,
+                hour=int(time_str[-12:-10]) if time_str[-12:-10] != "" else 0,
+                minute=int(time_str[-10:-8]) if time_str[-10:-8] != "" else 0,
+                second=int(time_str[-8:-6]) if time_str[-8:-6] != "" else 0,
+                microsecond=int(time_str[6:12]) if time_str[6:12] != "" else 0,
             )
             if time >= 0
             else None


zitrosolrac (approved):
I reviewed the performance files using SnakeViz and they reflect the performance upgrade.

datajoint/blob.py
```diff
@@ -487,17 +487,24 @@ def pack_cell_array(self, array):
...
         date, time = self.read_value("int32"), self.read_value("int64")
+        date_str = str(date)
+        time_str = str(time)
+
         date = (
-            datetime.date(year=date // 10000, month=(date // 100) % 100, day=date % 100)
+            datetime.date(
+                year=int(date_str[:4]) if date_str[:4] != "" else 0,
+                month=int(date_str[4:6]) if date_str[4:6] != "" else 0,
+                day=int(date_str[-2:] if date_str[-2:] != "" else 0),
```
dimitri-yatsenko, line 497:
```suggestion
                year=int(date_str[:4]) or 0,
                month=int(date_str[4:6]) or 0,
                day=int(date_str[-2:] or 0),
```

datajoint/blob.py
```diff
@@ -487,17 +487,24 @@ def pack_cell_array(self, array):
...
         time = (
             datetime.time(
-                hour=(time // 10000000000) % 100,
-                minute=(time // 100000000) % 100,
-                second=(time // 1000000) % 100,
-                microsecond=time % 1000000,
+                hour=int(time_str[-12:-10]) if time_str[-12:-10] != "" else 0,
+                minute=int(time_str[-10:-8]) if time_str[-10:-8] != "" else 0,
+                second=int(time_str[-8:-6]) if time_str[-8:-6] != "" else 0,
+                microsecond=int(time_str[6:12]) if time_str[6:12] != "" else 0,
```
dimitri-yatsenko, line 507:
```suggestion
                hour=int(time_str[-12:-10] or 0),
                minute=int(time_str[-10:-8] or 0),
                second=int(time_str[-8:-6] or 0),
                microsecond=int(time_str[6:12] or 0),
```

dimitri-yatsenko (changes requested):

datajoint/blob.py
```diff
@@ -487,17 +487,24 @@ def pack_cell_array(self, array):
     def read_datetime(self):
         """deserialize datetime.date, .time, or .datetime"""
         date, time = self.read_value("int32"), self.read_value("int64")
+        date_str = str(date)
+        time_str = str(time)
+
         date = (
-            datetime.date(year=date // 10000, month=(date // 100) % 100, day=date % 100)
```
dimitri-yatsenko, line 491: Why the change? Here are some timing tests:
image

dimitri-yatsenko (changes requested):

datajoint/blob.py
```diff
@@ -487,17 +487,24 @@ def pack_cell_array(self, array):
...
+                year=int(date_str[:4]) if date_str[:4] != "" else 0,
+                month=int(date_str[4:6]) if date_str[4:6] != "" else 0,
+                day=int(date_str[-2:] if date_str[-2:] != "" else 0),
+            )
             if date >= 0
             else None
         )
         time = (
             datetime.time(
-                hour=(time // 10000000000) % 100,
```
dimitri-yatsenko, line 497: Why is this better? Here is the timing test:

image

dimitri-yatsenko (changes requested):

dimitri-yatsenko:
The string parsing takes longer according to `%%timeit` tests and is more verbose.

datajoint/blob.py
jverswijver: > The string parsing takes longer according to `%%timeit` tests and is more verbose.

The way that I found that this results in a speedup is I profiled the unpacking of a nparray of 100000 datetime objects and then I overloaded the read_datetime method and profiled unpack again. When looking at the cProfile results the string process method resulted in less total time spent in read_datetime. But it seems like when you profile it you get different results. 

Do you want to tag-up on this sometime? Also I have a python script that generates the cProfile performance profiling which you can then visualize using a python package called snakeviz, I can send you this script and we can look at it to see if there is some error in my logic.

I will attach pictures of the visualized performance profiles below.

before overload:
image
image

after:
image
image

specifically I looked at the decrease in total execution time as well as the `totime` column which represents total time spent in each method across all method calls. Please let me know what you think @dimitri-yatsenko

dimitri-yatsenko: I just don't see a compelling reason why the string processing would produce a speedup. I think the reason the blob deserialization is slow is because of python's need to loop through the numbers and calling `datetime.date` separately for each time and each date.

dimitri-yatsenko: A real speedup can potentially be produced by using `numpy.datetime64` type support.

dimitri-yatsenko: Here is a 500x improvement in decoding speed:
image

dimitri-yatsenko: `np.datetime64` did not exist when we made the original time serializer.

dimitri-yatsenko: For now, we can recommend the workaround for storing datetimes as `int64` as shown. We can add native support for `datetime64` data type, which would eliminate the need for converting into uint64 and back.

jverswijver:
superseded by #1036

PR #1025, opened 2022-05-17
Add option to disable checksums

jverswijver:


A-Baji (approved):

datajoint/fetch.py
```diff
@@ -32,7 +32,7 @@ def to_dicts(recarray):
         yield dict(zip(recarray.dtype.names, rec.tolist()))
 
 
-def _get(connection, attr, data, squeeze, download_path):
+def _get(connection, attr, data, squeeze, download_path, disable_checksum):
```
dimitri-yatsenko, line 35: Consider renaming thing into a positive statement to avoid double negatives in logic. For example, replace `disiable_checksums=False` into `verify_file_checksum=True`

datajoint/fetch.py
```diff
@@ -149,7 +149,8 @@ def __call__(
         format=None,
         as_dict=None,
         squeeze=False,
-        download_path="."
+        download_path=".",
+        disable_checksum=False
```
dimitri-yatsenko, line 153: update the docstring

dimitri-yatsenko (changes requested):
work is needed

datajoint/fetch.py
```diff
@@ -170,6 +171,18 @@ def __call__(
         :param download_path: for fetches that download data, e.g. attachments
         :return: the contents of the relation in the form of a structured numpy.array or a dict list
         """
+        if disable_checksum and config["safemode"]:
```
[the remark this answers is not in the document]
guzman-raphael: Do you mean raise a warning on fetches that they explicitly disable the checksum? That makes sense to me. The logs might just get a bit verbose in the output... This would eliminate the need for interacting with `safemode` in this option.

Unless you mean raise a warning like: `WARNING: Checksum evaluation enabled for fetches from external storage`? If so, I think that would defeat the purpose as users would be motivated to disable the checksum just to eliminate the warning.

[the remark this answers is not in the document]
dimitri-yatsenko: Overall, I think that setting this in every fetch command is not the right approach. This should be set globally. Warnings can be generated not at the time of issuing the fetch but at the time when a blob completes downloading without a checksum check.

guzman-raphael:
@dimitri-yatsenko What if we added a `dj.config` setting that takes a list of tables to disable checksum verification on? That way the code can remain the same though the config will persist but users won't just naively apply it everywhere. We can start ignoring it after a certain release (with a warning initially). That way we have a plan to enforce (again) the checksums once it has been properly optimized.

The main reason we had included this at the scope of each `fetch` is so the user will do a value assessment to determine if a particular `fetch` could really use the benefit of disabling `checksums` as opposed to a blanket application everywhere. This could still protect many other external fetches if the data is <10GiB. We will eventually upgrade the efficiency of checksums; it will just take some time to get it right so this is a stopgap since urgent improvement is necessary.

dimitri-yatsenko:
I agree that a global `dj.config` setting separate from `safemode` would be appropriate for checksum verification on fetch.

guzman-raphael:
How about to summarize:
- We'll introduce a new `dj.config` key (`disable_fetch_file_verification`). It's value is a list of full table names i.e. ``[`schema`.`table`]``
- This will only affect `filepath@store` attributes. This is easier since `attachment` and `filepath` checksums are handled in 2 different places (`fetch.py:_get`, `external.py:download_filepath`) and it will get messy. Trying not to refactor since we are thinking of a new solution to supersede external storage anyway.
- We'll need to add a new kwarg to `download_filepath` so that it knows the table name for the fetch that is triggering the download and we can handle it all in one place. Pass the table name associated with the file download to `download_filepath` so that it can perform the validation against the config.
- If this key exists and the table associated with the file download is in the list, then disable the checksum validation and throw a warning during the download step.

@dimitri-yatsenko Thumbs up if you agree and let me know if you want to catch up on this.

dimitri-yatsenko:
@guzman-raphael 
1. Perhaps the setting should be global rather than doing this for each file. If anything 
2. Can we use a positive formulation, e.g. `disable_fetch_file_verification` -> `verify_file_checksum_on_fetch`.  Then statements with double negatives such as `if not disable_fetch_file_verification:` become a positive `if verify_file_checksum_on_fetch`.
3. I agree to only doing this for filepath. For attach, different files can have the same name, so checksum becomes a file identification mechanism, not just integrity check.

guzman-raphael:
@dimitri-yatsenko 

1. I disagree with this only because I would still discourage usage of disabling checksums. I'm normally in favor of simplicity but here it would cost us disabling this integrity check everywhere. There are probably only a handful of tables tracking such large raw data, right?

2. This would be nice but don't think it would be compatible with `1` as it would be cumbersome to explicitly list all tables. The check could be `if table in disable_fetch_file_verification`.

3. 👍

guzman-raphael:
@dimitri-yatsenko and I met yesterday and discussed this. We converged on a simpler approach that could address the issue more directly.

To summarize:
- Add a new global config to `dj.config` that will specify a file size limit where checksums will only be evaluated for fetch of `filepath` files less than that size.
- Perhaps it could be called `filepath_checksum_size_limit`. It could take its value as an integer representing the bytes.
- `None` (default) would indicate no limit i.e. perform the checksum for all filepath fetches.
- `0` would be discouraged but could basically disable fetch checksums for all filepath.
- Throw a warning if a checksum is skipped due to this setting.
- We'd recommend that for users interested in using this stopgap (until further optimization is developed) to prevent verification, they should do this at their own risk and ideally tune it to a reasonable size that is affecting performance for their case.

datajoint/external.py
```diff
@@ -310,12 +310,38 @@ def download_filepath(self, filepath_hash):
...
+            elif (
+                Path(local_filepath).is_file()
+                and config["filepath_checksum_size_limit"] is not None
+            ):
+                file_exists = (
+                    True
+                    if Path(local_filepath).stat().st_size  # size in bytes
+                    > config["filepath_checksum_size_limit"]
+                    else (uuid_from_file(local_filepath) == contents_hash)
+                )
```
dimitri-yatsenko, line 333:
```suggestion
                file_exists = (Path(local_filepath).stat().st_size  > config["filepath_checksum_size_limit"]
                    or uuid_from_file(local_filepath) == contents_hash
                )
```

datajoint/external.py
```diff
@@ -310,12 +310,38 @@ def download_filepath(self, filepath_hash):
             )
             external_path = self._make_external_filepath(relative_filepath)
             local_filepath = Path(self.spec["stage"]).absolute() / relative_filepath
-            file_exists = (
+            if (
+                config["filepath_checksum_size_limit"] is not None
```
guzman-raphael, line 314: You can just check the truthy value of `config["filepath_checksum_size_limit"]`

datajoint/external.py
```diff
@@ -310,12 +310,38 @@ def download_filepath(self, filepath_hash):
...
+                file_exists = False
+
             if not file_exists:
                 self._download_file(external_path, local_filepath)
+                if (
+                    config["filepath_checksum_size_limit"] is not None
+                    and Path(local_filepath).stat().st_size
+                    > config["filepath_checksum_size_limit"]
+                ):
+                    return str(local_filepath), contents_hash
```
guzman-raphael, line 344: We can eliminate this `return` if we include the checksum into this `if` condition.

guzman-raphael (changes requested):
@jverswijver :sparkles: Great work! This is coming together nicely. :sparkles: 

Suggesting minor points to simplify some sections.

Additionally, could you:
- Add changelog/releaselog
- Add a test

datajoint/external.py
```diff
@@ -310,47 +310,39 @@ def download_filepath(self, filepath_hash):
...
-            ):
-                file_exists = (
-                    True
-                    if Path(local_filepath).stat().st_size  # size in bytes
-                    > config["filepath_checksum_size_limit"]
-                    else (uuid_from_file(local_filepath) == contents_hash)
+
+            file_exists = Path(local_filepath).is_file() and (
+                (
+                    config["filepath_checksum_size_limit"]  # check if None
```
dimitri-yatsenko, line 316:
```suggestion
                    config.get("filepath_checksum_size_limit") is not None
```

datajoint/external.py
```diff
@@ -310,47 +310,39 @@ def download_filepath(self, filepath_hash):
...
-                        "'{file}' downloaded but did not pass checksum'".format(
-                            file=local_filepath
+                    checksum = uuid_from_file(local_filepath)
+                    if (
+                        checksum != contents_hash
+                    ):  # this should never happen without outside interference
+                        raise DataJointError(
+                            "'{file}' downloaded but did not pass checksum'".format(
+                                file=local_filepath
+                            )
```
dimitri-yatsenko, line 336:
```suggestion
                    if checksum != contents_hash:
                        # this should never happen without outside interference or data corruption
                        raise DataJointError(
                            "'{file}' downloaded but did not pass checksum'".format(
                                file=local_filepath
                            )
```

datajoint/external.py
```diff
@@ -307,6 +307,12 @@ def download_filepath(self, filepath_hash):
         :param filepath_hash: The hash (UUID) of the relative_path
         :return: hash (UUID) of the contents of the downloaded file or Nones
         """
+
+        def _verify_contents(local_filepath):
+            return config.get("filepath_checksum_size_limit") is None or Path(
+                local_filepath
+            ).stat().st_size < config.get("filepath_checksum_size_limit")
```
dimitri-yatsenko, line 311:
```suggestion
        def _need_checksum(local_filepath):
            limit = config.get("filepath_checksum_size_limit")
            return limit is None or Path(local_filepath).stat().st_size < limit
```

jverswijver: I have made the change.

datajoint/external.py
```diff
@@ -310,12 +310,38 @@ def download_filepath(self, filepath_hash):
             )
             external_path = self._make_external_filepath(relative_filepath)
             local_filepath = Path(self.spec["stage"]).absolute() / relative_filepath
-            file_exists = (
+            if (
+                config["filepath_checksum_size_limit"] is not None
+                and Path(local_filepath).stat().st_size
+                > config["filepath_checksum_size_limit"]
+            ):
+                print(f"WARNING SKIPPING CHECKSUM FOR {filepath_hash}\n", flush=True)
```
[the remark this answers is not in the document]
guzman-raphael: Discussed this with @jverswijver individually and seems like we'll implement logging in one-shot in a follow-up PR. :+1:

guzman-raphael (approved):
@jverswijver :clap: Way to go on this one! :clap: 

Required more coordination than expected by I really like how we ended up handling this!

datajoint/external.py
```diff
@@ -304,27 +304,37 @@ def download_filepath(self, filepath_hash):
...
-                ):  # this should never happen without outside interference
-                    raise DataJointError(
-                        "'{file}' downloaded but did not pass checksum'".format(
-                            file=local_filepath
+                if _need_checksum(local_filepath):
+                    checksum = uuid_from_file(local_filepath)
+                    if checksum != contents_hash:
+                        # this should never happen without outside interference
+                        raise DataJointError(
+                            "'{file}' downloaded but did not pass checksum'".format(
```
dimitri-yatsenko, line 331: considering that we are Python 3.6+, we can switch to f-strings

datajoint/external.py
```diff
@@ -304,27 +304,37 @@ def download_filepath(self, filepath_hash):
...
+                                file=local_filepath
+                            )
```
dimitri-yatsenko, line 333:
```suggestion
                    uuid_from_file(local_filepath) != contents_hash:
                        # this should never happen without outside interference
                        raise DataJointError(f"'{local_filepath}' downloaded but did not pass checksum'")
```

tests/test_filepath.py
```diff
@@ -169,6 +170,9 @@ def test_filepath_class(table=Filepath(), store="repo"):
     # delete from external table
     table.external[store].delete(delete_external_files=True)
     dj.errors._switch_filepath_types(False)
+    dj.config["filepath_checksum_size_limit"] = None
+    # Todo: assert here that the warning from external.download_file has been logged
+    # Waiting on DJ Logging implementation
```
dimitri-yatsenko, line 175: do not leave TODOs in the code. All issues should be tracked through the github issue manager.

jverswijver: I have removed it and filed an issue. https://github.com/datajoint/datajoint-python/issues/1030

datajoint/external.py
jverswijver: I have changed this `.format` to an fstring. I also changed a few more in external.py to fstrings but not all of them, some of them are sql queries which were too difficult to change in a short pass. If you want I can change those as well but it will take some time.

jverswijver:
Now relies on #1031 to be merged first. This is due to needing to use the new logger to test if a warning was thrown when disabling checksums.

datajoint/external.py
```diff
@@ -326,7 +329,9 @@ def _need_checksum(local_filepath):
                             f"'{local_filepath}' downloaded but did not pass checksum'"
                         )
             if not _need_checksum(local_filepath):
-                print(f"WARNING SKIPPED CHECKSUM FOR FILE WITH HASH: {contents_hash}")
+                logger.warning(
+                    f"WARNING SKIPPED CHECKSUM FOR FILE WITH HASH: {contents_hash}"
```
dimitri-yatsenko, line 333: why all caps?

jverswijver: No reason, do you think it should be normal case?

datajoint/external.py
```diff
@@ -330,7 +330,7 @@ def _need_checksum(local_filepath):
                         )
             if not _need_checksum(local_filepath):
                 logger.warning(
-                    f"WARNING SKIPPED CHECKSUM FOR FILE WITH HASH: {contents_hash}"
+                    f"Warning skipped checksum for file with hash: {contents_hash}"
```
dimitri-yatsenko, line 333: would it be more valuable to log the filepath than the checksum?

jverswijver: It might be best to log both, that way if they have any problems they can manually verify the checksum.

ttngu207:
```suggestion
                    f"Skipped checksum for file with hash: {contents_hash}"
```
Already at the [WARNING] level with the logger.

ttngu207: We should at least check for matching file size if we're not checking for checksum. Expected file size can be fetched from external table

datajoint/external.py
```diff
@@ -304,26 +303,39 @@ def download_filepath(self, filepath_hash):
...
-                    checksum != contents_hash
-                ):  # this should never happen without outside interference
-                    raise DataJointError(
-                        "'{file}' downloaded but did not pass checksum'".format(
-                            file=local_filepath
+                if _need_checksum(local_filepath):
+                    if uuid_from_file(local_filepath) != contents_hash:
+                        # this should never happen without outside interference
+                        raise DataJointError(
+                            f"'{local_filepath}' downloaded but did not pass checksum'"
```
dimitri-yatsenko, line 329:
```suggestion
                            f"'{local_filepath}' downloaded but did not pass checksum"
```

datajoint/external.py
```diff
@@ -304,26 +303,39 @@ def download_filepath(self, filepath_hash):
...
                         )
+            if not _need_checksum(local_filepath):
+                logger.warning(
+                    f"Skipped checksum for file with hash: {contents_hash}, and path: {local_filepath}"
+                )
+                if size != Path(local_filepath).stat().st_size:
+                    # this should never happen without outside interference
+                    raise DataJointError(
+                        f"'{local_filepath}' downloaded but size is not the same (skipped checksum due to config)'"
```
dimitri-yatsenko, line 338:
```suggestion
                        f"'{local_filepath}' downloaded but size did not match."
```

datajoint/plugin.py
```diff
@@ -15,10 +18,10 @@ def _update_error_stack(plugin_name):
         signature = plugin_meta.get_metadata("{}.sig".format(plugin_name))
         pubkey_path = str(Path(base_meta.egg_info, "{}.pub".format(base_name)))
         verify(pubkey_path=pubkey_path, data=data, signature=signature)
-        print("DataJoint verified plugin `{}` detected.".format(plugin_name))
+        logger.info("DataJoint verified plugin `{}` detected.".format(plugin_name))
         return True
     except (FileNotFoundError, InvalidSignature):
-        print("Unverified plugin `{}` detected.".format(plugin_name))
+        logger.warning("Unverified plugin `{}` detected.".format(plugin_name))
```
dimitri-yatsenko, line 24: Let's switch to f-strings since we no longer support Python 3.6.
```suggestion
        logger.warning(f"Unverified plugin `{plugin_name}` detected.")
```

jverswijver: sounds good, this is due to this PR relying on PR #1031, I will change it over in that PR.

guzman-raphael (approved):
@jverswijver :clap: Nicely done! :clap:

dimitri-yatsenko (approved):

PR #1026, opened 2022-05-18
Multi processing bug fix

### LINKED ISSUE #1013: ValueError thrown when multi-processing runs into previously computed tables
### Description
A workflow I am working on involves running populates on tables which in some cases have already been populated. It should be skipping these tables, but instead throws `ValueError: Number of processes must be at least 1`

### Reproducibility
Include:
- OS: Linux (Ubuntu)
- Image: datajoint/djbase:py3.9-debian
- Python Version: 3.9
- MySQL Version: 8.0.23
- MySQL Deployment Strategy: remote
- DataJoint Version
- Minimum number of steps to reliably reproduce the issue: Launch Neurophotonics ec2 intance
- Complete error stack as a result of evaluating the above steps:
```['podman', '--version', '']
using podman version: 3.4.2
podman logs docker_standard_worker_1
No protocol specified
Connecting ernaldis@rds.datajoint.io:3306
  File "/main/scripts/pop_fields.py", line 4, in <module>
    DField.populate(reserve_jobs=True, processes = 1024, display_progress=False)
  File "/opt/conda/lib/python3.9/site-packages/datajoint/autopopulate.py", line 236, in populate
    with mp.Pool(
  File "/opt/conda/lib/python3.9/multiprocessing/context.py", line 119, in Pool
    return Pool(processes, initializer, initargs, maxtasksperchild,
  File "/opt/conda/lib/python3.9/multiprocessing/pool.py", line 205, in __init__
    raise ValueError("Number of processes must be at least 1")
ValueError: Number of processes must be at least 1
exit code: 0
```

### Expected Behavior
This should have seen the relevant tables were already populated, done nothing to the tables, and moved on.

### Additional Research and Context
Suspected problem:
Since the script being run specifies multiple processes (1024 processes to be precise), and there are no keys left to be computed, processes gets assigned to 0 in autopopulate.py, [line 216](https://github.com/datajoint/datajoint-python/blob/56ced5dabe84687ce33f14c02889fe546c9bbc80/datajoint/autopopulate.py#L216). The processes variable is then used in line [236-237](https://github.com/datajoint/datajoint-python/blob/56ced5dabe84687ce33f14c02889fe546c9bbc80/datajoint/autopopulate.py#L237) under the assumption that processes > 1.

Ernaldis:
This PR proposes a fix to the bug described in #1013. In brief, when a user tries to populate a table which has already been populated while also setting processes to be greater than 1, the population should be skipped, just like it is when processes is equal to 1. Instead, a value error is thrown. 

This PR also alters the return value to also return the number of keys processed. This can be useful in automating workflows.

Closes #1013


datajoint/autopopulate.py
```diff
@@ -252,8 +256,7 @@ def handler(signum, frame):
         if reserve_jobs:
             signal.signal(signal.SIGTERM, old_handler)
 
-        if suppress_errors:
-            return error_list
+        return error_list, nkeys
```
dimitri-yatsenko, line 259: why does `populate` need to return `nkeys` now?

datajoint/autopopulate.py
```diff
@@ -222,7 +222,11 @@ def handler(signum, frame):
             make_kwargs=make_kwargs,
         )
 
-        if processes == 1:
+        if processes < 0:
+            raise Exception("processes must not be negative")
+        elif processes == 0:
+            return error_list, nkeys
```
dimitri-yatsenko, line 228: here `error_list` is empty.  Why does this need to be return? What's the use case for `processes == 0`?

datajoint/autopopulate.py
Ernaldis: This can be useful in automating workflows, since it allows us to determine how many keys were found to be processed.
For example, this features is used in the [neurophotonics](https://github.com/datajoint-company/neurophotonics) project in check_db. This allows the EC2 instance to continue working until the user stops inserting keys into the database, for those times that keys are inserted after the instance started, but before the initial workload is finished.

datajoint/autopopulate.py
```diff
@@ -222,7 +222,11 @@ def handler(signum, frame):
             make_kwargs=make_kwargs,
         )
 
-        if processes == 1:
+        if processes < 0:
+            raise Exception("processes must not be negative")
```
dimitri-yatsenko, line 226: Root `Exception` should be avoided. A more specific exception should be used. Why is the check for negativity necessary?

Ernaldis: I will add a more specific exception here. The reason I included this to begin with is as a simple input validation. If a user tries to set processes to a negative number, something has gone wrong, and raising an exception should allow them to handle it.

Ernaldis: I specified the exception to be a value error.

datajoint/autopopulate.py
dimitri-yatsenko: I think if `nkeys <= 0`, it should just quietly exit. That's consistent with previous behavior.

datajoint/autopopulate.py
dimitri-yatsenko: Current users expect `populate` to return the error list. This change may break existing code. Changing the API should be done carefully.

Ernaldis: That is a valid concern, but I think it is also useful to be able to tell how many keys were processed.

I notice that there is nothing returned if `suppress_errors == False`, the default value. What if the return was left as just `error_list` in the case `suppress_errors == True`, and returned nkeys only if `suppress_erros == False`?

dimitri-yatsenko (approved):

datajoint/autopopulate.py
```diff
@@ -252,7 +253,7 @@ def handler(signum, frame):
         if reserve_jobs:
             signal.signal(signal.SIGTERM, old_handler)
 
-        if suppress_errors:
+        if not suppress_errors:
```
dimitri-yatsenko, line 256: oops. 
```suggestion
        if suppress_errors:
```

CHANGELOG.md
```diff
@@ -1,7 +1,8 @@
 ## Release notes
 
-### 0.13.5 -- May 13, 2022
+### 0.13.5 -- TBD
```
dimitri-yatsenko, line 3:
```suggestion
### 0.13.6 -- TBD
```

jverswijver (approved):

datajoint/autopopulate.py
```diff
@@ -173,8 +173,7 @@ def populate(
         :param limit: if not None, check at most this many keys
         :param max_calls: if not None, populate at most this many keys
         :param display_progress: if True, report progress_bar
-        :param processes: number of processes to use. When set to a large number, then
-            uses as many as CPU cores
+        :param processes: number of processes to use. Set to None to use all cores
```
guzman-raphael, line 176: This is a great idea to improve how this can be specified but doesn't seem to be true based on what is here. 

If you do in fact set `process=None` the following would happen:
- [216]: Should throw an error like: `TypeError: '>' not supported between instances of 'NoneType' and 'int'`

You could probably just replace that particular if-block with something like this:

`min(nkeys, *([processes, mp.cpu_count()] if processes else [mp.cpu_count()]))`

OR

`min(*(_ for _ in (processes, nkeys, mp.cpu_count()) if _))`

These approaches would leverage argument expansion to our benefit to expand only the relevant values into the `min` function.

guzman-raphael (changes requested):
@Ernaldis Thanks for patching this! :handshake:

Only have a small issue with the updated documentation since it doesn't seem correct. Although, should be straightforward to include.

guzman-raphael (changes requested):
@Ernaldis  This will still raise a `TypeError`. If you update this [test](https://github.com/datajoint/datajoint-python/blob/e2aa1ad388887b29b361ca12052095caffa1a3e1/tests/test_autopopulate.py#L66) with `processes=None` you will see what I am talking about. That test is meant for testing multiprocessing anyway. Probably you can just get rid of the [if](https://github.com/Ernaldis/datajoint-python/blob/91c8db4d20b6279c994b262e1d23a5c4527cc130/datajoint/autopopulate.py#L216) since it seems unnecessary. The default is already 1 which should always be the min.

guzman-raphael (approved):

datajoint/autopopulate.py
```diff
@@ -211,9 +210,10 @@ def handler(signum, frame):
 
         keys = keys[:max_calls]
         nkeys = len(keys)
+        if not nkeys:
+            return
 
-        if processes > 1:
-            processes = min(processes, nkeys, mp.cpu_count())
+        processes = min(*(_ for _ in (processes, nkeys, mp.cpu_count()) if _))
```
dimitri-yatsenko, line 216: With this, `processes=0` will have the same effect as `processes=None`.

PR #1029, opened 2022-05-27
add more logging for `populate()`

ttngu207:


datajoint/autopopulate.py
```diff
@@ -275,7 +275,7 @@ def _populate1(
                 if jobs is not None:
                     jobs.complete(self.target.table_name, self._job_key(key))
             else:
-                logger.info("Populating: " + str(key))
+                logger.info("Start populating TABLE: {} - KEY: {}".format(self.target.table_name, key))
```
dimitri-yatsenko, line 278:
```suggestion
                logger.info(f"Populating {self.target.table_name}: {key}")
```

datajoint/autopopulate.py
```diff
@@ -288,6 +288,7 @@ def _populate1(
                         exception=error.__class__.__name__,
                         msg=": " + str(error) if str(error) else "",
                     )
+                    logger.info("Error in populating TABLE: {} - KEY: {}\n\t Error Message: {}".format(self.target.table_name, key, error_message))
```
dimitri-yatsenko, line 291:
```suggestion
                    logger.info(f"Error populating {self.target.table_name}: {key} - {error_message}")
```

datajoint/autopopulate.py
```diff
@@ -303,6 +304,7 @@ def _populate1(
                         return key, error if return_exception_objects else error_message
                 else:
                     self.connection.commit_transaction()
+                    logger.info("Successful in populating TABLE: {} - KEY: {}".format(self.target.table_name, key))
```
dimitri-yatsenko, line 307:
```suggestion
                    logger.info(f"Success populating {self.target.table_name}: {key}")
```

dimitri-yatsenko (changes requested):

datajoint/autopopulate.py
dimitri-yatsenko: Do we really want to log so much?

ttngu207: This is to facilitate the "notification" mechanism, e.g. having clear and distinct log messages so that it can be parsed easily by some custom loghandler and then fed into some notification scheme (e.g. send emails, slack messages, etc.)

Hence I purposely have the log message being a bit wordy but hopefully distinctive - `Start populating TABLE`, `Error in populating TABLE`, etc. - so that it is easy to parse the log later. 

`Making...` or `Error making...` may just be too generic.

guzman-raphael:
These logs should be `debug` so that that the default behavior is not too verbose. Additionally, agree with @dimitri-yatsenko's point on making them a bit more concise. However, to help with parsing/automation we can use the `making` keyword to help us indicate these events; it is currently not being used anywhere else. So in summary:

```
Making {self.target.table_name}: {key}
Error making {self.target.table_name}: {key} - {error_message}
Success making {self.target.table_name}: {key}
```

Also, this will be superseded in #1031

PR #1031, opened 2022-06-02
Add centralized logger.

jverswijver:
Adds a package level logger for the datajoint package.

### Should accomplish the following:
- [x] bring all the loggers under the same logger, in the current DJ we make loggers per module.
- [x] convert all warnings to logs with level WARNING 
- [x] update any tests
- [x] convert any applicable print statements to logs
~~- [ ] Evaluate log DB table to see if logging needs to be added~~

### Additional functionality requested by data science team:
~~- [ ] log warnings when user calls populate on empty `key_source`~~
~~- [ ] debug level logging for any insert/delete~~
- [X] example for adding handler to the logger so they can send log output to sources other than stdout

### Example for capturing the log output but maintaining dj logging
See the converted test `test_limit_warning` in the PR.
Basically you add a handler to the logger that sends all logs to a StringIO object and then retrieve the string from the object, then you remove the handler.

### DIFF
--- a/datajoint/autopopulate.py
+++ b/datajoint/autopopulate.py
@@ -275,7 +275,7 @@ class AutoPopulate:
                 if jobs is not None:
                     jobs.complete(self.target.table_name, self._job_key(key))
             else:
-                logger.debug(f"Making {key} -> {self.target.table_name} ")
+                logger.debug(f"Making {key} -> {self.target.full_table_name}")
                 self.__class__._allow_insert = True
                 try:
                     make(dict(key), **(make_kwargs or {}))
@@ -289,7 +289,7 @@ class AutoPopulate:
                         msg=": " + str(error) if str(error) else "",
                     )
                     logger.debug(
-                        f"Error making {key} -> {self.target.table_name} - {error_message}"
+                        f"Error making {key} -> {self.target.full_table_name} - {error_message}"
                     )
                     if jobs is not None:
                         # show error name and error message (if any)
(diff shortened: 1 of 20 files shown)


ttngu207:
Suggesting the logging related to transaction to be set at DEBUG level
https://github.com/datajoint/datajoint-python/blob/0ff34f283c8c54c8510ee2e0b65d134fb8321efd/datajoint/connection.py#L383
https://github.com/datajoint/datajoint-python/blob/0ff34f283c8c54c8510ee2e0b65d134fb8321efd/datajoint/connection.py#L391
https://github.com/datajoint/datajoint-python/blob/0ff34f283c8c54c8510ee2e0b65d134fb8321efd/datajoint/connection.py#L400

datajoint/logging.py
```diff
@@ -0,0 +1,57 @@
...
+        )
+    else:
+        logger.error(f"Uncaught exception: {exc_value}")
+
+
+sys.excepthook = excepthook
+
+
+# https://github.com/tqdm/tqdm/issues/313#issuecomment-267959111
+class TqdmToLogger(io.StringIO):
```
dimitri-yatsenko, line 38: Where is this used? Does TQDM need to output anything to logger?

guzman-raphael: Agree that we actually don't need this. We don't need to manage `tqdm`'s logs.

guzman-raphael (changes requested):
@jverswijver Great work! :muscle: 

Mostly ready just offering further improvements.

dimitri-yatsenko (approved):

guzman-raphael (approved):
@jverswijver Nice work! :handshake:

datajoint/autopopulate.py
```diff
@@ -288,6 +288,9 @@ def _populate1(
                         exception=error.__class__.__name__,
                         msg=": " + str(error) if str(error) else "",
                     )
+                    logger.debug(
+                        f"Error making {key} -> {self.target.table_name} - {error_message}"
```
dimitri-yatsenko, line 292: I think `table_name` does not include the schema name whereas `full_table_name` does. 
```suggestion
                        f"Error making {key} -> {self.target.full_table_name} - {error_message}"
```

PR #1038, opened 2022-06-26
fix: don't show preview if database connection not activated

iamamutt:
currently, the only way to not throw an error when printing class methods or attributes for an inactive UserTable is to set config to "debug"

### DIFF
--- a/datajoint/expression.py
+++ b/datajoint/expression.py
@@ -1,21 +1,22 @@
-from itertools import count
-import logging
-import inspect
 import copy
+import inspect
+import logging
 import re
-from .settings import config
-from .errors import DataJointError
-from .fetch import Fetch, Fetch1
-from .preview import preview, repr_html
+from itertools import count
+
 from .condition import (
     AndList,
     Not,
-    make_condition,
+    PromiscuousOperand,
     assert_join_compatibility,
     extract_column_names,
-    PromiscuousOperand,
+    make_condition,
 )
 from .declare import CONSTANT_LITERALS
+from .errors import DataJointError
+from .fetch import Fetch, Fetch1
+from .preview import preview, repr_html
+from .settings import config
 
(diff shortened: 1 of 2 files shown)


datajoint/expression.py
```diff
@@ -639,6 +640,7 @@ def __repr__(self):
         return (
             super().__repr__()
             if config["loglevel"].lower() == "debug"
+            or getattr(self, "database", None) is None
```
dimitri-yatsenko, line 643: perhaps this is not the right place for this check. Derived query expressions do not have `.database`.  Only base expressions do. Perhaps you can move this logic into `Table` or `UserTable`.

dimitri-yatsenko (changes requested):

datajoint/expression.py
dimitri-yatsenko: The fix above will turn off display for many derived expressions such as joins for example since the Expression object resulting from a join will not have a `table` property.

iamamutt: What if instead of `.database` we check if `.heading` exists? `.heading` will always be `None` if a table is not yet activated, correct? `preview` will always require a `Heading` class.
 
https://github.com/datajoint/datajoint-python/blob/07c5553e403a3fdf13a0675a81b27ab128068aa4/datajoint/preview.py#L7

iamamutt: @dimitri-yatsenko If the error is intentional (e.g., trying to instantiate a non-activated `UserTable`), maybe there should be a specific error thrown in `preview` that describes the activation issue instead of errors like: `AttributeError: 'NoneType' object has no attribute 'non_blobs'`

dimitri-yatsenko: I agree we need to properly handle this case. Raising an informative error sounds reasonable. We just need to test this behavior for different types of queries, including derived.

iamamutt:
closing this to create an issue

PR #1043, opened 2022-08-11
add mechanism to check for processing progress at schema level

ttngu207:
Add new method `progress` for the schema object, returning the populate progress for all imported/computed tables for a given schema (also accounts for `reserved`, `error`, `ignore` jobs)


guzman-raphael:
@ttngu207 Perhaps we should consider utilizing populate's [progress](https://github.com/datajoint/datajoint-python/blob/8f9e81c6a1cae5bfc2c9cd5482b2dbeab89127e8/datajoint/autopopulate.py#L317)? You should be able to call this to determine the table populate status.

ttngu207:
@guzman-raphael , the problem with `table.progress()` is that it is not taking into account entries in the jobs table (e.g. `reserved`, `error`, `ignore`).

guzman-raphael:
Right, just thinking how we can report progress consistently at the table level and the schema level.

ttngu207:
Yeah, ideally we'd fix the `.progress()` at the table level to account for `reserve`, `error` and `ignore` jobs, but that would break backward compatibility.

guzman-raphael:
We can introduce an option to return as `pandas` too but have it default to the current behavior.

ttngu207:
You mean for the new `schema.progress()` to return a query expression instead of pandas dataframe?

dimitri-yatsenko (approved):

dimitri-yatsenko:
@ttngu207 Please resolve issues

ttngu207:
@dimitri-yatsenko , this `schema.progress()` feature is convenient but NOT essential, if it makes sense to close this PR without merging, I'm fine with that

dimitri-yatsenko:
Per discussion with @ttngu207 , we are closing this PR. This feature is not critical.

PR #1044, opened 2022-08-17
Update logger

jverswijver:
small changes to logging behavior:
- add check to update the logger when the `dj.config` is updated
- make sure that environment variables are given priority over the config
- should work with context manager now

### DIFF
--- a/datajoint/settings.py
+++ b/datajoint/settings.py
@@ -249,7 +249,7 @@ class Config(collections.abc.MutableMapping):
             else:
                 raise DataJointError("Validator for {0:s} did not pass".format(key))
             valid_logging_levels = ["DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"]
-            if "loglevel" in key and key in valid_logging_levels:
+            if "loglevel" in key and value in valid_logging_levels:
                 logger.setLevel(value)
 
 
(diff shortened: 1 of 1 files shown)


guzman-raphael:
@jverswijver From our prior discussion, the largest issue I see here is that if we are properly addition `loglevel` to be used in the config, we should verify the case where a user has a `dj_local_conf.json` and first imports `datajoint`. Their log level should be appropriately set.

jverswijver:
Yes, we should try to look for the config and if it is there we should load the log level for the user when we create the logger. Except of they have the log level set in their env which should take priority.

datajoint/settings.py
```diff
@@ -248,6 +248,9 @@ def __setitem__(self, key, value):
                 self._conf[key] = value
             else:
                 raise DataJointError("Validator for {0:s} did not pass".format(key))
+            valid_logging_levels = ["DEBUG", "INFO", "WARNING", "ERROR", "CRITICAL"]
+            if "loglevel" in key and value in valid_logging_levels:
```
dimitri-yatsenko, line 252:
```suggestion
            if "loglevel" == key and value in valid_logging_levels:
```

dimitri-yatsenko: should probably raise an error otherwise?

dimitri-yatsenko:
These changes are in implemented in #1186.

PR #1045, opened 2022-08-19
fix(logging): :bug: fix arguments sent to logger during config update

iamamutt:
--- Logging error --- 

`TypeError: not all arguments converted during string formatting`


guzman-raphael (approved):
@iamamutt Thank you! :pray:

PR #1050, opened 2022-09-02
Returning success count from the `.populate()` call

ttngu207:
Introduce a new argument `return_success_count` to the `.populate()` routine, that would return the count of successful `make()` calls (i.e. number of successful jobs) in one `.populate()` call. 

This functionality allows for us to answer the question: "Did anything happen in my `.populate()` call?" without having to call the `key_source` again (e.g. with `.progress()`)


datajoint/autopopulate.py
```diff
@@ -222,49 +225,62 @@ def handler(signum, frame):
...
         # restore original signal handler:
         if reserve_jobs:
             signal.signal(signal.SIGTERM, old_handler)
 
+        if suppress_errors and return_success_count:
+            return sum(success_list), error_list
         if suppress_errors:
             return error_list
+        if return_success_count:
+            return sum(success_list)
```
dimitri-yatsenko, line 283: If success_list is only used to report the sum, why not use a counter instead?

dimitri-yatsenko, line 283: These tuple returns are tough because the order matters. How about we return this as a dict instead? 
```suggestion
        ret = {}
        if suppress_errors:
            ret['error_list'] = error_list
        if return_success_count:
            ret['success_count'] = sum(success_list)
        return ret
```
This would be a backward-incompatible change, so we need to update users.

dimitri-yatsenko: If we use this dictionary return, then we can always return the success count and the user does not need to request it.

ttngu207: Agreed, but as you said, the biggest thing is breaking backward compatibility and if we can avoid it, we should.
Breaking backward-compatibility for such as minor new feature is not worth it in my view.

dimitri-yatsenko: I think it's okay to break backward compatibility for better design decisions long-term.

datajoint/autopopulate.py
```diff
@@ -322,6 +338,7 @@ def _populate1(
                     )
                     if jobs is not None:
                         jobs.complete(self.target.table_name, self._job_key(key))
+                    return True
```
dimitri-yatsenko, line 341: Not sure about this `True`. The docstring needs an update to explain. It's a bit odd to have so many different return types.

ttngu207:
Hi @dimitri-yatsenko , circling back to this as we need this feature for sciops. 
I do like much better the return for `autopopulate` as dict, very clean. But that means breaking backward compatibility. So, are we firm with the decision to do this? What is our protocol for these kind of backward incompatible changes - new major version, warning statement, etc.?

datajoint/autopopulate.py
ttngu207: What about returning `True/False` for success/failure instead of None?

dimitri-yatsenko (changes requested):
Would you update the CHANGELOG?

datajoint/autopopulate.py
```diff
@@ -176,6 +177,8 @@ def populate(
         :param max_calls: if not None, populate at most this many keys
         :param display_progress: if True, report progress_bar
         :param processes: number of processes to use. Set to None to use all cores
+        :param return_success_count: if True, return the count of successful `make()` calls.
```
dimitri-yatsenko, line 180: the docstring appears to be missing a description of the return argument.

datajoint/autopopulate.py
dimitri-yatsenko: At this point, the docstring did not specify the return argument. So it's okay to introduce a modification now with proper documentation.

datajoint/autopopulate.py
dimitri-yatsenko: Should we omit this input and always include the success count as part of the return dict?

datajoint/autopopulate.py
dimitri-yatsenko: make sense.

datajoint/autopopulate.py
ttngu207: I'm a bit unclear what you mean here. The docstring is included there. 
Do you mean it is not descriptive enough?

datajoint/autopopulate.py
dimitri-yatsenko: When does it ever return `False`?

datajoint/autopopulate.py
ttngu207: Oh, I see what you mean

datajoint/autopopulate.py
ttngu207: @dimitri-yatsenko , yes, docstring did not specify the return argument. But we would still break backward compatibility if we make the changes you suggested. I'm okay with it if you're okay with it

datajoint/autopopulate.py
dimitri-yatsenko: Then you should add `return False` to the case when the function returns without populating anything.

ttngu207: Does not populating anything constitute a `False`? `False` seems to indicate failure (maybe that's just me thinking so).

datajoint/autopopulate.py
dimitri-yatsenko: The docstring should include ` :return: ...`

tests_old/test_autopopulate.py
```diff
@@ -53,6 +53,23 @@ def test_populate(self):
...
+        assert_false(self.experiment, "table already filled?")
+        success_count = self.experiment.populate(return_success_count=True)
+        assert_equal(len(self.experiment.key_source & self.experiment), success_count)
+
+        # test restricted populate
+        assert_false(self.trial, "table already filled?")
+        restriction = self.subject.proj(animal="subject_id").fetch("KEY")[0]
+        d = self.trial.connection.dependencies
+        d.load()
+        success_count, _ = self.trial.populate(
```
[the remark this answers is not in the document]
dimitri-yatsenko: the test should reflect that `populate` returns a dict.

datajoint/autopopulate.py
```diff
@@ -222,49 +225,60 @@ def handler(signum, frame):
...
-                    if display_progress:
-                        progress_bar.update()
-            self.connection.connect()  # reconnect parent process to MySQL server
+            populate_kwargs = dict(
+                suppress_errors=suppress_errors,
+                return_exception_objects=return_exception_objects,
+                make_kwargs=make_kwargs,
+            )
+
+            if processes == 1:
```
dimitri-yatsenko, line 241: is `processes == 0` handled? Perhaps
```python
if not processes:
       return {
            "success_count": 0,
            "error_list": [],
        }
```

datajoint/autopopulate.py
```diff
@@ -222,49 +225,60 @@ def handler(signum, frame):
...
 
         # restore original signal handler:
         if reserve_jobs:
             signal.signal(signal.SIGTERM, old_handler)
 
-        if suppress_errors:
-            return error_list
+        return {
+            "success_count": sum(success_list),
+            "error_list": error_list if suppress_errors else None,
```
dimitri-yatsenko, line 280: I think it's okay to keep it as an empty list rather than `None`.
```suggestion
            "error_list": error_list,
```

datajoint/autopopulate.py
```diff
@@ -180,6 +180,9 @@ def populate(
             to be passed down to each ``make()`` call. Computation arguments should be
             specified within the pipeline e.g. using a `dj.Lookup` table.
         :type make_kwargs: dict, optional
+        :return: a dict with two keys
+            "success_count": the count of successful ``make()`` calls in this ``populate()`` call
+            "error_list": the error list if "suppress_errors" is set to True, otherwise None
```
dimitri-yatsenko, line 185:
```suggestion
            "error_list": the error list that is filled if `suppress_errors` is True
```

datajoint/autopopulate.py
dimitri-yatsenko: The current function produces `True` if successful, `None` if nothing is populated, throws an exception, or returns `(key, error)`. It never returns `False`. It seems that if a function returns `True`, it should also return `False` in other cases. Perhaps instead of `None`, we should return `False`. It would not indicate an error, but simply not populating for other reasons than an error.

dimitri-yatsenko (changes requested):

datajoint/autopopulate.py
ttngu207: It is handled in the `else` block

datajoint/autopopulate.py
ttngu207: Agreed, I'll update accordingly

datajoint/autopopulate.py
```diff
@@ -222,49 +225,60 @@ def handler(signum, frame):
...
+                    tqdm(keys, desc=self.__class__.__name__)
+                    if display_progress
+                    else keys
+                ):
+                    status = self._populate1(key, jobs, **populate_kwargs)
+                    if status is not None:
+                        if isinstance(status, tuple):
+                            error_list.append(status)
+                        elif status:
+                            success_list.append(1)
```
dimitri-yatsenko, line 252: If we change `_populate` to return `True`, `False`, or `(key, error)`
```suggestion
                    status = self._populate1(key, jobs, **populate_kwargs)
                    if status is True:
                        success_list.append(1)
                    elif isinstance(status, tuple):
                        error_list.append(status)
                    else:
                        assert(status is False)
```

dimitri-yatsenko (approved):

PR #1051, opened 2022-09-10
Add `json` data type

guzman-raphael:
- Add support for `json` type. Useful doc on querying with [JSON_VALUE](https://dev.mysql.com/doc/refman/8.0/en/json-search-functions.html#function_json-value).
  - [x] insert
  - [x] fetch
  - [x] projection
  - [x] querying
  - [x] indexing
  - [x] describe
  - [x] documentation
- Proposing here that we should only support `json` type in MySQL >=8.0 since `JSON_VALUE` approach is quite useful in returning with specific types, however, it was introduced in 8.x.
- Implement using MySQL's approach since it should be compatible with Percona as well. Though there is similar functionality in [MariaDB](https://mariadb.com/kb/en/json_value/), they've chosen to use `json` as an [alias](https://mariadb.com/kb/en/json-data-type/) for `longtext` + check constraint using `JSON_VALID`. Due to how different the implementations are, MariaDB won't be supported in our implementation of `json` type.
- Full fetch is supported and partial fetches can be performed by projecting first (since a proper attribute name should be set).
- From MySQL docs, looks like the size upper limit is the same as `longblob` (i.e. 4GB) per value.
- ~~Known gaps:~~
  - ~~support for creating indexes. It is possible but there is more needed work to make it compatible with `describe()`.~~

Depends on #1052, #1055, #1073


datajoint/condition.py
```diff
@@ -94,35 +113,59 @@ def make_condition(query_expression, condition, columns):
...
-            return '"%s"' % v
+            return f'{k}="{v}"'
         if isinstance(v, str):
-            return '"%s"' % v.replace("%", "%%")
-        return "%r" % v
+            return f'{k}="{v.replace("%", "%%")}"'
+        return f"{k}={v}"
+
+    def template(restrictions, operator=None):
+        return ("NOT (%s)" if negate else "%s") % (
```
dimitri-yatsenko, line 153: should `negate` be passed into template?

datajoint/condition.py
```diff
@@ -94,35 +113,59 @@ def make_condition(query_expression, condition, columns):
...
         ):
-            return '"%s"' % v
+            return f'{k}="{v}"'
         if isinstance(v, str):
-            return '"%s"' % v.replace("%", "%%")
-        return "%r" % v
+            return f'{k}="{v.replace("%", "%%")}"'
+        return f"{k}={v}"
+
+    def template(restrictions, operator=None):
```
dimitri-yatsenko, line 152: `operator=None` leads to an invalid template. I think it should default to `operator="AND"`.

datajoint/condition.py
```diff
@@ -94,35 +113,59 @@ def make_condition(query_expression, condition, columns):
...
+        return ("NOT (%s)" if negate else "%s") % (
+            restrictions[0]
+            if len(restrictions) == 1
+            else f"({f') {operator} ('.join(restrictions)})"
```
dimitri-yatsenko, line 156:
```suggestion
            f"({f') {operator} ('.join(restrictions)})"
```

datajoint/condition.py
```diff
@@ -151,23 +194,19 @@ def prep_value(k, v):
...
-                    if condition[k] is None
-                    else f"={prep_value(k, condition[k])}",
-                )
-                for k in common_attributes
-            )
-            + ")"
+        return template(
+            restrictions=[
+                prep_value(k, v)
+                for k, v in condition.items()
```
dimitri-yatsenko, line 206: what happened to the `IS NULL` comparison?

datajoint/condition.py
guzman-raphael: Sure, we can. Was just trying to keep it less verbose since we call this quite a bit and the argument would probably just be `negate=negate`, however, I can see how that would be less confusing though since it would be closer to true functional programming.

datajoint/condition.py
guzman-raphael: Good point. It does reduce nicely for the case of a single restriction still. Will issue a new update soon.

datajoint/condition.py
dimitri-yatsenko: also, the function name `template` is not quite accurate or descriptive. It's make a condition string.

guzman-raphael: It still works (mainly because I check `restriction` length first) but I see now a cleaner approach based on your feedback. New update coming soon.

datajoint/condition.py
guzman-raphael: Moved that to where we `prep_value` located [here](https://github.com/datajoint/datajoint-python/pull/1051/files#diff-27f86ee6bd86a0f1f7ffaa9ae916330d6ec989e73ae0db9215801892ac30989cR126-R127).

datajoint/condition.py
guzman-raphael: Perhaps `build_restriction`? or `assemble_restriction`?

dimitri-yatsenko: `join_conditions` also works

datajoint/autopopulate.py
```diff
@@ -323,9 +323,9 @@ def progress(self, *restrictions, display=True):
         total = len(todo)
         remaining = len(todo - self.target)
         if display:
```
dimitri-yatsenko, line 325: Then does it even make sense to have this `display` argument?

datajoint/condition.py
```diff
@@ -8,8 +8,29 @@
...
+        return match, key
+    match = match.groupdict()
+    if match["path"] is None:
+        return match, match["attr"]
+    else:
+        return match, "json_value(`{}`, _utf8mb4'$.{}'{})".format(
+            *{
+                k: ((f" returning {v}" if k == "type" else v) if v else "")
+                for k, v in match.items()
+            }.values()
```
dimitri-yatsenko, line 31: Here it's not necessary to create the dict for taking its `values()`
```suggestion
            *[
               ((f" returning {v}" if k == "type" else v) if v else "")
                for k, v in match.items()
            ]
```

datajoint/declare.py
```diff
@@ -515,10 +501,23 @@ def alter(definition, old_definition, context):
...
 def compile_index(line, index_sql):
-    match = index_parser.parseString(line)
+    def format_attribute(attr):
+        match, attr = translate_attribute(attr)
+        if match is None:
+            return attr
+        elif match["path"] is None:
+            return f"`{attr}`"
+        else:
+            return f"({attr})"
```
dimitri-yatsenko, line 511:
```suggestion
        if match is not None:
            return f"`{attr}`" if match["path"] is None else f"({attr})"
```

datajoint/table.py
```diff
@@ -761,7 +764,7 @@ def describe(self, context=None, printout=True):
                 unique="UNIQUE " if v["unique"] else "", attrs=", ".join(k)
             )
         if printout:
-            print(definition)
+            logger.info("\n" + definition)
         return definition
 
     def _update(self, attrname, value=None):
```
dimitri-yatsenko, line 770: If this is datajoint 0.14, then `_update` can be removed.

datajoint/autopopulate.py
guzman-raphael: Let's revisit in another PR. `display_progress` might need to be looking into as well.

The downside of removing this option is that using the `log_level` to control the display means the user is enabling a lot more logging than desired.

dimitri-yatsenko:
@guzman-raphael Would you remove the obsolete functionality as a separate PR, so that it's not tied to the introduction of JSON fields?

guzman-raphael:
> @guzman-raphael Would you remove the obsolete functionality as a separate PR, so that it's not tied to the introduction of JSON fields?

Makes sense. I've created a separate [PR](https://github.com/datajoint/datajoint-python/pull/1073) with the non-JSON updates. Once it is merged, I'll update this PR so that the diff is cleaner.

datajoint/condition.py
```diff
@@ -94,35 +115,58 @@ def make_condition(query_expression, condition, columns):
...
-            return '"%s"' % v.replace("%", "%%").replace("\\", "\\\\")
-        return "%r" % v
+            v = v.replace("%", "%%").replace("\\", "\\\\")
+            return f'{k}="{v}"'
+        return f"{k}={v}"
+
+    def join_conditions(negate, restrictions, operator="AND"):
+        return ("NOT (%s)" if negate else "%s") % (
+            f"({f') {operator} ('.join(restrictions)})"
+        )
```
dimitri-yatsenko, line 158:
```suggestion
        return   f"{'NOT ' if negate else ''} (%s)" % f") {operator} (".join(restrictions)"
```

dimitri-yatsenko (approved):

PR #1058, opened 2022-10-05
Schema activate ignores `create_tables=False`

guzman-raphael:


datajoint/schemas.py
```diff
@@ -223,8 +223,6 @@ def _decorate_table(self, table_class, context, assert_declared=False):
         instance = table_class()
         is_declared = instance.is_declared
         if not is_declared:
-            if not self.create_tables or assert_declared:
```
dimitri-yatsenko, line 226: removing the error message was all that it took to fix it?

guzman-raphael: It was for the test. I'll confirm with @yambottle when we get in if my fork also works for him. I'll post back here the result.

guzman-raphael: @dimitri-yatsenko Latest updates now appropriately reflect what worked for @yambottle and this is ready for re-review.

A few points to summarize:
- Previous check was requiring tables to be declared even if `create_tables=False` which is a bit unexpected.
- When we added a test case of 2 tables that the user does not have access to but that relate to each other, we experienced a permission error. Appears like the presence of a FK was enough for DJ to try to declare it. Was expecting something lazy in that it should only be declared on first select access (possible improvement idea for the future? :shrug:)
- Fixed this by only proceeding with declaration if all the conditions meet.

dimitri-yatsenko (approved):

PR #1062, opened 2022-11-02
`.populate(reserve_jobs=True)` to exclude `error` and `ignore` keys

ttngu207:
Currently, DataJoint's `.populate()` call with specified `max_calls` argument does not achieve the expected behavior. 

For example, if there are 10 keys remaining to be populated and of those 10, 3 are "error" and 2 are "ignore", and if those 5 error/ignore keys are top of the list, calling `Table.populate(reserve_jobs=True, max_calls=3)` will do nothing. The expected behavior should be that, with the same call, DataJoint should know to fetch back `keys` to be worked on but excluding those that are `error/ignore`. In the example above, this should be just the 5 remaining, not all 10, and from those keys, call `make()` on 3 of them. 

This PR fixes the described issue.


dimitri-yatsenko:
Looks good. I was hoping this could be done without that initial query by simply keeping track how many calls have already been made. But the multiprocessing makes it difficult.

dimitri-yatsenko (approved):

dimitri-yatsenko:
Please update the `CHANGELOG` and I will merge.

ttngu207:
> Please update the `CHANGELOG` and I will merge.

Thanks, CHANGELOG updated

PR #1067, opened 2022-12-09
Add support for insert CSV

CBroz1:
- [X] For `insert` method, check if object is pathlib.Path. If so, load as comma-delimited data
- [X] Modify test to run this method
- [X] Pass tests
- [X] Edit root README to reflect change from sphinx to mkdocs
- [X] Add documentation for new feature (add inserting various data types)

### DIFF
--- a/datajoint/table.py
+++ b/datajoint/table.py
@@ -353,10 +353,9 @@ class Table(QueryExpression):
             will be inserted.
         :param replace: If True, replaces the existing tuple.
         :param skip_duplicates: If True, silently skip duplicate inserts.
-        :param ignore_extra_fields: If False, fields that are not in the heading raise
-            error.
-        :param allow_direct_insert: applies only in auto-populated tables. If False
-            (default), insert are allowed only from inside the make callback.
+        :param ignore_extra_fields: If False, fields that are not in the heading raise error.
+        :param allow_direct_insert: Only applies in auto-populated tables. If False (default),
+            insert may only be called from inside the make callback.
 
         Example:
 
--- a/datajoint/table.py
+++ b/datajoint/table.py
@@ -6,6 +6,7 @@ import numpy as np
 import pandas
 import logging
 import uuid
+import csv
 import re
 from pathlib import Path
 from .settings import config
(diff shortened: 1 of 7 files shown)


datajoint/table.py
```diff
@@ -366,6 +370,10 @@ def insert(
                 drop=len(rows.index.names) == 1 and not rows.index.names[0]
             ).to_records(index=False)
 
+        if isinstance(rows, Path):
+            with open(rows, newline="") as data_file:
+                rows = list(csv.DictReader(data_file, delimiter=","))
```
dimitri-yatsenko, line 375: Awesome.  

Is `list` required here? Would it not work with just the `csv.DictReader` alone?

CBroz1: Leaving it as a `csv.DictReader` object here results in errors later when we go to read the rows outside of the `open` context
```python
File ~/Documents/dev/datajoint-python/datajoint/table.py:418, in <genexpr>(.0)
    415     return
    417 field_list = []  # collects the field list from first row (passed by reference)
--> 418 rows = list(
    419     self.__make_row_to_insert(row, field_list, ignore_extra_fields)
    420     for row in rows
    421 )
    422 if rows:
    423     try:

ValueError: I/O operation on closed file.
```

dimitri-yatsenko: Oh, that's because the `insert` is outside the `with` clause and the file has been closed. 

It's okay to leave as is and read into a list before inserting.

dimitri-yatsenko:
Looks great! Thanks!  Please update the CHANGELOG and we'll merge.

dimitri-yatsenko (commented):
Would it make sense to also accept the filepath in the form of a `str` as well as `pathlib.Path`?

CBroz1:
> Please update the CHANGELOG and we'll merge.

Updated

> Would it make sense to also accept the filepath in the form of a `str` as well as `pathlib.Path`?

My thought process was that a pathlib object is unambiguous. But I can adjust if you prefer accepting filepaths as strings. Is there any case where we would want to accept data as a string?

dimitri-yatsenko (approved):

PR #1068, opened 2023-01-09
add "ignore" function to set key's status to "ignore" in jobs table

ttngu207:
Add method to the "Jobs" class to set a job `key` to "ignore" status (similar to the existing methods of "reserve" and "error")


datajoint/jobs.py
```diff
@@ -87,6 +87,32 @@ def reserve(self, table_name, key):
...
         return True
 
+    def ignore(self, table_name, key):
+        """
+        Set a job to be ignored for computation.  When a job is ignored, the job table contains an entry for the
+        job key, identified by its hash, with status "ignore".
+
+        :param table_name: `database`.`table_name`
+        :param key: the dict of the job's primary key
+        :return: True if ignore job successfully. False = the jobs is already taken
```
guzman-raphael, line 97: Would you update this to a [Google Style](https://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings) docstring? That is what we are supposed to be using for API docs w/ MkDocs Material.

guzman-raphael (changes requested):
@ttngu207 Thanks for this contribution! :clap: 

Feature looks great. Minor suggestion about documentation but otherwise this is ready.

dimitri-yatsenko (commented):
Looks good. Just add entry in CHANGELOG

guzman-raphael (approved):

Issue #1060, opened 2022-10-07
Hidden Attributes


dimitri-yatsenko:
### Problem

Implement support for hidden attributes. These attributes will be prefixed with an underscore and not displayed.  DataJoint may store metadata like creation timestamps, usernames, etc in the hidden attributes. U

### Requirements
* DataJoint will support hidden attributes in tables. Any attribute whose name begins with a `_` will not be displayed and will not appear as part of the reverse-engineered table definition. 
* DataJoint declaration will allow declaring such attributes for internal purposes but the user API will be prohibited (by default) from using them.
* Hidden attributes will not participate in joins. 
* Hidden attributes will not be included in results of user queries

### Justification
* The singleton tables in #113 will be supported by a hidden attribute
* Automated metadata such timestamps, data versions, checksums, and usernames will use hidden attributes

CBroz1:
@A-Baji - Do you have a sense of your timeline for #1091? I think it could be really useful for something I'm working on

A-Baji:
> @A-Baji - Do you have a sense of your timeline for #1091? I think it could be really useful for something I'm working on

Hey Chris! I think that PR has just been waiting for a review from @dimitri-yatsenko. If any changes are requested feel free to PR to me with the fixes as I'm not currently working on any of the open source projects.

dimitri-yatsenko:
Fixed in #1091

Issue #113, opened 2015-06-30
Tables whose primary keys have zero attributes


dimitri-yatsenko:
Currently datajoint prohibits creating tables with no attributes in the primary key.  We can relax this prohibition. 

If the primary key has no attributes, it can only have one value.  Therefore, the relation can only contain one tuple.  So it's good for relations that contain a single fact about the universe rather than a collection of similar facts.  The relational algebra using such relations remains identical to what it is now.  In particular, joining two relations with no primary key produces another relation with no primary key and one tuple.

dimitri-yatsenko:
Note that this is different from SQL's tables that don't have a primary key.  SQL is not exactly relational and its tables can contain duplicate rows.  SQL does not have the equivalent of a primary key with zero attributes as defined here.

dimitri-yatsenko:
As @eywalker suggested, one way to think about this topic is to imaging that all tables have an additional attribute `_epsilon_` whose domain contains a single value `_Omega_`.  Then a relation with an empty primary key actually has one attribute in the primary key with only one possible value.  As such, this relation can contain only one tuple.

dimitri-yatsenko:
Although interesting to think about, there is little practical demand for this feature.  Closing.

dimitri-yatsenko:
If the primary key is empty, the table can have only one row.  It is accomplished by adding an invisible column, defined as 
```
_  = 1 : tinyint   # universe 
```

austin-hilberg:
Corresponding issue created for [datajoint-matlab](https://github.com/datajoint/datajoint-matlab/issues/127)

yambottle:
This issue is closed because there has been no activity for more than 365 days, please reopen if it's still applicable.

PR #1070, opened 2023-02-03
Add Params table tier. Add hidden heading attribute.

CBroz1:
In this PR, I ...
- Add a `Params` table tier.
  - This is motivated by the consistent design pattern across Elements whereby we want to set 'params' dictionaries as unique ([example](https://github.com/datajoint/element-array-ephys/blob/main/element_array_ephys/ephys_acute.py#L626)). This is cumbersome as the insert class methods need to package params as unique IDs and the visibility of these hashes may confuse new users.
  - In my implementation, I add a hidden hash field as a unique index that hashes all other fields with `param` in the name. Notably, this may include `params_a` and `params_b`, but would exclude other fields like descriptions
  - In diagrams, these are currently styled like Lookup tables but with a blue font
  - To appropriately intercept inserted data, I migrated a portion of the `insert` method to a `_normalize_insert_data` hidden method. In current implementation, this would run twice for Params inserts. Alternatively, a flag could skip duplicated normalization process.
- Add a `hide` attribute to fields
  - We've discussed adding hidden columns for other features. Here, this was motivated by a desire to hide the above param hash. 
  - Here, this is based on a prefix for fields , using `__` from the python convention. Notably, this required adjusting the set of characters permitted as the first character in an attribute name, but this will be backwards compatible with existing databases.
  - If loglevel is debug, there is no change. For all other log-levels, hidden fields are excluded from `preview()`, `fetch()`, and `fetch1()`, etc.
  - This is a partial (but not full) redress of #1060 - as it does not change the nature of joins and there is no barrier to a user adding a hidden attribute. In the current implementation, it could not be used for singleton tables, as described in (#113)
- Misc minor change: Add font color to diagram nodes - was this intentionally removed?

[nosetest output + coverage](https://github.com/datajoint/datajoint-python/files/10679365/nosetests_020723.txt)


dimitri-yatsenko (changes requested):
0. Before suggesting a solution as a PR, would you please write up an issue that would 
1. I would solve #1060 completely with all its repercussions before using hidden attributes as part of another solution. 
2. Adding a new data tier is a substantial enhancement of the data model across all supported APIS and we should go through a rigorous process for justifying it. 
3. The proposed implementation in this PR differs substantially from how the other tiers are handled with the use of prefixes.

datajoint/user_tables.py
```diff
@@ -152,6 +157,75 @@ class Lookup(UserTable):
...
+                row_values = tuple(row)
+            elif isinstance(row, collections.abc.Mapping):
+                row_values = row.values()
+            else:
+                row_values = row
+
+            paramset_dict = {
+                field: value
+                # zip_longest accounts for nullable params fields
+                for field, value in zip_longest(
```
dimitri-yatsenko, line 204: I advise against using `zip_longest` since this will likely lead to errors. When inserting records without naming the attributes, then all values must be provided.

datajoint/user_tables.py
```diff
@@ -152,6 +157,75 @@ class Lookup(UserTable):
...
+                    self.heading, row_values, fillvalue=None
```
dimitri-yatsenko, line 205: The default value for missing attributes is not `NULL`. The default value is specified in the table definition.

dimitri-yatsenko: We could create a special type for indicating the use of the `DEFAULT` value.

datajoint/user_tables.py
```diff
@@ -152,6 +157,75 @@ class Lookup(UserTable):
...
+                )
+                if "param" in field and "hash" not in field
+            }
+            paramset_hash = uuid_from_buffer(f"{paramset_dict}".encode())
+
+            existing_key = (self & {"__paramset_hash": paramset_hash}).fetch("KEY")
```
dimitri-yatsenko, line 211: The `insert` command should not use any other additional queries since it's considered to be an atomic operation.

tests/test_params.py
```diff
@@ -0,0 +1,63 @@
+from nose.tools import (
```
dimitri-yatsenko, line 1: We are getting away from `nose`. All new tests should not be implemented using `pytest`

tests/schema_simple.py
```diff
@@ -279,3 +279,18 @@ class OutfitPiece(dj.Part, dj.Lookup):
...
+    id: int
+    ---
+    desc='' : varchar(2)
+    params_a : longblob
+    params_b=null : longblob
+    """
+
+    def fill(self):
+        if not self:
+            self.insert([dict(id=n, desc="d", params_a=dict(a=n)) for n in range(3)])
```
dimitri-yatsenko, line 296:
```suggestion
            self.insert(dict(id=n, desc="d", params_a=dict(a=n)) for n in range(3))
```

datajoint/user_tables.py
```diff
@@ -152,6 +157,75 @@ class Lookup(UserTable):
...
+    """
+
+    _prefix = "##"
+    tier_regexp = (
+        r"(?P<params>" + _prefix + _base_regexp.replace("TIER", "params") + ")"
+    )
+
+    def __init__(self):
+        super().__init__()
+        if "params" not in self.definition:
```
dimitri-yatsenko, line 173: What does this condition test?  This would return `True` if the definition contains the substring `"params"` anywhere in the definition. This is not a meaningful test.

CBroz1:
Closing pending further discussion

PR #1084, opened 2023-05-11
dj.Top restriction

A-Baji:


datajoint/condition.py
```diff
@@ -61,6 +61,13 @@ def append(self, restriction):
             super().append(restriction)
 
 
+class Top:
+    def __init__(self, order_by, limit, offset=0):
```
dimitri-yatsenko, line 65: There may be a default value `limit=None`

datajoint/condition.py
```diff
@@ -183,6 +190,24 @@ def combine_conditions(negate, conditions):
             return not negate  # and empty AndList is True
         return combine_conditions(negate, conditions=items)
 
+    # restrict by Top
+    if isinstance(condition, Top):
```
dimitri-yatsenko, line 194: This may be okay for now, but ultimately we should figure out how to collapse multiple Top restrictions into one.

datajoint/expression.py
```diff
@@ -537,6 +591,14 @@ def tail(self, limit=25, **fetch_kwargs):
 
     def __len__(self):
         """:return: number of elements in the result set e.g. ``len(q1)``."""
+        subquery = None
+        if hasattr(self, "top_restriction") and self.top_restriction:
+            subquery = "(SELECT DISTINCT {fields} FROM {from_}{where}{sorting}) AS subquery".format(
```
dimitri-yatsenko, line 596: why `DISTINCT`?

datajoint/expression.py
```diff
@@ -112,24 +129,53 @@ def from_clause(self):
             )
         return clause
 
-    def where_clause(self):
+    def where_clause(self, right=False):
```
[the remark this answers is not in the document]
A-Baji: It is to handle restrictions that occur on the left or right of a `dj.Top` restriction. For example:
```Student & dj.Top('grade desc', 10) & 'lastname=Smith'```
vs
```Student & 'lastname=Smith' & dj.Top('grade desc', 10)```
 The first query restricts by the `lastname` attribute on the right side of the `dj.Top`, meaning it is applied to the result of `dj.Top`. The second query restricts by `lastname` first, and then gets the top.

File: datajoint/expression.py (hunk 3 above)
Comment (A-Baji) [reply datajoint/expression.py:596]: I wasn't sure how close the subquery should be to the original outer query. Is it not necessary?

File: datajoint/expression.py (hunk 4 above)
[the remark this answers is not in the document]
Comment (A-Baji) [reply datajoint/expression.py:132]: If we allow multiple Top restrictions in a QueryExpression then this method does not work.

File: datajoint/condition.py (hunk 2 above)
Comment (dimitri-yatsenko) [reply datajoint/condition.py:194]: Generally, we want to keep the query language algebraically closed. This means that any operation is applicable to any query.

File: datajoint/expression.py (hunk 4 above)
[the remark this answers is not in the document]
Comment (A-Baji) [reply datajoint/expression.py:132]: Could you provide an example to make sure I understand. What would the query for this look like:
```Student & 'grade>=70' & dj.Top('grade desc', 10) & 'lastname=Smith' & dj.Top('grade desc', 5)```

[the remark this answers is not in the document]
Comment (dimitri-yatsenko) [reply datajoint/expression.py:132]: This will likely be a subquery, effectively:

File: datajoint/expression.py [hunk 5]
```diff
@@ -112,24 +121,69 @@ def from_clause(self):
...
+        if offset and limit is None:
+            raise DataJointError("limit is required when offset is set")
+        clause = ""
+        if order_by is not None:
+            clause += " ORDER BY " + ", ".join(order_by)
+        if limit is not None:
+            clause += " LIMIT %d" % limit + (" OFFSET %d" % offset if offset else "")
+        return clause
+
+    def make_top_subquery(self, tops, fields=None, i=0):
```
dimitri-yatsenko, line 143: I would like why `make_top_subquery` necessary in addition to `make_subquery`. Why is not enough to use a regular subquery?

datajoint/condition.py
```diff
@@ -61,6 +61,13 @@ def append(self, restriction):
             super().append(restriction)
 
 
+class Top:
```
dimitri-yatsenko, line 64: Would `@dataclass` be appropriate? It's new in 3.7, so we didn't have it for other classes before we dropped support for 3.6.

datajoint/condition.py
```diff
@@ -183,6 +190,10 @@ def combine_conditions(negate, conditions):
             return not negate  # and empty AndList is True
         return combine_conditions(negate, conditions=items)
 
+    # restrict by Top
+    if isinstance(condition, Top):
+        return query_expression.make_subquery(top_restriction=condition)
```
dimitri-yatsenko, line 195: Should we pass `top_restriction` as a function argument or treat it as other restrictions, i.e. a member of the `QueryExpression` class?

dimitri-yatsenko: Perhaps even just add it as the last member of the `self.restrictions` list.

datajoint/condition.py
dimitri-yatsenko: Also, there is no good reason to handle `dj.Top` in `condition.py` since this module is dedicated to forming the `WHERE` and `HAVING` clauses for `SELECT` statements.  I would handle `dj.Top` entirely within `expression.py`

datajoint/condition.py
```diff
@@ -61,6 +61,13 @@ def append(self, restriction):
             super().append(restriction)
 
 
+class Top:
+    def __init__(self, order_by, limit=None, offset=0):
```
dimitri-yatsenko, line 65: The `order_by` argument should also be optional. It's possible to limit the entries without ordering them.

datajoint/condition.py
A-Baji: I am already setting it as an attribute of QueryExpression within `make_subquery`, but I can simply set it first and then make the subquery, removing the need for a `top_restriction' argument. It makes more sense to me to have it as its own attribute rather than including it in the `restriction` list.

dimitri-yatsenko: I take my comment back. I think we should intercept `dj.Top` in `QueryExpression.restrict`.

datajoint/condition.py
A-Baji: Like [so](https://github.com/datajoint/datajoint-python/pull/1084/files#diff-27f86ee6bd86a0f1f7ffaa9ae916330d6ec989e73ae0db9215801892ac30989cR66-R70)?

datajoint/condition.py
A-Baji: What should the first argument be? `limit` makes the most sense to me since `dj.Top(5, order_by='key desc')` seems more intuitive to me than `dj.Top('key desc', limit=5)`.

Also how do we want to handle empty `dj.Top()`s since they have no effect? Should we allow them silently, throw a warning, or throw an error?

dimitri-yatsenko: Just a few thoughts:

- `dj.Top(order_by="KEY")` should order by the primary key -- this has a minimal performance effect.
- Perhaps ordering by the primary key should be default behavior, so `oder_by` defaults to `"KEY"`.
- We could assume a sensible default limit, e.g. `limit=10` 
- Yes, then it may make sense to put limit as the first argument, something like: `__init__(limit=10, *, order_by="KEY", offset=0)`

datajoint/condition.py
dimitri-yatsenko: I wonder if we should default `order_by="KEY"`

datajoint/condition.py
```diff
@@ -61,6 +63,17 @@ def append(self, restriction):
             super().append(restriction)
 
 
+@dataclass
+class Top:
+    """
+    doc string
```
A-Baji, line 69: @dimitri-yatsenko Any suggestions for the doc string?

dimitri-yatsenko: You can refer to the documentation page and the original issue / PR that documents the feature.

datajoint/expression.py
```diff
@@ -73,6 +75,17 @@ def heading(self):
...
         return self._heading
 
+    @property
+    def top(self):
+        """a top object to form the ORDER BY, LIMIT, and OFFSET clauses"""
+        if self._top and self._top["order_by"]:
+            if isinstance(self._top["order_by"], str):
+                self._top["order_by"] = [self._top["order_by"]]
+            if "KEY" in self._top["order_by"]:
+                i = self._top["order_by"].index("KEY")
```
dimitri-yatsenko, line 85: compare how this is done in `fetch.Fetch.__call__` and replicate the same processing.

dimitri-yatsenko: check out `_flatten_attribute_list` in `fetch.py`.

datajoint/expression.py
```diff
@@ -119,17 +132,34 @@ def where_clause(self):
...
+        if self.top:
+            limit = self.top["limit"]
+            offset = self.top["offset"]
+            order_by = self.top["order_by"]
+        else:
+            return ""
+        if offset and limit is None:
+            raise DataJointError("limit is required when offset is set")
+        clause = ""
+        if order_by is not None:
```
dimitri-yatsenko, line 145: I am thinking that perhaps we should never use `LIMIT` without an `ORDER BY`. The default ordering should be by `KEY` rather than unordered. Let's discuss.

datajoint/expression.py
```diff
@@ -187,6 +217,13 @@ def restrict(self, restriction):
         string, or an AndList.
         """
         attributes = set()
+        if isinstance(restriction, Top):
+            self._top = dict(
```
dimitri-yatsenko, line 221: Why convert into a dict? Is `Top` not good enough?  You can use `dataclasses.asdict` too if wish to convert.

datajoint/condition.py
A-Baji: https://github.com/datajoint/datajoint-python/pull/1084/files#diff-27f86ee6bd86a0f1f7ffaa9ae916330d6ec989e73ae0db9215801892ac30989cR68-R71

dimitri-yatsenko (changes requested):

datajoint/condition.py
```diff
@@ -61,6 +63,18 @@ def append(self, restriction):
...
 
+@dataclass
+class Top:
+    """
+    A "restriction" to set the sorting clauses of a query. Since it is not a true
+    restriction, it has no effect on the WHERE clause.
+    """
+
+    limit: int = 10
+    order_by: Union[str, List[str]] = "KEY"
```
dimitri-yatsenko, line 74: I agree that we should not provide an option for `order_by=None`.  Restricting by `dj.Top` will always imply ordering. Ordering by primary key has no performance penalty.

dimitri-yatsenko: Since you are not allowing for `order_by=None`, then you should not be handling this case in the parsing code.

datajoint/expression.py
```diff
@@ -73,6 +75,17 @@ def heading(self):
         """a dj.Heading object, reflects the effects of the projection operator .proj"""
         return self._heading
 
+    @property
+    def top(self):
```
dimitri-yatsenko, line 79: I am not sure if parsing `self._top` is ncessary every time `self.top` is evoked. I would just the entire `dj.Top` object and parse it once at SQL generation time.

datajoint/expression.py
```diff
@@ -187,6 +217,13 @@ def restrict(self, restriction):
         string, or an AndList.
         """
         attributes = set()
+        if isinstance(restriction, Top):
+            self._top = dict(
+                limit=restriction.limit,
+                offset=restriction.offset,
+                order_by=restriction.order_by,
+            )
+            return self.make_subquery()
```
dimitri-yatsenko, line 226: Is this the right time to make a subquery? I think the subquery should be created when applying other operators after `dj.Top` has been applied.

datajoint/condition.py
```diff
@@ -61,6 +63,18 @@ def append(self, restriction):
             super().append(restriction)
 
 
+@dataclass
+class Top:
+    """
+    A "restriction" to set the sorting clauses of a query. Since it is not a true
+    restriction, it has no effect on the WHERE clause.
```
dimitri-yatsenko, line 70:
```suggestion
    A restriction to the top entities of a query. In SQL, this corresponds to ORDER BY ... LIMIT ... OFFSET
```

datajoint/condition.py
```diff
@@ -61,6 +63,18 @@ def append(self, restriction):
...
+    """
+
+    limit: int = 10
```
dimitri-yatsenko, line 73: Giving it another thought, I think the most sensible default value for limit is `limit = 1`.

datajoint/expression.py
A-Baji: If a limit is specified but no order, it will order by "KEY" by [default](https://github.com/datajoint/datajoint-python/pull/1084/files#diff-27f86ee6bd86a0f1f7ffaa9ae916330d6ec989e73ae0db9215801892ac30989cR74). However a user is still able to manually pass in `None` as a value, such as `dj.Top(None, None, 1)`. That is the edge case being handled here.

datajoint/condition.py
dimitri-yatsenko: restricting by `dj.Top` meets DataJoint's definiton of a restriction, so it's a true restriction. The fact that SQL does this differently does not make this family of restrictions any less true.

datajoint/expression.py
dimitri-yatsenko: In `dj.Top`, the input for `oder_by` does not allow the value `None`. I am okay with not allowing `None`. It simplifies the documentation burden and cognitive burden.  Since there is no performance hit for ordering by the primary key, there is no good reason to ever specify `order_by=None`.

tests/test_relational_operand.py
```diff
@@ -487,6 +488,72 @@ def test_restrictions_by_lists():
...
+            {"id_l": 10, "cond_in_l": 1},
+            {"id_l": 5, "cond_in_l": 1},
+        ]
+        assert z.fetch(as_dict=True) == [
+            {"id_l": 17, "cond_in_l": 1},
+            {"id_l": 20, "cond_in_l": 1},
+        ]
+
+    @staticmethod
+    @raises(DataJointError)
```
dimitri-yatsenko, line 548: This uses nosetests. Should we write all new tests for `pytests`?

datajoint/expression.py
A-Baji: You are correct. I did not think of this :sweat_smile:

datajoint/expression.py
```diff
@@ -661,10 +661,10 @@ def cursor(self, offset=0, limit=None, order_by=None, as_dict=False):
         """
         if offset and limit is None:
             raise DataJointError("limit is required when offset is set")
-        self._top = dict(
-            offset=offset,
-            limit=limit,
-            order_by=order_by,
+        self._top = Top(
```
dimitri-yatsenko, line 664: No, this does not belong here. These are the sorting values from the fetch, they should not override the existing `.top` proprty.

datajoint/expression.py
```diff
@@ -661,10 +661,10 @@ def cursor(self, offset=0, limit=None, order_by=None, as_dict=False):
         """
```
dimitri-yatsenko, line 657: `cursor` is only used internally in datajoint, so it's best to handle the sorting logic in the calling context as appropritate to each case, taking into consideration the existing `top` restruction.

datajoint/expression.py
A-Baji: It is actually still possible to pass `None` as an argument
image

datajoint/condition.py
A-Baji: Actually, the `QueryExpression.cursor` [method](https://github.com/A-Baji/datajoint-python/blob/top-restrcition/datajoint/expression.py#L653) can still pass in `None` arguments.

tests/test_relational_operand.py
A-Baji: I discussed this with @guzman-raphael . We would have to add the `pytest` module as a requirement for the dev container. It was more convenient to just to use `nosetest` in this case.

datajoint/expression.py
A-Baji: The top level `QueryExpression` will never have a `_top` property since `Tops` immediately terminate into a subquery.

dimitri-yatsenko: This may need to be fixed because now it sounds like `dj.Top` always generates a subquery even when one is unecessary.

Since now you have this new mechanism for `& dj.Top`, you can re-use inside `fetch` and consolidate the code rather than handling it separately here.

datajoint/expression.py
dimitri-yatsenko: yes, because Python ignores the type hinting.

datajoint/expression.py
```diff
@@ -119,17 +121,43 @@ def where_clause(self):
...
+        if offset and limit is None:
+            raise DataJointError("limit is required when offset is set")
+
+        # if 'order_by' passed in a string, make into list
+        if isinstance(order_by, str):
+            order_by = [order_by]
+        # expand "KEY" or "KEY DESC"
+        order_by = list(_flatten_attribute_list(self.primary_key, order_by))
+
+        clause = " ORDER BY " + ", ".join(order_by)
```
dimitri-yatsenko, line 144: No need to create a list
```suggestion
        clause = " ORDER BY " + ", ".join(o_flatten_attribute_list(self.primary_key, order_by))
```

datajoint/condition.py
```diff
@@ -61,6 +66,37 @@ def append(self, restriction):
...
+        if not (
+            isinstance(self.order_by, str)
+            or (
+                hasattr(self.order_by, "__iter__")
+                and all(isinstance(r, str) for r in self.order_by)
+            )
+        ):
+            raise DataJointError("Top order_by attributes must all be strings")
+        if not (isinstance(self.offset, int)):
+            raise DataJointError("Top offset must be an integer")
```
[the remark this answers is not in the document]
A-Baji: Do I apply the same error message pattern for `limit` and `order_by` or is this change specifically for `offset`?

dimitri-yatsenko (approved):

dimitri-yatsenko:
We just need documentation for this feature but I will be adding it shortly.

datajoint/condition.py
```diff
@@ -61,6 +63,37 @@ def append(self, restriction):
...
+        ):
+            raise TypeError("Top order_by attributes must all be strings")
+        if not isinstance(self.offset, int):
+            raise TypeError("The offset argument must be an integer")
+        if self.offset and self.limit is None:
+            self.limit = 999999999999  # arbitrary large number to allow query
+        if isinstance(self.order_by, str):
+            self.order_by = [
+                self.order_by
+            ]  # if 'order_by' passed in a string, make into list
```
dimitri-yatsenko, line 94: This comment just says what the code already says.

A-Baji: True 😅 . Remnant from copy-paste.

datajoint/expression.py
```diff
@@ -968,11 +968,14 @@ def _flatten_attribute_list(primary_key, attrs):
...
-    :return: generator of attributes where "KEY" is replaces with its component attributes
+    :return: generator of attributes where "KEY" is replaced with its component attributes
     """
     for a in attrs:
         if re.match(r"^\s*KEY(\s+[aA][Ss][Cc])?\s*$", a):
-            yield from primary_key
+            if primary_key:
+                yield from primary_key
+            else:
+                continue
```
dimitri-yatsenko, line 978:
```suggestion
```

datajoint/expression.py
```diff
@@ -968,11 +968,14 @@ def _flatten_attribute_list(primary_key, attrs):
...
         elif re.match(r"^\s*KEY\s+[Dd][Ee][Ss][Cc]\s*$", a):
             yield from (q + " DESC" for q in primary_key)
```
dimitri-yatsenko, line 980:
```suggestion
            if primary_key:
                yield from (q + " DESC" for q in primary_key)
```

A-Baji: Not sure I understand. I added this to handle an edge case for `QueryExpressions` that don't have primary keys ([example](https://github.com/datajoint/datajoint-python/blob/master/tests/test_relation_u.py#L60)), but still have a valid `order_by` value. If the `order_by` contains, for example, `["KEY", "first_name"]`, then `"KEY"` needs to be skipped.

dimitri-yatsenko: This handles `oder_by=["KEY desc", "first_name"]`

A-Baji: Ah I see. The diff was a little confusing.

datajoint/expression.py
```diff
@@ -119,17 +121,37 @@ def where_clause(self):
...
+    def sorting_clauses(self):
+        if not self._top:
+            return ""
+        limit = self._top.limit
+        order_by = self._top.order_by
+        offset = self._top.offset
+
+        clause = (
+            ""
+            if not self.primary_key and order_by == ["KEY"]
```
dimitri-yatsenko, line 133: Does this logic work for `["KEY desc"]`, or `["KEY "]`?

dimitri-yatsenko: perhaps this logic could be rolled into `_flatten_attribute_list`?

datajoint/expression.py
```diff
@@ -119,17 +121,37 @@ def where_clause(self):
             else " WHERE (%s)" % ")AND(".join(str(s) for s in self.restriction)
         )
 
+    def sorting_clauses(self):
+        if not self._top:
+            return ""
+        limit = self._top.limit
+        order_by = self._top.order_by
+        offset = self._top.offset
```
dimitri-yatsenko, line 129: Do we need to create these variables?

A-Baji: Probably not anymore, since they aren't occurring as often in this context after moving the type handling to the `__post_init__`.

datajoint/expression.py
A-Baji: It wouldn't, good catch. I'll convert it to regex matching.

datajoint/expression.py
```diff
@@ -119,17 +121,33 @@ def where_clause(self):
...
 
+    def sorting_clauses(self):
+        if not self._top:
+            return ""
+        clause = (
+            ""
+            if not any(_flatten_attribute_list(self.primary_key, self._top.order_by))
+            else f" ORDER BY \
+                {', '.join(_flatten_attribute_list(self.primary_key, self._top.order_by))}"
+        )
```
dimitri-yatsenko, line 132: we can avoid `calling _flatten_attribute_list` twice:
```suggestion
        clause = ', '.join(_flatten_attribute_list(self.primary_key, self._top.order_by)) 
        if clause:
            clause = f" ORDER BY {clause}"
```

datajoint/expression.py
```diff
@@ -931,3 +957,20 @@ def aggr(self, group, **named_attributes):
...
     aggregate = aggr  # alias for aggr
+
+
+def _flatten_attribute_list(primary_key, attrs):
+    """
+    :param primary_key: list of attributes in primary key
+    :param attrs: list of attribute names, which may include "KEY", "KEY DESC" or "KEY ASC"
+    :return: generator of attributes where "KEY" is replaced with its component attributes
+    """
+    for a in attrs:
```
dimitri-yatsenko, line 968: could you test this using attributes that are SQL reserved words, e.g. "select", "key", "index"? We may need to escape them.

A-Baji: They do seem to cause an issue:
```
datajoint.errors.QuerySyntaxError: ("You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'select, key, index LIMIT 1) as `$0`' at line 1", 'SELECT count(DISTINCT `id_l`) FROM (SELECT `id_l`,`cond_in_l` FROM `djtest_relational`.`#l` ORDER BY select, key, index LIMIT 1) as `$0`')
```

dimitri-yatsenko: SQL has a ton of reserved words. We typically do not prevent the user from using these names in attribute names. This means that we should escape these attribute names. In MySQL this is done with backticks around the attribute names.

dimitri-yatsenko: In MySQL, attributes can be escaped using backquotes.

datajoint/expression.py
```diff
@@ -931,3 +956,69 @@ def aggr(self, group, **named_attributes):
...
     aggregate = aggr  # alias for aggr
+
+
+def _flatten_attribute_list(primary_key, attrs):
+    """
+    :param primary_key: list of attributes in primary key
+    :param attrs: list of attribute names, which may include "KEY", "KEY DESC" or "KEY ASC"
+    :return: generator of attributes where "KEY" is replaced with its component attributes
+    """
+    sql_keywords = [
```
dimitri-yatsenko, line 967: I wouldn't track all keywords. Just enclose all attribute names in backquotes.

dimitri-yatsenko: This works for MySQL and we already use this approach for declaring tables and forming queries. See how the WHERE clause is formed from dict resitrctions.

datajoint/expression.py
A-Baji: Neither this:
```
@schema
class Keywords(dj.Lookup):
    definition = """  # tests sql keyword escaping
    select  : int
    key  : int
    ---
    index: int
    """
```
Nor this:
```
@schema
class Keywords(dj.Lookup):
    definition = """  # tests sql keyword escaping
    `select`  : int
    ---
    `index`: int
    """
```
Is working for me.

datajoint/expression.py
A-Baji: Using a similar method as how the WHERE clause is built wasn't working because it would wrap "asc" and "desc" too if it was included, which raises a "Column not found" error.

datajoint/expression.py
dimitri-yatsenko: What's the error message?  The first declaration should work as datajoint encloses the attribute names in backquotes. If that does not work, we should fix it.

A-Baji:
```
ERROR: Failure: AttributeError ('NoneType' object has no attribute 'groupdict')
----------------------------------------------------------------------
  File "/usr/local/lib/python3.7/site-packages/nose/failure.py", line 39, in runTest
    raise self.exc_val.with_traceback(self.tb)
  File "/usr/local/lib/python3.7/site-packages/nose/loader.py", line 418, in loadTestsFromName
    addr.filename, addr.module)
  File "/usr/local/lib/python3.7/site-packages/nose/importer.py", line 47, in importFromPath
    return self.importFromDir(dir_path, fqname)
  File "/usr/local/lib/python3.7/site-packages/nose/importer.py", line 94, in importFromDir
    mod = load_module(part_fqname, fh, filename, desc)
  File "/usr/local/lib/python3.7/imp.py", line 234, in load_module
    return load_source(name, filename, file)
  File "/usr/local/lib/python3.7/imp.py", line 171, in load_source
    module = _load(spec)
  File "<frozen importlib._bootstrap>", line 696, in _load
  File "<frozen importlib._bootstrap>", line 677, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 728, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/workspaces/datajoint-python/tests_old/test_relational_operand.py", line 19, in <module>
    from .schema_simple import (
  File "/workspaces/datajoint-python/tests_old/schema_simple.py", line 18, in <module>
    class Keywords(dj.Lookup):
  File "/workspaces/datajoint-python/datajoint/schemas.py", line 177, in __call__
    self._decorate_master(cls, context)
  File "/workspaces/datajoint-python/datajoint/schemas.py", line 189, in _decorate_master
    cls, context=dict(context, self=cls, **{cls.__name__: cls})
  File "/workspaces/datajoint-python/datajoint/schemas.py", line 226, in _decorate_table
    instance.declare(context)
  File "/workspaces/datajoint-python/datajoint/table.py", line 96, in declare
    sql, external_stores = declare(self.full_table_name, self.definition, context)
  File "/workspaces/datajoint-python/datajoint/declare.py", line 310, in declare
    ) = prepare_declare(definition, context)
  File "/workspaces/datajoint-python/datajoint/declare.py", line 265, in prepare_declare
    compile_index(line, index_sql)
  File "/workspaces/datajoint-python/datajoint/declare.py", line 447, in compile_index
    r"(?P<unique>unique\s+)?index\s*\(\s*(?P<args>.*)\)", line, re.I
AttributeError: 'NoneType' object has no attribute 'groupdict'
```

dimitri-yatsenko: that's a regular expression mismatch, which we should replace with a more meaningful error by the way. Can you rethrow it as `DataJointError("table definition syntax error in line ....")`?

datajoint/declare.py
```diff
@@ -443,9 +443,13 @@ def format_attribute(attr):
...
-    match = re.match(
-        r"(?P<unique>unique\s+)?index\s*\(\s*(?P<args>.*)\)", line, re.I
-    ).groupdict()
+    try:
+        match = re.match(
+            r"(?P<unique>unique\s+)?index\s*\(\s*(?P<args>.*)\)", line, re.I
+        ).groupdict()
+    except AttributeError:
+        raise DataJointError(f'Table definition syntax error in line "{line}"')
+
```
dimitri-yatsenko, line 452:
```suggestion
    match = re.match(
        r"(?P<unique>unique\s+)?index\s*\(\s*(?P<args>.*)\)", line, re.I)
    if match is None:
        raise DataJointError(f'Table definition syntax error in line "{line}"')
    match = match.groupdict()
```

docs/src/query/operators.md
```diff
@@ -229,7 +229,15 @@ Session & dj.top(limit=10, order_by='session_date')
...
 The result of this expression returns the first 10 rows of `Session` and sorts them 
-by their `session_date` in ascending order. If the `order_by` argument was instead: `session_date DESC`, then it would be sorted in descending order.
+by their `session_date` in ascending order. 
+
+### `order_by`
+
+| Example                                   | Description                                                                       |
+|-------------------------------------------|-----------------------------------------------------------------------------------|
+| `order_by="session_date DESC"`            | Sorts by `session_date` in *descending* order                                     |
+| `order_by="KEY"`                          | Sorts by the primary key(s)                                                       |
```
dimitri-yatsenko, line 239:
```suggestion
| `order_by="KEY"`                          | sort by the primary key |
| `order_by="KEY desc"                  | sort by primary key in descending order  |
```

docs/src/query/operators.md
```diff
@@ -229,7 +229,15 @@ Session & dj.top(limit=10, order_by='session_date')
...
+| `order_by=["subject_id", "session_date"]` | Sorts by `subject_id`, then sorts matching `subject_id`s by their `session_date`  |
 
 The default values for `dj.top` parameters are `limit=1`, `order_by="KEY"`, and `offset=0`.
```
dimitri-yatsenko, line 242:
```suggestion
The default values for `dj.Top` parameters are `limit=1`, `order_by="KEY"`, and `offset=0`.
```

A-Baji:
Screenshot for reference:
image

docs/src/query/operators.md
```diff
@@ -218,6 +219,29 @@ The examples below will use the table definitions in [table tiers](../reproduce/
 
 <!-- ## Join appears here in the general docs -->
 
+## Top
+
+Similar to the univeral set operator, the top operator uses `dj.Top` notation. It is used to 
+restrict a query by the given `limit`, `order_by`, and `offset` parameters:
```
kabilar, line 225:
```suggestion
Similar to the universal set operator, the top operator uses `dj.Top` notation. It is used to 
restrict a query by the given `limit`, `order_by`, and `offset` parameters.
```

docs/src/query/operators.md
```diff
@@ -218,6 +219,29 @@ The examples below will use the table definitions in [table tiers](../reproduce/
...
+
```
kabilar, line 226:
```
| Parameter | Keyword Arguments | Default Argument | Description |
| -- | -- | -- | -- |
| `limit` | | 1 | Restrict the number of rows returned. |
| `order_by` | "KEY", "DESC", "ASC" | "KEY" |  |
| `offset` | | 0  |  |
```

kabilar: Please fill out the rest of the table.  And perhaps the headings I am using (`Keyword Arguments`, `Default Argument`) are not the clearest, so feel free to replace.

docs/src/query/operators.md
```diff
@@ -218,6 +219,29 @@ The examples below will use the table definitions in [table tiers](../reproduce/
...
+```python
```
kabilar, line 227:
```suggestion
Examples

```python
```

docs/src/query/operators.md
```diff
@@ -218,6 +219,29 @@ The examples below will use the table definitions in [table tiers](../reproduce/
...
+Session & dj.Top(limit=10, order_by='session_date')
+```
+
+The result of this expression returns the first 10 rows of `Session` and sorts them 
+by their `session_date` in ascending order. 
+
+### `order_by`
```
kabilar, line 234:
```suggestion
```

docs/src/query/operators.md
```diff
@@ -218,6 +219,29 @@ The examples below will use the table definitions in [table tiers](../reproduce/
...
+
+| Example                                   | Description                                                                     |
+|-------------------------------------------|---------------------------------------------------------------------------------|
+| `order_by="session_date DESC"`            | Sort by `session_date` in *descending* order                                    |
+| `order_by="KEY"`                          | Sort by the primary key                                                         |
+| `order_by="KEY DESC"`                     | Sort by the primary key in *descending* order                                     |
+| `order_by=["subject_id", "session_date"]` | Sort by `subject_id`, then sort matching `subject_id`s by their `session_date`  |
+
+The default values for `dj.Top` parameters are `limit=1`, `order_by="KEY"`, and `offset=0`.
+
```
kabilar, line 244:
```suggestion
```

ethho:
This PR needs to:

- Merge `master`
- Update tests so that they're compatible with the new pytest-based suite.

Continuing work on this at https://github.com/datajoint/datajoint-python/pull/1178.

Issue #1083, opened 2023-05-11
S3 external files OS Error


horsto:
I am trying to retrieve external blobs saved in a Linode (Akamai) bucket (S3). This works fine most of the time, but larger blobs seem to run into errors during fetch: 

```python
---------------------------------------------------------------------------
OSError                                   Traceback (most recent call last)
File ~/miniconda3/envs/octo_code/lib/python3.10/site-packages/urllib3/response.py:444, in HTTPResponse._error_catcher(self)
    443 try:
--> 444     yield
    446 except SocketTimeout:
    447     # FIXME: Ideally we'd like to include the url in the ReadTimeoutError but
    448     # there is yet no clean way to get at it from this context.

File ~/miniconda3/envs/octo_code/lib/python3.10/site-packages/urllib3/response.py:567, in HTTPResponse.read(self, amt, decode_content, cache_content)
    566 with self._error_catcher():
--> 567     data = self._fp_read(amt) if not fp_closed else b""
    568     if amt is None:

File ~/miniconda3/envs/octo_code/lib/python3.10/site-packages/urllib3/response.py:533, in HTTPResponse._fp_read(self, amt)
    531 else:
    532     # StringIO doesn't like amt=None
--> 533     return self._fp.read(amt) if amt is not None else self._fp.read()

File ~/miniconda3/envs/octo_code/lib/python3.10/http/client.py:481, in HTTPResponse.read(self, amt)
    480 try:
--> 481     s = self._safe_read(self.length)
    482 except IncompleteRead:

File ~/miniconda3/envs/octo_code/lib/python3.10/http/client.py:630, in HTTPResponse._safe_read(self, amt)
    624 """Read the number of bytes requested.
    625 
    626 This function should be used when <amt> bytes "should" be present for
    627 reading. If the bytes are truly not available (due to EOF), then the
    628 IncompleteRead exception can be used to detect the problem.
    629 """
--> 630 data = self.fp.read(amt)
    631 if len(data) < amt:

File ~/miniconda3/envs/octo_code/lib/python3.10/socket.py:705, in SocketIO.readinto(self, b)
    704 try:
--> 705     return self._sock.recv_into(b)
    706 except timeout:

OSError: [Errno 22] Invalid argument

During handling of the above exception, another exception occurred:

ProtocolError                             Traceback (most recent call last)
Cell In[34], line 1
----> 1 data_to_export = (CuttleHeadMantle & key).fetch1(data2retrieve)
[...]

PR #1085, opened 2023-05-12
When retrieving data from S3, account for chunking

horsto:
There seems to be an issue with retrieval of larger ("external") files from S3 compatible buckets. I documented the error here: https://github.com/datajoint/datajoint-python/issues/1083. I brought it up as issue in the minio-api repository as well: https://github.com/minio/minio-py/issues/1280. 

The problem is that files can be returned as chunks via the minio API, such that `.data` cannot be called directly. Instead, `.stream` and subsequent concatenation of bytes seems to work for all cases. This behavior is already correctly implemented  in `fget()`  (https://github.com/datajoint/datajoint-python/blob/ee95f02e50ab077de46a35ebc020516751250160/datajoint/s3.py#L88)


dimitri-yatsenko (approved):

horsto:
This PR changes the `.data` method to `.stream` in the `get()` method for S3 (external) objects, which accommodates chunking of larger files (opposed to assuming that every / the whole file is loaded at once).

dimitri-yatsenko:
@horsto Would you merge this PR https://github.com/horsto/datajoint-python/pull/1 ?

dimitri-yatsenko (approved):

PR #1080, opened 2023-02-21
Fix readability of tutorials in dark mode

kushalbakshi:
This PR fixes the readability of the Jupyter Notebook outputs in the `Using the json type` tutorials. This resolves [issue 1078](https://github.com/datajoint/datajoint-python/issues/1078). 

Below are the screenshots from local testing after applying changes to the code as seen in this PR: 
Dark mode - the font in the tables previously appeared white, and was therefore, not readable. See linked issue above for images.

image

Light mode - there was no issue here previously and remains the same as before.

image

PR #1091, opened 2023-06-07
hidden attributes

A-Baji:


datajoint/heading.py
```diff
@@ -120,7 +121,7 @@ def table_status(self):
     def attributes(self):
         if self._attributes is None:
             self._init_from_database()  # lazy loading from database
-        return self._attributes
+        return {k: v for k, v in self._attributes.items() if not v.is_hidden}
```
ethho, line 124: Note: this changes from direct access to the `_attributes` dictionary to a copy. This means that the following will raise AssertionError as of this PR, but not on `master`:

```python
orig_value = my_table.heading.attributes['my_attr']
my_table.heading.attributes['my_attr'] = "foo"
assert my_table.heading.attributes['my_attr'] == "foo"
assert my_table.heading.attributes['my_attr'] != orig_value
```

ethho:
Moving development to #1080, which merges `master` and makes tests compatible with the new pytest framework.

PR #1094, opened 2023-06-16
[WIP] Add converted pytests from backup

guzman-raphael:
Issuing a in-progress PR on the pytest conversion of the test suite.

 I dug this up from an old backup. Unfortunately, the original source was corrupted so I used `decompyle3` to extract the source from `*.pyc` binaries. Seems like `assert` statements aren't quite right but there should be enough to go on.

Publishing it here as a draft in case anyone else wants to continue this since my availability will be limited.


ethho:
Implemented in #1142.

PR #1107, opened 2023-09-06
exclude "reserved" jobs in `Autopopulate.populate`

ttngu207:
This PR is somewhat related to https://github.com/datajoint/datajoint-python/pull/1062

Currently, the `max_calls` still does not achieve the expected behavior in cases of existing `reserved` jobs.

Consider the case where there already are 2 reserved jobs (two other processes working on one table with `reserve_jobs=True`). Another call of `Table.populate(max_calls=1, reserve_jobs=True)` will exit right away, instead of working the next one job (the expected behavior). 

This is because the list of jobs to be worked on next also include the "reserved" ones, then `max_calls=1` leads to selecting the one that is already "reserved".

This PR fixes the described problem.

However, there is a bit of a redundancy here, the logic to exclude "reserved" jobs already exists in `.populate1()`, just not early enough.

### DIFF
--- a/tests_old/test_autopopulate.py
+++ b/tests_old/test_autopopulate.py
@@ -58,17 +58,19 @@ class TestPopulate:
         assert_true(self.subject, "root tables are empty")
         assert_false(self.experiment, "table already filled?")
 
-        keys = self.experiment.key_source.fetch("KEY", limit=2)
+        keys = self.experiment.key_source.fetch("KEY", limit=3)
         for idx, key in enumerate(keys):
             if idx == 0:
                 schema.schema.jobs.ignore(self.experiment.table_name, key)
-            else:
+            elif:
                 schema.schema.jobs.error(self.experiment.table_name, key, "")
+            else:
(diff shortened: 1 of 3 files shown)


dimitri-yatsenko (approved):

tests_old/test_autopopulate.py
```diff
@@ -58,17 +58,19 @@ def test_populate_exclude_error_and_ignore_jobs(self):
         assert_true(self.subject, "root tables are empty")
         assert_false(self.experiment, "table already filled?")
 
-        keys = self.experiment.key_source.fetch("KEY", limit=2)
+        keys = self.experiment.key_source.fetch("KEY", limit=3)
         for idx, key in enumerate(keys):
             if idx == 0:
                 schema.schema.jobs.ignore(self.experiment.table_name, key)
-            else:
+            elif:
```
dimitri-yatsenko, line 65: `elif` what?

ttngu207: Thanks for the catch, fixed.

dimitri-yatsenko (approved):

Issue #1110, opened 2023-09-21
Error on cascade delete in MySQL 8


samuelbray32:
### Description
- Some users have been getting an IntegrityError 1217 when calling delete on an entry in a table with populated children
- The same user is able to manually delete each child table entry, and then delete the original entry once all children are removed
- The issue comes from the IntegrityError from `delete_quick` not being caught in the `cascade` function in `delete`, which is why all the children need deleted before can delete a table.
- The reason it's not getting caught is that datajoint is expecting the error code for ROW_IS_REFERENCED_2 (1451) but is instead getting the error code ROW_IS_REFERENCED (1217). The ROW_IS_REFERENCED error also returns less information about the failing table

### Reproducibility
- Linux
- Python Version: 3.9.16
- MySQL Version: 8.0.34
- MySQL Deployment Strategy: remote
- DataJoint Version: 0.14.1
- Minimum number of steps to reliably reproduce the issue:
Error only occurs for some users. Occurs when calling `TableName.delete({restriction}) ` on a table with child tables populated for the restriction

- Example error stack

> ---------------------------------------------------------------------------

> IntegrityError                            Traceback (most recent call last)
> Cell In[2], line 1
> ----> 1 Session().delete({"nwb_file_name":nwb_file_name})
> 
> File [~/mambaforge-pypy3/envs/spyglass/lib/python3.9/site-packages/datajoint/table.py:588](https://file+.vscode-resource.vscode-cdn.net/home/sambray/Documents/spyglass/notebooks/~/mambaforge-pypy3/envs/spyglass/lib/python3.9/site-packages/datajoint/table.py:588), in Table.delete(self, transaction, safemode, force_parts)
>     586 # Cascading delete
>     587 try:
> --> 588     delete_count = cascade(self)
>     589 except:
>     590     if transaction:
> 
> File [~/mambaforge-pypy3/envs/spyglass/lib/python3.9/site-packages/datajoint/table.py:506](https://file+.vscode-resource.vscode-cdn.net/home/sambray/Documents/spyglass/notebooks/~/mambaforge-pypy3/envs/spyglass/lib/python3.9/site-packages/datajoint/table.py:506), in Table.delete..cascade(table)
>     504 for _ in range(max_attempts):
>     505     try:
> --> 506         delete_count = table.delete_quick(get_count=True)
>     507     except IntegrityError as error:
>     508         match = foreign_key_error_regexp.match(error.args[0]).groupdict()
> 
[...]

PR #1112, opened 2023-10-25
Add MySQL 8 error code for cascade delete 

CBroz1:
To address #1110, I've added this error code to the set of checks performed by `translate_query_error`. 

To protect against potential issues for MySQL 5.7 users, I checked error code manuals for both [5.7](https://dev.mysql.com/doc/mysql-errors/5.7/en/server-error-reference.html) and [8.0](https://dev.mysql.com/doc/mysql-errors/8.0/en/server-error-reference.html), which indicated that theses error codes would be categorized as `IntegrityError` across both cases.

Are additional tests required for this change?

### DIFF
--- a/datajoint/connection.py
+++ b/datajoint/connection.py
@@ -79,6 +79,8 @@ def translate_query_error(client_error, query):
     # Integrity errors
     if err == 1062:
         return errors.DuplicateError(*args)
+    if err == 1217:  # MySQL 8 error code
+        return errors.IntegrityError(*args)
     if err == 1451:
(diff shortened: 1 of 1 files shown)


dimitri-yatsenko (commented):
Thank you! @CBroz1

lfrank:
@dimitri-yatsenko Just checking in on this, as it would be really helpful for us to have it merged. Thoughts on when that might be possible?

dimitri-yatsenko (approved):
Need to add a comment to CHANGELOG

PR #1113, opened 2023-11-28
convert `test_aggr_regressions.py` to pytest

A-Baji:


A-Baji:
We've decided to create a dev branch to which all test conversions will be merged to.

PR #1114, opened 2023-12-01
PLAT-138 Convert pytest-compatible nosetest modules

A-Baji:


ethho (approved):
LGTM. Only comment is that we should make it more explicit what we're passing as the `context`. For instance, in `dj.Schema('my_schema', context=locals())` it's not clear what the context contains. We'll clean this up next PR.

PR #1115, opened 2023-12-02
Fix dev container

ethho:
Followed the instructions in https://code.visualstudio.com/docs/devcontainers/create-dev-container to recreate the `.devcontainer` directory and contents from scratch, from the `LNX-docker-compose.yml` compose stack as a base. @A-Baji could you test to see if this works for you in VS code?


A-Baji (commented):
The container builds successfully now. The only issue is that it seems to be missing `nosetest` and the `mysql` CLI dependencies. These are listed as features in the current docs so it's best to include them. Also I'd set `DJ_PASS` to `simple` to be consistent with the example in the docs.

It's probably a good idea to test all the [features mentioned in the docs](https://datajoint.com/docs/core/datajoint-python/0.14/develop/#features). Once those work we'll be good to merge.

ethho:
After changing any of the configuration in the `LNX-docker-compose.yml` or `.devcontainer/*`, one must do the following to accurately reproduce the dev container with a clean environment:

1. Quit VS code if it is open
2. Stop and remove all running Docker containers: `docker ps -aq | xargs docker stop | xargs docker rm`
3. Reopen the repo in VS code: `code .`
4. Issue the VS code command: `Dev Containers: Rebuild Without Cache and Reopen in Container` if the project is opened locally `Dev Containers: Rebuild Container Without Cache` if project is already open in dev container.

ethho:
> Also I'd set `DJ_PASS` to `simple` to be consistent with the example in the docs.

For MySQL 8.0 only (not 5.7), `MYSQL_ROOT_PASSWORD` is sometimes not respected, and switches to the image default `password`. I believe this happens when there is an existing data directory, see https://stackoverflow.com/questions/40149880/docker-mysql-root-password-do-not-work. I think the best option is to start using the MySQL default password `password`, and I'll update the developer docs. But let me know if you think otherwise.

dimitri-yatsenko (approved):

ethho:
`DJ_PASS=password PY_VER=3.8 MYSQL_VER=8.0 DISTRO=alpine MINIO_VER=RELEASE.2022-08-11T04-37-28Z HOST_UID=$(id -u) docker compose -f LNX-docker-compose.yml up --exit-code-from app --build --always-recreate-deps --force-recreate` now failing. This is likely due to the devcontainer image being tagged as `datajoint/djtest`. Fix is WIP.

PR #1116, opened 2023-12-02
PLAT-138  Follow Up to #1114

ethho:
I made some edits to the tests that were recently migrated to pytest in #1114. Minor edits so that we're closer to pytest standard practice. This includes decreasing the scope of the fixtures that return schemas.

This includes the changes from #1115 so please merge that one before this one.


tests/schema_advanced.py
```diff
@@ -135,3 +133,6 @@ class GlobalSynapse(dj.Manual):
     -> Cell.proj(pre_slice="slice", pre_cell="cell")
     -> Cell.proj(post_slice="slice", post_cell="cell")
     """
+
+LOCALS_ADVANCED = {k: v for k, v in locals().items() if inspect.isclass(v)}
+__all__ = list(LOCALS_ADVANCED.keys())
```
dimitri-yatsenko, line 138:
```suggestion
__all__ = list(LOCALS_ADVANCED)
```

tests/schema.py
```diff
@@ -450,3 +448,7 @@ class Longblob(dj.Manual):
     ---
     data: longblob
     """
+
+
+LOCALS_ANY = {k: v for k, v in locals().items() if inspect.isclass(v)}
+__all__ = list(LOCALS_ANY.keys())
```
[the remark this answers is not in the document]
ethho: Thanks for the review. I'll make these fixes in another WIP PR to `datajoint:dev-tests` branch.

tests/test_adapted_attributes.py
```diff
@@ -0,0 +1,153 @@
...
+def enable_filepath_feature(monkeypatch):
+    monkeypatch.setenv(FILEPATH_FEATURE_SWITCH, 'TRUE')
+    yield
+    monkeypatch.delenv(FILEPATH_FEATURE_SWITCH, raising=True)
+
+
+@pytest.fixture
+def schema_name_custom_datatype():
+    schema_name = PREFIX + "_test_custom_datatype"
+    return schema_name
```
A-Baji, line 35: Similar to `schema_adapted.GraphAdapter()`, we could just use a global variable here.

tests/test_adapted_attributes.py
```diff
@@ -0,0 +1,153 @@
...
+import networkx as nx
+from itertools import zip_longest
+from . import schema_adapted
+from .schema_adapted import Connectivity, Layout
+from . import PREFIX, S3_CONN_INFO
+
+
+@pytest.fixture
+def adapted_graph_instance():
+    yield schema_adapted.GraphAdapter()
```
A-Baji, line 15: Can't we just use `schema_adapted.GraphAdapter()` directly instead of accessing it through a fixture?

tests/test_adapted_attributes.py
```diff
@@ -0,0 +1,153 @@
...
+        **schema_adapted.LOCALS_ADAPTED,
+        'graph': adapted_graph_instance,
+        'layout_to_filepath': layout_to_filepath,
+    }
+    schema = dj.schema(schema_name_custom_datatype, context=context, connection=connection_test)
+    graph = adapted_graph_instance
+    schema(schema_adapted.Connectivity)
+    schema(schema_adapted.Layout)
+    yield schema
+    schema.drop()
```
A-Baji, line 60: Any reason not to set up `schema_advanced` in `conftest.py`?

tests/test_adapted_attributes.py
ethho: We could make it a global var too. I put it in a fixture to avoid duplicating logic, because it's used by multiple other fixtures.

tests/test_adapted_attributes.py
ethho: We probably could, yes.

ethho:
CI is failing `black tests --check`. This is fixed in a downstream PR #1117

tests/test_adapted_attributes.py
A-Baji: I think it's less confusing as a global constant: `SCHEMA_NAME = PREFIX + "_test_custom_datatype"`

PR #1124, opened 2023-12-11
PLAT-150: Migrate test_bypass_serialization.py

ethho:
- cp to tests
- Migrate test_bypass_serialization

Also contains changes from #1123 and before.


tests/conftest.py
```diff
@@ -307,6 +308,20 @@ def schema_ext(connection_test, stores_config, enable_filepath_feature):
     schema.drop()
 
 
+@pytest.fixture
```
ethho, line 311: I defined this schema in `conftest.py` and `schema_uuid.py` instead of the test module. The reasoning is that `tests/test_aggr_regressions.py` also uses the `schema_uuid` fixture for test `test_left_join_len`.

A-Baji: Generally I think any schema that had it's own file in `tests_old` should be kept that way, with a fixture for it set up in `conftest.py` unless it's deemed unnecessary.

A-Baji (approved):

PR #1127, opened 2023-12-11
PLAT-154: Migrate test_external

ethho:
- cp to tests
- nose2pytest test_external
- Use pytest tmpdir fixtures for mocking stores
- Migrate test_external
- Format with black

PR #1129, opened 2023-12-12
PLAT-156: Migrate test_fetch

ethho:
- Changes in #1127
- Changes in #1128
- cp to tests
- First pass at migrating test_fetch

### DIFF
--- a/tests/schema.py
+++ b/tests/schema.py
@@ -307,7 +307,7 @@ class DecimalPrimaryKey(dj.Lookup):
     definition = """
     id  :  decimal(4,3)
     """
-    contents = zip((0.1, 0.25, 3.99))
+    contents = list(zip((0.1, 0.25, 3.99)))
 
 
 class IndexRich(dj.Manual):
--- a/tests/test_fetch.py
+++ b/tests/test_fetch.py
@@ -1,0 +1,390 @@
+from nose.tools import (
+    assert_true,
+    raises,
+    assert_equal,
+    assert_dict_equal,
+    assert_list_equal,
+    assert_set_equal,
+)
+from operator import itemgetter
+import itertools
+import numpy as np
+import decimal
+import pandas
+import warnings
+from . import schema
+from .schema import Parent, Stimulus
+import datajoint as dj
+import os
+import logging
+import io
+
+logger = logging.getLogger("datajoint")
+
+
+class TestFetch:
+    @classmethod
+    def setup_class(cls):
+        cls.subject = schema.Subject()
+        cls.lang = schema.Language()
+
+    def test_getattribute(self):
+        """Testing Fetch.__call__ with attributes"""
+        list1 = sorted(
+            self.subject.proj().fetch(as_dict=True), key=itemgetter("subject_id")
+        )
+        list2 = sorted(self.subject.fetch(dj.key), key=itemgetter("subject_id"))
+        for l1, l2 in zip(list1, list2):
+            assert_dict_equal(l1, l2, "Primary key is not returned correctly")
+
+        tmp = self.subject.fetch(order_by="subject_id")
+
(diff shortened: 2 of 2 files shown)


tests/test_fetch.py
```diff
@@ -0,0 +1,402 @@
...
+
+@pytest.fixture
+def lang():
+    yield schema.Language()
+
+
+@pytest.fixture
+def languages(lang) -> List:
+    og_contents = lang.contents
+    languages = og_contents.copy()
```
ethho, line 24: I had to copy, store, and restore the `contents` of this table because the `contents` would change between tests. This caused some tests such as `test_fetch1_step1` to fail if other tests (that sort `contents`) such as `test_order_by` were run before it.

tests/test_fetch.py
```diff
@@ -0,0 +1,402 @@
...
+        """
+        Issue #595
+        """
+        attrs = ("species", "date_of_birth")
+        result = subject.fetch(*attrs, as_dict=True)
+        assert bool(result) and len(result) == len(subject)
+        assert set(result[0]) == set(attrs)
+
+    def test_fetch1_step1(self, schema_any, lang, languages):
+        assert (
```
ethho, line 164: Added this assertion to check if upstream tests like `test_offset_by` changed the `contents` and broke this test.

tests/schema.py
```diff
@@ -307,7 +307,7 @@ class DecimalPrimaryKey(dj.Lookup):
     definition = """
     id  :  decimal(4,3)
     """
-    contents = zip((0.1, 0.25, 3.99))
+    contents = list(zip((0.1, 0.25, 3.99)))
```
ethho, line 310: Strangely, the logic in `test_fetch::test_offset` would exhaust this zip object, causing it to be empty when `test_fetch::test_decimal` was run, causing the latter test to fail.

It's possible that `ORDER BY` on the upstream table causes `DecimalPrimaryKey` to also sort (and therefore exhaust the zip object):

https://github.com/ethho/datajoint-python/blob/761c1663f4a2be45e521e84f3d33ef79b34b3ba4/tests/test_fetch.py#L205-L207

```python
    def test_offset(self, schema_any, lang, languages):
        """Tests offset"""
        cur = lang.fetch(limit=4, offset=1, order_by=["language", "name DESC"])
```

tests/test_fetch.py
```diff
@@ -0,0 +1,399 @@
...
+from . import schema
+import datajoint as dj
+import os
+import logging
+import io
+
+
+@pytest.fixture
+def lang():
+    yield schema.Language()
```
A-Baji, line 18: What is the purpose of accessing `schema.Language()` through a fixture vs directly? I tested directly accessing the table and the suite passed.

We also stick with the `setup_class` method for the same scenario in [test_autopopulate.py](https://github.com/datajoint/datajoint-python/blob/dev-tests/tests/test_autopopulate.py#L13-L20) and [test_relation_u.py](https://github.com/datajoint/datajoint-python/blob/dev-tests/tests/test_relation_u.py#L13-L23). 

I think it's important to maintain a consistent pattern for recurring scenarios. That way we establish a standard for future cases.

A-Baji (approved):

tests/test_fetch.py
ethho: The purpose is consistency. We want to control the state that each test receives, isolating at the test level as much as possible. In some cases, like the `languages` fixture below, this is necessary (https://github.com/datajoint/datajoint-python/pull/1129/files#r1424702143) because the fixture actually changes the table instance.

tests/schema.py
dimitri-yatsenko: zip can only be used once. Once exhausted, it's empty. So if you are using the same code more than once to populate a lookup table, you need to convert into a list first.

PR #1132, opened 2023-12-13
PLAT-160: Migrate test_privileges

ethho:
- No changes from previous PRs. This branch was forked directly from `dev-tests`
- cp to tests
- Migrate TestUnprivileged
- Format with black
- Separate fixture for DB creds dict
- First pass at migrating test_privileges
- Clean up


tests/conftest.py
```diff
@@ -52,12 +53,17 @@ def enable_filepath_feature(monkeypatch):
 
 
 @pytest.fixture(scope="session")
-def connection_root_bare():
-    connection = dj.Connection(
+def db_creds_root() -> Dict:
```
ethho, line 56: We need the value of `os.getenv("DJ_HOST")` in several test modules. Best to perform this once and inject it as a fixture. Making a note to replace the pattern `os.getenv("DJ_HOST")` with this fixture in other modules besides `test_privileges`.

ethho: Another thing that we should move to fixtures during the "big clean up".

A-Baji (approved):

tests/conftest.py
A-Baji: Why can't we just import `CONN_INFO` (as done in #1133)

PR #1135, opened 2023-12-14
PLAT-163: Migrate test_relational_operand

ethho:
- cp to tests
- nose2pytest test_relational_operand
- First pass at migrating test_relational_operand
- Format with black

### DIFF
--- a/tests/schema.py
+++ b/tests/schema.py
@@ -147,7 +147,7 @@ class Experiment(dj.Imported):
         from datetime import date, timedelta
 
         users = [None, None] + list(User().fetch()["username"])
-        random.seed("Amazing Seed")
+        random.seed("Amazing Seed4")
         self.insert(
             dict(
                 key,
--- a/tests/test_relational_operand.py
+++ b/tests/test_relational_operand.py
@@ -1,0 +1,672 @@
+import random
+import string
+import pandas
+import datetime
+
+import numpy as np
+from nose.tools import (
+    assert_equal,
+    assert_false,
+    assert_true,
+    raises,
+    assert_set_equal,
+    assert_list_equal,
+)
+
+import datajoint as dj
+from .schema_simple import (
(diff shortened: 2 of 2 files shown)


tests/schema.py
```diff
@@ -147,7 +147,7 @@ def make(self, key):
         from datetime import date, timedelta
 
         users = [None, None] + list(User().fetch()["username"])
-        random.seed("Amazing Seed")
+        random.seed("Amazing Seed4")
```
ethho, line 150: For `test_restriction_by_null`, we needed some users in this table with name `None`. I changed the seed until the table contents fulfilled this restriction, and `test_restriction_by_null` passed.

tests/test_relational_operand.py
```diff
@@ -0,0 +1,596 @@
...
+def schema_any_pop(schema_any):
+    """
+    Schema any with data populated.
+    """
+    Experiment.populate()
+    yield schema_any
+
+
+def test_populate(schema_simp_pop):
+    assert not B().progress(display=False)[0], "B incompletely populated"
```
ethho, line 63: I did not separate `B()` into a separate fixture in this module. In my opinion, the best practice that balances consistency and readability is:

- If the table instance was being instantiated in a `setup_class` in the old tests (like `cls.subject = Subject()`), then create and use a `subject` fixture instead. In these cases, the original test authors used `self.subject` instead of `Subject()` in the tests.
- If the table instance is being constructed inline, as is the case for `B()` in this module, then the original test authors weren't worried about state that carried over between tests, and we can leave the inline construction.

Will gladly take feedback on this convention (which I'm aware is changing as we go through). I could replace every `B()` with a fixture `b`, but I think this would make the module much less readable, even if it's more consistent with other modules that need fixtures for this.

A-Baji (approved):

PR #1136, opened 2023-12-14
PLAT-166: Migrate test_schema

ethho:
- cp to tests
- nose2pytest test_schema
- Add db_creds_test fixture
- First pass at migrating test_schema
- Mock schema_empty module
- Move call to spawn_missing_classes to test

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -57,6 +57,15 @@ def enable_filepath_feature(monkeypatch):
 
 
 @pytest.fixture(scope="session")
+def db_creds_test() -> Dict:
+    return dict(
+        host=os.getenv("DJ_TEST_HOST", "fakeservices.datajoint.io"),
+        user=os.getenv("DJ_TEST_USER", "datajoint"),
+        password=os.getenv("DJ_TEST_PASSWORD", "datajoint"),
+    )
+
+
+@pytest.fixture(scope="session")
 def db_creds_root() -> Dict:
     return dict(
         host=os.getenv("DJ_HOST", "fakeservices.datajoint.io"),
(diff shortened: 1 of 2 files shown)


tests/conftest.py
```diff
@@ -56,6 +56,15 @@ def enable_filepath_feature(monkeypatch):
...
 
 
+@pytest.fixture(scope="session")
+def db_creds_test() -> Dict:
+    return dict(
+        host=os.getenv("DJ_TEST_HOST", "fakeservices.datajoint.io"),
+        user=os.getenv("DJ_TEST_USER", "datajoint"),
+        password=os.getenv("DJ_TEST_PASSWORD", "datajoint"),
+    )
+
```
ethho, line 66: Another part of the move towards putting these credentials in fixtures.

tests/test_schema.py
```diff
@@ -0,0 +1,244 @@
...
+        'schema': schema_empty,
+        'Ephys': Ephys,
+    }
+    module = types.ModuleType('schema_empty')
+
+    # Add classes to the module's namespace
+    for k, v in namespace_dict.items():
+        setattr(module, k, v)
+
+    return module
```
ethho, line 48: This was an interesting problem. In `tests_old/schema_empty.py`, `spawn_missing_classes` relies on the `schema.schema` (called the `schema_any` fixture in new tests) being activated on import. Because this is now in the `schema_any` fixture, we weren't able to create `schema_empty.schema` the same way in the pytest suite (in pytest, the value of the fixture is only available *after* everything's been imported). Instead, I used the `types.ModuleType` to mock a module that mimics `tests_old/schema_empty.py`.

tests/test_schema.py
```diff
@@ -0,0 +1,244 @@
...
+
+    # hack to update the locals dictionary
+    locals()
+
+    @test_schema
+    class Cell(dj.Manual):
+        definition = """
+        type:  varchar(32)    # type of cell
+        """
+
```
ethho, line 171: A bit inconsistent in the way we add `Cell` to `test_schema`, but this is exactly the way it was written in the old tests, so I'm just going to leave it.

ethho:
Merging into `dev-tests`. Please include review as part of #1142 instead.

PR #1139, opened 2023-12-14
PLAT-169: Migrate test_update1

ethho:
- cp to tests
- nose2pytest test_update1
- Format with black


ethho:
Merging into `dev-tests`. Please include review as part of #1142 instead.

Issue #151, opened 2015-08-28
Cascading delete for Part


eywalker:
The cascading delete still doesn't handle the case involving `dj.Part` correctly. Basically, any deletion on `dj.Part` should cause deletion on its _parent_ table. Marking deletion on the _parent_ table then should cause appropriate cascading of deletion on the _parent's_ as well as `Part`'s subtables.

dimitri-yatsenko:
The term "subtable", as we have sometimes used it, is now replaced with "part relation". Let's also use "master relation" to indicate the relation containing the part relation.  We have also decided that part relations cannot have their own part relations.

So to restate the present issue: 
- Deleting from a part relation should not be possible without deleting matching tuples from its master relation. 

We have two options to address this issue:
1. prohibit deletes from a part relation that are not cascaded through its master relation 
2. delete from the master relation when a delete is issued on the part relation

Option 1 seems overly restrictive since a part relation may have dependencies not shared by its master and some cascading deletes may halt and force the user to break up a single delete command into several more specific commands.  If we pursue this solution, we need to make sure that the error message and the suggested course of action are clear.

Option 2 seems a bit aggressive since now deletes will cascade one level up the dependency hierarchy as well as down. This may be counterintuitive to some users.  This concern is partly addressed by displaying a summary of the data to be deleted before committing the delete.  Also option 2 may produce more complex queries that may be much slower than ones resulting from Option 1.

Which option should we choose?

eywalker:
I almost always prefer strong and convenient (but potentially destructive) solution, provided that we keep it informative. So if I have to choose, I'd prefer to use option 2. However, I also think this is a perfect opportunity for us to use config to alter between different behaviors, and offer both option 1 and option 2.

fabiansinz:
I prefer option 2 and I think we should offer to change the behavior in the config.

dimitri-yatsenko:
I support Option 2. However, making datajoint configurable between these two options would create undue documentation and testing burden. Let's implement Option 2 and document it well.

dimitri-yatsenko:
I relabeled this as a 'feature' because the current implementation acts reasonably: it fails to delete when attempting to delete from parts, bypassing the master. I just don't want panic to set in that cascading deletes are unsafe.

mspacek:
Hello! OK, so the term "subtable" is no longer used. It's crucial to update the documentation to reflect that. I just spent several hours trying to figure out how to define a subtable, including searching through the code, and finally ended up here.

[...]

PR #1106, opened 2023-09-02
Update set_password to work on MySQL 8

rly:
- Updated `set_password` to work on MySQL 8
- Added missing tests for `set_password` (not sure how to test different MySQL versions though)
- Updated `logger.warn` -> `logger.warning` because `logger.warn` is deprecated

PR #1158, opened 2024-04-24
Allow including master table in cascading delete

CBroz1:
#151 included a discussion of 'Option 2' in which the master would be deleted during a cascade, with a possible solution proposed in this PR. This feature would be particularly useful to the Spyglass team because our ['Merge tables'](https://lorenfranklab.github.io/spyglass/0.5/misc/merge_tables/) make use of foreign-key references in parts to version pipelines. 

Happy to incorporate feedback

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@
 - Added - Missing tests for set_password - PR [#1106](https://github.com/datajoint/datajoint-python/pull/1106)
 - Changed - Returning success count after the .populate() call - PR [#1050](https://github.com/datajoint/datajoint-python/pull/1050)
 - Fixed - `Autopopulate.populate` excludes `reserved` jobs in addition to `ignore` and `error` jobs 
-- Fixed - Issue [#1159]((https://github.com/datajoint/datajoint-python/pull/1159)  (cascading delete)  - PR [#1160](https://github.com/datajoint/datajoint-python/pull/1160)
+- Fixed - Issue [#1159](https://github.com/datajoint/datajoint-python/pull/1159) (cascading delete) - PR [#1160](https://github.com/datajoint/datajoint-python/pull/1160)
 - Fixed - `docker compose` commands in CI [#1164](https://github.com/datajoint/datajoint-python/pull/1164)
 - Changed - Default delete behavior now includes masters of part tables - PR [#1158](https://github.com/datajoint/datajoint-python/pull/1158)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 - Fixed - `Autopopulate.populate` excludes `reserved` jobs in addition to `ignore` and `error` jobs 
 - Fixed - Issue [#1159]((https://github.com/datajoint/datajoint-python/pull/1159)  (cascading delete)  - PR [#1160](https://github.com/datajoint/datajoint-python/pull/1160)
 - Fixed - `docker compose` commands in CI [#1164](https://github.com/datajoint/datajoint-python/pull/1164)
+- Changed - Default delete behavior now includes masters of part tables - PR [#1158](https://github.com/datajoint/datajoint-python/pull/1158)
 
 ### 0.14.1 -- Jun 02, 2023
 - Fixed - Fix altering a part table that uses the "master" keyword - PR [#991](https://github.com/datajoint/datajoint-python/pull/991)
--- a/datajoint/table.py
+++ b/datajoint/table.py
@@ -486,7 +486,7 @@ class Table(QueryExpression):
         transaction: bool = True,
         safemode: Union[bool, None] = None,
         force_parts: bool = False,
-        include_parts: bool = True,
+        force_masters: bool = True,
     ) -> int:
         """
         Deletes the contents of the table and its dependent tables, recursively.
--- a/datajoint/table.py
+++ b/datajoint/table.py
@@ -498,7 +498,7 @@ class Table(QueryExpression):
             safemode: If `True`, prohibit nested transactions and prompt to confirm. Default
                 is `dj.config['safemode']`.
             force_parts: Delete from parts even when not deleting from their masters.
-            include_parts: If `True`, include part/master pairs in the cascade.
+            force_masters: If `True`, include part/master pairs in the cascade.
                 Default is `True`.
 
         Returns:
@@ -573,7 +573,7 @@ class Table(QueryExpression):
 
                     master_name = get_master(child.full_table_name)
                     if (
-                        include_parts
+                        force_masters
                         and master_name
                         and master_name != table.full_table_name
                         and master_name not in visited_masters
--- a/datajoint/table.py
+++ b/datajoint/table.py
@@ -512,6 +512,12 @@ class Table(QueryExpression):
         deleted = set()
         visited_masters = set()
 
+        if force_masters and not force_parts:
+            logger.warn(
+                "force_masters=True implies force_parts=True. "
+                + "Ignoring force_parts=False."
+            )  # No need to reset, as force_masters will cascade to parts.
+
         def cascade(table):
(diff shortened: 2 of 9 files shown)


lfrank:
@dimitri-yatsenko @ttngu207 Hi Dmitri and Thinh. I just wanted to check in on this pull request and another one (https://github.com/datajoint/datajoint-python/pull/1160). It would really help us to get these into a new release if that's possible.

ethho:
> it would be helpful to know for our (Spyglass) planning purposes whether this PR is likely to get reviewed/merged in the near future.

Hi @edeno, I'll review this by end of week. Thanks all for the PR.

@ttngu207 @dimitri-yatsenko @lfrank Do we need to publish a new release? Or is it sufficient to merge to `main`?

edeno:
It would be greatly helpful for us to have a release.

datajoint/table.py
```diff
@@ -486,6 +486,7 @@ def delete(
         transaction: bool = True,
         safemode: Union[bool, None] = None,
         force_parts: bool = False,
+        include_parts: bool = True,
```
ttngu207, line 489: I'd suggest a different argument name. It's not super obvious what `include_parts` means, as ones must expect that cascade delete will also delete the parts.

Here, we mean to include also the "master" table(s) of the part table(s) being deleted.

Suggesting to rename to something like `force_masters` (for consistency with the arg `force_parts`)
Or perhaps consider using an environment variable or configuration (in `dj.config`) to control for this behavior (similar to `DJ_SUPPORT_FILEPATH_MANAGEMENT`)

dimitri-yatsenko: I would prefer to have a single consistent behavior, not making it configurable.

ethho:
@CBroz1 I would recommend checking out changes from #1164 so that tests run.

CBroz1:
> @CBroz1 I would recommend checking out changes from #1164 so that tests run.

Thanks @ethho! If you plan to merge that first, I'll keep an eye out and fetch upstream when it does

ethho:
> If you plan to merge that first, I'll keep an eye out and fetch upstream when it does

@CBroz1 #1164 was merged to `master`.

CBroz1:
Thanks, @ethho. I fetched from upstream

ethho (changes requested):
#### Summary

Looks good overall. I request the following changes:

1. Update documentation to reflect the new behavior. It is sufficient to update [this section on deletes](https://github.com/datajoint/datajoint-python/blob/685b42ccbc85c4b03a347311e272539cfc368f5e/docs/src/manipulation/delete.md#L23).
2. Rename kwarg to `force_masters` as discussed above
3. Expand regression tests to cover edge cases

#### Edge Cases

The logic changes in `delete` are sensible and address the requirements laid out in #151.

While I can't think of any breaking edge cases, @ttngu207 and I think that we should include some of these cases in regression tests.
Notably, we should test what happens when a part table is visited more than once in the cascade.
The new tests cover the case where the master table is visited twice, and we save some work by skipping `visited_masters`.
But what if a part table shares a master with its children or grandchildren?
`cascade` will visit the downstream part table twice.
This is harmless currently because `quick_delete` is idempotent, but it's worth testing in case delete behavior changes. For example, we could add a table to `schema_simple`:

```python
class I(dj.Part):
    definition = """ # test no additional fk reference
    -> E
    id_i :int
    ---
    -> H
    """
```

#### Kwarg Interaction

No change requested, but note that `my_part_table.delete(force_parts=False, force_masters=True)` will delete the master table, even though `force_parts=False`. This behavior makes sense to me: `force_masters=True` will always supercede `force_parts`.

CBroz1:
Thanks for your review @ethho!

> But what if a part table shares a master with its children or grandchildren?

Sure, happy to test this case. It looks like there are issues with a part being a child of a another part of the same master, given the tests failing in my latest commit. I'll spend some time tomorrow looking at some alternatives, but let me know if you have other suggestions

> `my_part_table.delete(force_parts=False, force_masters=True)` will delete the master table, even though `force_parts=False`. 

Given that this is currently the default behavior, should `force_parts=True` turn off `force_masters`, or log a warning? Otherwise, in order to get the previous `force_parts=True` behavior, users will need to set `force_parts=True, force_masters=False`.

ethho:
> Given that this is currently the default behavior, should force_parts=True turn off force_masters, or log a warning?

@CBroz1 Based on my reading, `force_masters=True` always implies `force_parts=True`, since [master is always in deleted in this expression](https://github.com/datajoint/datajoint-python/blob/685b42ccbc85c4b03a347311e272539cfc368f5e/datajoint/table.py#L609) when `force_masters=True`. `force_masters=False` restores the behavior before this PR, where the value of `force_parts` is relevant.

Then we should document the behavior well and keep the logic as is, without raising a warning. A warning would get very annoying as the default behavior. One option that I don't endorse is: change default behavior to `force_masters=True, force_parts=True`, but this would result in surprising behavior for users accustomed to `delete()` who refactor to `delete(force_masters=False)` in the new version.

CBroz1:
Thanks @ethho, sounds good. I made note of this in the changelog. 

If we could get this into a `14.2` release, my team would really benefit both from this PR as well as #1106. Happy to make adjustments here if that makes sense, but I understand if there are other features you're targeting for inclusion in `14.2`

ethho:
> If we could get this into a 14.2 release, my team would really benefit both from this PR

In my opinion, this change is a large enough to be in the major release `0.15` instead of `0.14.2`. The preferred approach for us would be a pre-release `0.15.dev0`. Would this work for you? Note that `datajoint` will continue to point to `datajoint==0.14.1` as the latest, but your team could explicitly set `datajoint==0.15.dev0`.

CBroz1:
Sure, that works for me, thanks! Should I make any edits to the changelog to reflect the pre-release?

ethho (approved):

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
 - Fixed - `Autopopulate.populate` excludes `reserved` jobs in addition to `ignore` and `error` jobs 
 - Fixed - Issue [#1159]((https://github.com/datajoint/datajoint-python/pull/1159)  (cascading delete)  - PR [#1160](https://github.com/datajoint/datajoint-python/pull/1160)
```
ethho, line 14:
```suggestion
- Fixed - Issue [#1159](https://github.com/datajoint/datajoint-python/pull/1159) (cascading delete) - PR [#1160](https://github.com/datajoint/datajoint-python/pull/1160)
```

@CBroz1 Just this change. I need to pivot to other tasks today but will work on publishing `0.15.dev0` tomorrow.

CBroz1: > publishing `0.15.dev0` tomorrow.

Thanks so much!

ethho:
For anyone urgently blocked on `0.15.dev0`, note that you can always `pip install git+`: https://stackoverflow.com/a/20101940

tests/test_cascading_delete.py
```diff
@@ -110,11 +111,24 @@ def test_delete_master(schema_simp_pop):
...
+def test_delete_parts(schema_simp_pop):
+    """test issue #151"""
+    Profile().populate_random()
+    Website().delete(force_masters=True)
+
+
+def test_delete_parts_complex(schema_simp_pop):
+    """test issue #151 with complex master/part. PR #1158."""
+    prev_len = len(G())
+    (A() & "id_a=1").delete()
```
ethho, line 130:
```suggestion
    (A() & "id_a=1").delete(force_masters=True)
```

ethho (changes requested):

ethho (approved):

ethho:
@CBroz1 @edeno @lfrank This PR was included in [release 0.4.2](https://pypi.org/project/datajoint/0.14.2/) built by [CI](https://github.com/datajoint/datajoint-python/actions/runs/10514098880/job/29131408143).

PR #1163, opened 2024-07-17
Bumped minimum version to 3.8

### LINKED ISSUE #1162: Remove Python 3.7 support
Hello there,
As Python 3.7 has been end-of-life since [2023-June](https://devguide.python.org/versions/), we can update the minimum supported version of Python to 3.8 which would allow the removal of the package pin [here](https://github.com/datajoint/datajoint-python/blob/685b42ccbc85c4b03a347311e272539cfc368f5e/requirements.txt#L7), which prevents users from using Python 3.12 as the pinned package uses a deprecated feature since 3.10 and a removed feature since 3.12.

The minimum version check [here](https://github.com/datajoint/datajoint-python/blob/685b42ccbc85c4b03a347311e272539cfc368f5e/setup.py#L6) would also have to be updated.

tabedzki:
To close #1162 
Happy to make other contributions as necessary to upgrade the minimum version.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,8 +12,6 @@
 - Changed - Returning success count after the .populate() call - PR [#1050](https://github.com/datajoint/datajoint-python/pull/1050)
 - Fixed - `Autopopulate.populate` excludes `reserved` jobs in addition to `ignore` and `error` jobs 
 - Fixed - Issue [#1159]((https://github.com/datajoint/datajoint-python/pull/1159)  (cascading delete)  - PR [#1160](https://github.com/datajoint/datajoint-python/pull/1160)
-
-### 0.14.2 -- July 29, 2024
 - Changed - Minimum Python version for Datajoint-Python is now 3.8 PR #1163
 
 ### 0.14.1 -- Jun 02, 2023
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,9 @@
 - Fixed - `Autopopulate.populate` excludes `reserved` jobs in addition to `ignore` and `error` jobs 
 - Fixed - Issue [#1159]((https://github.com/datajoint/datajoint-python/pull/1159)  (cascading delete)  - PR [#1160](https://github.com/datajoint/datajoint-python/pull/1160)
 
+### 0.14.2 -- July 29, 2024
+- Changed - Minimum Python version for Datajoint-Python is now 3.7 PR #1163
+
 ### 0.14.1 -- Jun 02, 2023
 - Fixed - Fix altering a part table that uses the "master" keyword - PR [#991](https://github.com/datajoint/datajoint-python/pull/991)
 - Fixed - `.ipynb` output in tutorials is not visible in dark mode ([#1078](https://github.com/datajoint/datajoint-python/issues/1078)) PR [#1080](https://github.com/datajoint/datajoint-python/pull/1080)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@
 - Fixed - Issue [#1159]((https://github.com/datajoint/datajoint-python/pull/1159)  (cascading delete)  - PR [#1160](https://github.com/datajoint/datajoint-python/pull/1160)
 
 ### 0.14.2 -- July 29, 2024
-- Changed - Minimum Python version for Datajoint-Python is now 3.7 PR #1163
+- Changed - Minimum Python version for Datajoint-Python is now 3.8 PR #1163
 
 ### 0.14.1 -- Jun 02, 2023
 - Fixed - Fix altering a part table that uses the "master" keyword - PR [#991](https://github.com/datajoint/datajoint-python/pull/991)
--- a/.github/workflows/development.yaml
+++ b/.github/workflows/development.yaml
@@ -28,7 +28,7 @@ jobs:
       DOCKER_CLIENT_TIMEOUT: "120"
       COMPOSE_HTTP_TIMEOUT: "120"
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - name: Validate version and release notes
         run: |
           DJ_VERSION=$(grep -oP '\d+\.\d+\.\d+' datajoint/version.py)
@@ -58,8 +58,6 @@ jobs:
         py_ver: ["3.9"]
         mysql_ver: ["8.0", "5.7"]
         include:
-          - py_ver: "3.12"
-            mysql_ver: "8.0"
           - py_ver: "3.11"
             mysql_ver: "8.0"
           - py_ver: "3.10"
@@ -67,9 +65,9 @@ jobs:
           - py_ver: "3.8"
             mysql_ver: "5.7"
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - name: Set up Python ${{matrix.py_ver}}
-        uses: actions/setup-python@v4
+        uses: actions/setup-python@v5
         with:
           python-version: ${{matrix.py_ver}}
       - name: Install dependencies
@@ -96,9 +94,9 @@ jobs:
       matrix:
         py_ver: ["3.11"]
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - name: Set up Python ${{matrix.py_ver}}
-        uses: actions/setup-python@v4
+        uses: actions/setup-python@v5
         with:
           python-version: ${{matrix.py_ver}}
       - name: Install dependencies
@@ -119,7 +117,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
       - name: Codespell
         uses: codespell-project/actions-codespell@v2
   publish-docs:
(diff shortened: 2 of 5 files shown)


tabedzki:
@dimitri-yatsenko  @ethho  Can we update the Python versioning to drop support for end-of-life versions of Python (3.7)?

ethho:
Thanks for the PR @tabedzki! Agreed that we should drop 3.7 in favor of 3.12.

Can you also change the CI so that we test Python3.12 + MySQL 8.0, and drop the 3.7 tests? I can't make the change on your fork:

https://github.com/datajoint/datajoint-python/blob/685b42ccbc85c4b03a347311e272539cfc368f5e/.github/workflows/development.yaml#L60-L68

dimitri-yatsenko:
Also please update the CHANGELOG.md

tabedzki:
Done! Let me know if there are any other changes I should make. 
Also, will the pip package automatically update after this PR or is there a manual step involved?

ethho:
[CI job](https://github.com/datajoint/datajoint-python/actions/runs/10146647425/job/28055221027?pr=1163) failed because there is no Docker image `datajoint/djtest:py3.12-alpine`.

We need to add 3.12 and build the image from the [datajoint/djbase-docker](https://github.com/datajoint/djbase-docker) repository (has CI, looks like it's never been used), similar to the addition of 3.11 in [this PR](https://github.com/datajoint/djbase-docker/pull/34).

ethho:
Depends on https://github.com/datajoint/djbase-docker/pull/36

ethho:
To summarize: we need to add python 3.12 to every image in the [dependency tree](https://datajoint.github.io/miniconda3-docker/), in this order:

- `datajoint/miniconda3`. PR: https://github.com/datajoint/miniconda3-docker/pull/50
- `datajoint/djbase`. PR: https://github.com/datajoint/djbase-docker/pull/36
- `datajoint/djtest`. PR: https://github.com/datajoint/djtest-docker/pull/7
- Re-run CI in this PR

tabedzki:
@ethho Since Python 3.13 will be released on October 1st of this year, is there any prep work we can do now or should this be broached in October? Python 3.8 also goes end of life that same month. 

[Python support schedule](https://devguide.python.org/versions/)

ethho:
> Since Python 3.13 will be released on October 1st of this year, is there any prep work we can do now or should this be broached in October

We install Python in the image via conda, so let's wait until a stable 3.13 release is available via conda channels before updating.

This whole dependency tree of Docker images is cumbersome; we should simplify it by maintaining a minimal set of images either in this repository or `datajoint/djbase`.

tabedzki:
@ethho  

I took a look at the datajoint/miniconda3-docker and datajoint/dj-base repositories to get an understanding of the workflow. I saw that the code is tested on both alpine and distros and packaged with `conda`. 

- Since conda is used to install python, would you be open to using the github action [setup-python](https://github.com/marketplace/actions/setup-python)? 
- Also, would you be open to using the github action [setup alpine](https://github.com/marketplace/actions/setup-alpine-linux-environment) as an alternative to the alpine docker images? 

Please let me know if there are any consideration details that I have may missed in these suggestions (i.e., if conda installs more than just python/if docker images are used elsewhere for testing and debugging purposes).

ethho:
> * Since conda is used to install python, would you be open to using the github action [setup-python](https://github.com/marketplace/actions/setup-python)?

@tabedzki AFAIK, `setup-python` GHA cannot be used to set up python in a Docker context, which is what we need here. I've also never used it with conda before.

> if docker images are used elsewhere for testing and debugging purposes

You are correct. Unfortunately, images such as `datajoint/djbase` and `datajoint/miniconda3` are used in many of the DataJoint codebases, which is primarily why unwinding this tech debt is expensive.

IMO the best approach now is to manually build a 3.12 miniconda image, with the Dockerfile (pseudocode):

```
FROM datajoint/miniconda3:py312
RUN conda install python=3.12
```

It doesn't make sense to maintain separate miniconda images for different python versions anyway.

tabedzki:
@ethho thanks for giving me a fuller picture. I didn't realize how heavily dependent the system was on those docker images. Therefore, moving away from the conda installation and using the github actions, in the current setup, doesn't make sense. 

Looking forward to the Python 3.12 support when it happens. Thank you for your effort on this.

tabedzki:
@ethho since there is more work to add 3.12 than there is to remove 3.7 support, I've removed aspects of the PR that would be more burdensome and only focused on removing things that tied the repo to <3.8.

Additionally, I've updated the tags used for a couple of github actions since I noticed they were out of the date.

I can open a separate issue that PR that would focus on 3.12 support.

CHANGELOG.md
```diff
@@ -13,6 +13,9 @@
 - Fixed - `Autopopulate.populate` excludes `reserved` jobs in addition to `ignore` and `error` jobs 
 - Fixed - Issue [#1159]((https://github.com/datajoint/datajoint-python/pull/1159)  (cascading delete)  - PR [#1160](https://github.com/datajoint/datajoint-python/pull/1160)
 
+### 0.14.2 -- July 29, 2024
```
dimitri-yatsenko, line 16: The changes above have already been implemented, so your change will have to be rolled into the next release.

tabedzki: Hi @dimitri-yatsenko 

> The changes above have already been implemented, so your change will have to be rolled into the next release.

Could you clarify what you meant by "these changes have already been implemented"? Where was this updated/implemented? I looked and didn't see any changes elsewhere according to the public branches of this repo nor in the public forks of this repository, though I may have missed it. 

Thanks in advance

dimitri-yatsenko: The fixes listed in the CHANGELOG are queued for the next release. This issue will be added to that list rather than becoming its own release.

tabedzki: Ahh okay. I can fix that. Thanks for letting me know

ethho:
> I can open a separate issue that PR that would focus on 3.12 support.

@tabedzki Please do make a separate issue for this, thanks! The docker builds need some work; I'll make another ticket to remove dependency on these images as noted in [my previous comment](https://github.com/datajoint/datajoint-python/pull/1163#issuecomment-2256818665).

CI runs [such as this one](https://github.com/datajoint/datajoint-python/actions/runs/10222188835/job/28413114065?pr=1163) now failing due to use of `docker-compose` instead of subcommand `docker compose`. Plan to fix in #1164, merge to master, then merge master to this branch.

tabedzki:
@ethho I just learned about [SPEC 0 — Minimum Supported Dependencies](https://scientific-python.org/specs/spec-0000/), a coordinated effort endorsed by the major scientific packages (SciPy, NumPy, scikit-image, matplotlib, etc) to adopt a common time-based policy for dropping dependencies. Based off this recommendations, the oldest Python version that should be supported is Python 3.10, which would be dropped according to SPEC 0. 

Likewise, `numpy` itself has officially dropped support for Python 3.9 as noted [here](https://numpy.org/neps/nep-0029-deprecation_policy.html) and [SciPy takes guidance from Numpy's schedule](https://docs.scipy.org/doc/scipy/dev/toolchain.html#python-versions):
```
On Apr 05, 2024 drop support for Python 3.9 (initially released on Oct 05, 2020)
On Jun 22, 2024 drop support for NumPy 1.23 (initially released on Jun 22, 2022)
```

As we are already updating which versions of Python `datajoint` supports, I think we should consider which version of Python `datajoint` officially support. If the minimum version was updated to Python 3.10, there would be fewer versions to support and hence less testing and possible better coordination between package dependencies.

tabedzki:
@ethho  and @dimitri-yatsenko  What are your thoughts on adopting the same support timetable as other scientific Python packages (e.g., numpy, scipy, etc), or adopting the timetable from Scientific Python?

- **Numpy/Scipy's current timetable**: 
  - Now: Bump minimum supported version to 3.10 
  - Apr 04, 2025:  Bump minimum supported version to 3.11
- **Scientific Python's current timetable**:
   - Now: Bump minimum supported version to 3.10 
   - 2024 - Quarter 4: Bump minimum supported version to 3.11

ethho:
> What are your thoughts on adopting the same support timetable as other scientific Python packages (e.g., numpy, scipy, etc), or adopting the timetable from Scientific Python?

@tabedzki Thanks for bringing this to our attention. I see the motivation here: SciPy has historically had a tough time supporting many Python versions.

In my opinion, datajoint should continue to support older versions of Python as long as the maintenance cost is minimal. We should deprecate older versions on a case-by-case basis, when we encounter instances where the maintenance cost is non-trivial. This ticket is one example; it seems to be caused by a breaking networkx change.

The reasoning is that datajoint occupies a different space in the scientific Python stack than packages like numpy, scipy, and pandas. New features in numpy / scipy are tightly coupled to Python version and target arch. datajoint is not tightly coupled to Python version, and therefore the maintenance cost of supporting legacy versions is relatively small. In other words, new datajoint features are unlikely to use specific Python language features. datajoint is, however, tightly coupled to the database backend. This is why we adopted this approach and deprecated MySQL 5.7 and only test for MySQL 8.

cc: @ttngu207 @dimitri-yatsenko

tabedzki:
@ethho  thank you for the in-depth explanation. I can understand and appreciate the reasoning. 

@ethho  and @dimitri-yatsenko 
Since I refocused this PR to just bumping to 3.8 by removing references to Python 3.12, and because this PR passes all the current CI checks, can we merge this PR?

PR #1171, opened 2024-08-21
Solve database lock during long computations by breaking up the make callback into three parts

### LINKED ISSUE #1170: Long make calls lock table metadata
### Description

A client locks table metadata for the entire duration of a `make` function call. When other clients attempt to `drop` or `declare` child tables, the call is blocked until the first client finishes `make`. This approach scales poorly with number of clients and number of child tables.

### Reproducibility

Include:
- OS: Any
- Python Version: Any
- MySQL Version: Any
- MySQL Deployment Strategy: Any
- DataJoint Version: 0.14.1
- Minimum steps required to reproduce:
    - See the test case presented https://github.com/LorenFrankLab/spyglass/issues/1030

### Proposed Solution

As an alternative to writing a `Computed.make` function, allow user to write three functions:

1. `make_fetch` for reading inputs
2. `make_compute`, which is not run in a transaction, and is passed the return value of `make_fetch`
3. `make_insert`, which inserts computed values using the same transaction semantics as `make`.

In pseudocode, these three functions will be used in the following routine as such:

```python
if hasattr(table, "make"):
    return make()
else:
    assert hasattr(table, "make_fetch")
    assert hasattr(table, "make_compute")
    assert hasattr(table, "make_insert")
    input = make_fetch()
    conn.disconnect() # I assume this disconnect step is to ensure that make_compute cannot insert?
    result = make_compute(input)
    tx = conn.start_transaction()
    input2 = make_fetch()
    if hash(serialize(input2)) == hash(serialize(input)):
        result = make_insert(result)
        tx.commit()
        return result
    else:
        print("ERROR: inputs have changed")
        tx.abort()
        return None
```

#### Related Issues

- https://github.com/LorenFrankLab/spyglass/issues/1030
- https://github.com/LorenFrankLab/spyglass/pull/1067

---

cc: @dimitri-yatsenko @ttngu207 @CBroz1 @samuelbray32 @peabody124

ethho:
Fixes #1170, implementing the [proposed solution](https://github.com/datajoint/datajoint-python/issues/1170#issue-2478964792) to long-running `make` calls that lock table metadata.

## To Do

- [ ] Use `threading` library to write concurrency pytest
- [ ] Basic regression test that uses three-part `make`

### DIFF
--- a/tests/schema_simple.py
+++ b/tests/schema_simple.py
@@ -92,7 +92,33 @@ class D(dj.Computed):
         # make reference to a random tuple from L
         random.seed(str(key))
         lookup = list(L().fetch("KEY"))
-        self.insert(dict(key, id_d=i, **random.choice(lookup)) for i in range(4))
+        self.insert(dict(key, id_d=i, **random.choice(lookup))
+                    for i in range(4))
+
+
+class N(dj.Computed):
+    definition = """
+    # test for three part make function
+    -> A
+    id_d :int
+    ---
+    -> L
+    """
+
+    def make_fetch(self, key):
+        # make reference to a random tuple from L
+        lookup = list(L().fetch("KEY"))
+        inputs = key, lookup
+        return inputs
+
+    def make_compute(self, inputs):
+        key, lookup = inputs
+        random.seed(str(key))
+        values = [dict(key, id_d=i, **random.choice(lookup)) for i in range(4)]
+        return values
+
+    def make_insert(self, values):
+        self.insert(values)
 
 
 class E(dj.Computed):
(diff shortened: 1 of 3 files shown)


tests/schema_simple.py
```diff
@@ -92,7 +92,33 @@ def _make_tuples(self, key):
...
+    -> A
+    id_d :int
+    ---
+    -> L
+    """
+
+    def make_fetch(self, key):
+        # make reference to a random tuple from L
+        lookup = list(L().fetch("KEY"))
+        inputs = key, lookup
```
dimitri-yatsenko, line 111: You shouldn't have to return the `key` since its provided by `populate`

dimitri-yatsenko:
This will be superseded by #1204 - a simpler solution

PR #1174, opened 2024-09-05
Fix CD docs build

ethho:
Fix documentation builds that are failing, such as https://github.com/datajoint/datajoint-python/actions/runs/10514098880/job/29131407848

### DIFF
--- a/docs/mkdocs.yaml
+++ b/docs/mkdocs.yaml
@@ -111,6 +111,9 @@ plugins:
       default_handler: python
       handlers:
         python:
+          paths:
+            - "."
+            - /main/
           options:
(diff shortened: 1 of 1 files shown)


ethho:
Was able to reproduce issue from CD job with:

```bash
MODE=BUILD PACKAGE=datajoint UPSTREAM_REPO=https://github.com/datajoint/datajoint-python.git HOST_UID=$(id -u) docker compose -f docs/docker-compose.yaml up --exit-code-from docs --build
```

on commit c3577724da6a.

dimitri-yatsenko (approved):

PR #1184, opened 2024-09-13
Fix #1103, #1057

dimitri-yatsenko:
* Fix #1103 - Dependencies are now loaded in `table.parts()`
* Fix #1057 - Nodes are now topologically sorted, regardless of parts.

PR #1185, opened 2024-09-14
Remove old tests

dimitri-yatsenko:
We can find them if we need them. Do this after merging #1184 .


ethho (approved):
TODO: remove references to `tests_old` in `develop.md`

PR #1180, opened 2024-09-12
hidden attributes continued (#1091)

ethho:
Patches #1091 so that it is compatible with new testing framework

PR #1183, opened 2024-09-13
Refactor Docker image and package builds

ethho:
Dependency on datajoint/djbase as a base image makes it difficult to maintain the datajoint/datajoint image. For example, adding support for Python 3.12 #1165 requires rebuild of the entire ["image hierarchy"](https://datajoint.github.io/miniconda3-docker/), for which many of the builds are not reproducible. The task for this PR is to refactor the datajoint/datajoint Docker image so that it is lightweight, easy to maintain, and does not depend on the image hierarchy. Includes changes from #1187

PR #1188, opened 2024-09-22
Disable hidden timestamp attribute by default

ethho:
Adds configuration flag `datajoint.config["add_hidden_timestamp"]`, set to `False` by default, which toggles automatic addition of the timestamp hidden attribute. Patches #1091, #1180.


ethho:
CI runs blocked by either https://github.com/datajoint/nginx-docker/pull/56 or #1183.

datajoint/declare.py
```diff
@@ -311,17 +312,18 @@ def declare(full_table_name, definition, context):
...
 
-    metadata_attr_sql = [
-        "`_{full_table_name}_timestamp` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP"
-    ]
-    attribute_sql.extend(
-        attr.format(
-            full_table_name=sha1(
-                full_table_name.replace("`", "").encode("utf-8")
-            ).hexdigest()
+    if config.get("enable_hidden_attributes", False):
```
dimitri-yatsenko, line 315: How about changing to `add_hidden_timestamp` to make this feature more specific. Hidden attributes will always be enabled. But the default timestamp will be configurable.

dimitri-yatsenko (changes requested):
make the setting name more specific to implicit timestamps.

CHANGELOG.md
```diff
@@ -1,14 +1,15 @@
 ## Release notes
 
 ### 0.14.3 -- Sep 20, 2024
```
dimitri-yatsenko, line 3: Sep 23 perhaps?

dimitri-yatsenko (approved):

PR #1273, opened 2025-09-16
Impr/modernize pre commit

d-v-b:
this PR modernizes the pre-commit workflow via the following changes:
- uses ruff / ruff fmt instead of flake8/black for a huge speed increase
- moves configuration of linting to pyproject.toml where necessary
- automatically applies code formatting as part of the pre-commit hook.

Ruff is flagging a pretty large volume of code here. it will take some time to fix those problems, but it's ultimately worth it. not sure if this should be done in this PR or a later effort.


pyproject.toml
```diff
@@ -85,21 +85,57 @@ test = [
...
   "pytest",
   "pytest-cov",
 ]
 
-[tool.isort]
-profile = "black"
+[tool.ruff]
+# Equivalent to flake8 configuration
+line-length = 127
+target-version = "py39"
```
dimitri-yatsenko: Should we change to py310?

d-v-b:
I am fixing some test failures locally, I will ping you when everything is green

PR #1274, opened 2025-09-16
update test workflow to use src layout

d-v-b:
the move to a src layout did not extend to the github workflows. this PR fixes that.


d-v-b:
@dimitri-yatsenko since this is on the pre/v2.0 branch is it OK if I go ahead and merge this kind of thing?

PR #1279, opened 2025-11-10
Chore/dev env fixes

d-v-b:
Applies more fixes to the developer environment. I will add explanatory comments to new / altered files.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -50,7 +50,6 @@ repos:
       - --max-complexity=62
       - --max-line-length=127
       - --statistics
-      - --per-file-ignores=datajoint/diagram.py:C901
     files: src/ # a lot of files in tests are not compliant
 - repo: https://github.com/rhysd/actionlint
   rev: v1.7.7


.devcontainer/devcontainer.json
```diff
@@ -1,56 +1,7 @@
...
-			]
-		}
-	},
-	"remoteEnv": {
-		"LOCAL_WORKSPACE_FOLDER": "${localWorkspaceFolder}"
-	}
-	// Uncomment to connect as an existing user other than the container default. More info: https://aka.ms/dev-containers-non-root.
-	// "remoteUser": "devcontainer"
-}
+  "image": "mcr.microsoft.com/devcontainers/typescript-node:0-18",
```
d-v-b: we just need pixi and docker now

d-v-b: nodejs is in here for LLM tooling, we can remove if that's not needed

activate.sh
```diff
@@ -0,0 +1,4 @@
+#! /usr/bin/bash
+# This script registers dot plugins so that we can use graphviz 
+# to write png images
+dot -c
```
d-v-b: the `dot` tool used by graphviz uses a post-installation step to register different file format backends. this script performs that step. without this, the graphviz-related tests will fail in a fresh environment.

pyproject.toml
```diff
@@ -22,7 +22,7 @@ dependencies = [
   "urllib3",
   "setuptools",
 ]
-requires-python = ">=3.9,<4.0"
+requires-python = ">=3.9,<3.14"
```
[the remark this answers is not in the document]
dimitri-yatsenko: 3.9 has reached its end of life.

PR #1281, opened 2025-12-17
switch settings to use pydantic-settings

d-v-b:
This replaces the settings management with a pydantic-settings-based approach. For backwards compatibility, old APIs are added on top of the new pydantic-based foundation. Opening as a draft while I iterate on the implementation.


d-v-b:
this is ready for review. I did not change how the config is used in the codebase. I recommend deprecating the mutable-mapping approach (`config["database.password"] = new_password`) because this is harder to make type-safe and also doesn't support autocomplete. A simpler API is probably something like `config.database.password = new_password` or `config.database.set_password(new_password)`.

PR #1294, opened 2025-12-25
Refine primary key rules for relational operators

dimitri-yatsenko:
## Summary

This PR refines the rules for determining primary keys in relational operators, building on the semantic matching foundation. The key insight is that join primary keys should be derived from **functional dependencies** between operands, not simply from the union of their primary keys.

## Primary Key Rules for Join

The join operator now determines its primary key based on functional dependency analysis:

| Condition | Primary Key | Attribute Order |
|-----------|-------------|-----------------|
| A → B | PK(A) | A's attributes first |
| B → A (and not A → B) | PK(B) | B's attributes first |
| Neither | PK(A) ∪ PK(B) | A's attributes first |

Where **A → B** means: every attribute in PK(B) is in A.

### Example
```
Session: session_id*, date
Trial: session_id*, trial_num*, response_time

Session * Trial → PK = {session_id} (Session → Trial)
Trial * Session → PK = {session_id} (Session → Trial, so PK(Session) used)
```

### Left Join Constraint

For left joins (`A.join(B, left=True)`), the functional dependency **A → B is required**.

**Why this constraint exists:**

In a left join, all rows from A are retained even if there's no matching row in B. For unmatched rows, B's attributes are NULL. This creates a problem for primary key validity:

| Scenario | PK by inner join rule | Left join problem |
|----------|----------------------|-------------------|
| A → B | PK(A) | ✅ Safe — A's attrs always present |
| B → A | PK(B) | ❌ B's PK attrs could be NULL |
| Neither | PK(A) ∪ PK(B) | ❌ B's PK attrs could be NULL |

### Bypassing the Left Join Constraint

For special cases where the user takes responsibility for handling the potentially invalid primary key, the constraint can be bypassed using `allow_invalid_primary_key=True`:

```python
# Normally blocked - A → B not satisfied
A.join(B, left=True)  # Error

# Bypass the constraint - user takes responsibility
A.join(B, left=True, allow_invalid_primary_key=True)  # Allowed, PK = PK(A) ∪ PK(B)
```

When bypassed, the resulting primary key is the union of both operands' primary keys. The user must ensure that subsequent operations (such as `GROUP BY` or projection) establish a valid primary key.

This mechanism is used internally by `aggr` with `keep_all_rows=True`, which resets the primary key via the `GROUP BY` clause.

## Aggregation Exception
A.aggr(B, keep_all_rows=True) uses a left join internally but requires B → A (opposite of the left join constraint). This works because:

- Aggregation requires B → A so B can be grouped by A's primary key
- The intermediate left join would have an invalid PK
- Aggregation internally allows this invalid PK (`via _allow_invalid_primary_key`)
- The GROUP BY PK(A) clause resets the primary key to PK(A)
- The final result has valid, non-NULL primary keys

Note: The semantic check (homologous namesake validation) is still performed for aggregation. Only the PK validity constraint is bypassed.

### DIFF
--- a/docs/src/design/semantic-matching-spec.md
+++ b/docs/src/design/semantic-matching-spec.md
@@ -341,6 +341,22 @@ The following attributes from the right operand's primary key are not determined
 the left operand: ['z']. Use an inner join or restructure the query.
 ```
 
+### Bypassing the Left Join Constraint
+
+For special cases where the user takes responsibility for handling the potentially invalid primary key, the constraint can be bypassed using `allow_invalid_primary_key=True`:
+
+```python
+# Normally blocked - B does not determine A
+A.join(B, left=True)  # Error: A → B not satisfied
+
+# Bypass the constraint - user takes responsibility
+A.join(B, left=True, allow_invalid_primary_key=True)  # Allowed, PK = PK(A) ∪ PK(B)
+```
+
+When bypassed, the resulting primary key is the union of both operands' primary keys (PK(A) ∪ PK(B)). The user must ensure that subsequent operations (such as `GROUP BY` or projection) establish a valid primary key.
+
+This mechanism is used internally by aggregation (`aggr`) with `keep_all_rows=True`, which resets the primary key via the `GROUP BY` clause.
+
 ### Aggregation Exception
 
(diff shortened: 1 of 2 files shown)


dimitri-yatsenko:
Reimplemented in #1302

PR #1295, opened 2025-12-25
Fix 1276 - pass kwargs to Part Delete

### LINKED ISSUE #1276: Can't pass `super.delete` kwargs to `Part.delete`
https://github.com/datajoint/datajoint-python/blob/63ebc380ecdd1ba1b0cff02f9927fe2666a59e24/datajoint/user_tables.py#L222

- `Part.delete` does not allow pass-through of kwargs to the `super.delete` call. 
- This prevents some functionality
    - e.g. I want to wrap several operations including `Part.delete` in a single transaction. Since I can't pass `transaction=False` to `Part.delete` (I think) this is not possible

dimitri-yatsenko:
Fix #1276


dimitri-yatsenko:
Will try a different approach

PR #1299, opened 2026-01-01
Merge upgraded Type Management System

dimitri-yatsenko:


dimitri-yatsenko:
Will try a different approach

PR #1309, opened 2026-01-07
tests: skip minio-dependent tests when minio is not installed

sarpit2907:
This change replaces the hard dependency on `minio` during test collection
with `pytest.importorskip`, so tests are skipped gracefully when the optional
dependency is not installed.

This improves the contributor experience and prevents pytest collection
failures due to missing optional test dependencies.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -6,9 +6,9 @@ from pathlib import Path
 from typing import Dict, List
 
 import certifi
-import minio
 import networkx as nx
 import pytest
+minio = pytest.importorskip("minio")
 import urllib3
 from packaging import version
 


dimitri-yatsenko:
Thanks for raising this! DataJoint 2.0 (on the `docs-2.0-migration` branch) addresses test dependency isolation more comprehensively:

## Current 2.0 Approach

1. **Automatic test markers**: Tests are auto-marked with `requires_mysql` or `requires_minio` based on fixture usage—no manual decorators needed.

2. **Lazy container imports**: The `testcontainers` imports are inside fixture functions, not at module level:
   ```python
   @pytest.fixture(scope="session")
   def minio_container():
       if USE_EXTERNAL_CONTAINERS:
           yield None
           return
       from testcontainers.minio import MinioContainer  # Import only when needed
       ...
   ```

3. **Run unit tests without Docker**:
   ```bash
   pytest -m "not requires_mysql"
   ```

4. **External container support** for CI or local docker-compose:
   ```bash
   DJ_USE_EXTERNAL_CONTAINERS=1 pytest
   ```

## Additional Improvement

Your suggestion to use `pytest.importorskip` would complement this nicely for the edge case where someone runs integration tests without `testcontainers` installed. We could add:

```python
@pytest.fixture(scope="session")
def minio_container():
    if USE_EXTERNAL_CONTAINERS:
        yield None
        return
    
    MinioContainer = pytest.importorskip("testcontainers.minio").MinioContainer
    ...
```

This would gracefully skip instead of failing during fixture setup.

Would you like to submit a PR targeting the `docs-2.0-migration` branch with this enhancement?

Issue #1220, opened 2025-03-21
IMPR: 'import datajoint' slow on Mac amd64


drewyangdev:
### Problem
Not really a problem, but Mac import datajoint is way less performant comparing to import datajoint on Linux.

### Requirements
```
# make a directory
mkdir test_dj_import
cd ~/Downloads/test_dj_import

## make import.py
echo "import datajoint" > import.py

## CLI
conda create -n dj313 -c defaults python=3.13
pip install datajoint snakeviz

# generate cProfiler result
python -m cProfile -o result.log import.py
# filter and sort result to top 20 longest cumulative cpu time - check cProfile doc 
python -c 'import pstats; p=pstats.Stats("result.log");p.sort_stats("cumulative").print_stats(20)' > sorted_result.log

# if you want more
# visualize non filtered result.log, since can't visualize filtered result, the func recursive format was broken
pip install snakeviz
snakeviz result.log
# you will get an error saying too many depth, just change the filter to "depth: 3", "cutoff: 1/100"
```

### Justification
On either Mac/amd64 or Linux/amd64, the results are in the same order, but it takes ~6 sec to import on Mac, but ~1 sec on Linux.

On Mac/amd64
```
# sort_result.log
cat sorted_result.log | grep .py
        1    0.000    0.000    6.138    6.138 import.py:1(<module>)
        1    0.000    0.000    6.136    6.136 /Users/yam/Documents/DataJoint/datajoint-python/datajoint/__init__.py:1(<module>)
        1    0.000    0.000    3.010    3.010 /Users/yam/Documents/DataJoint/datajoint-python/datajoint/diagram.py:1(<module>)
        1    0.000    0.000    2.531    2.531 /Users/yam/Documents/DataJoint/datajoint-python/datajoint/admin.py:1(<module>)
        1    0.000    0.000    1.986    1.986 /Users/yam/Documents/DataJoint/datajoint-python/datajoint/connection.py:1(<module>)
        1    0.000    0.000    1.543    1.543 /Users/yam/Documents/DataJoint/datajoint-python/datajoint/table.py:1(<module>)
        1    0.000    0.000    1.416    1.416 /Users/yam/opt/miniconda3/envs/dj313/lib/python3.13/site-packages/pandas/__init__.py:1(<module>)
        1    0.000    0.000    1.068    1.068 /Users/yam/opt/miniconda3/envs/dj313/lib/python3.13/site-packages/pandas/core/api.py:1(<module>)
```

[...]

PR #1325, opened 2026-01-09
Perf: lazy-load heavy modules to speed up 'import datajoint' on macOS

sarpit2907:
This PR addresses #1220 by reducing eager imports during 'import datajoint'.

### Problem
On macOS, importing DataJoint was slow (~13s), primarily due to eager imports of:
- diagram
- table (which pulls in pandas)
- S3/minio stack

### Solution
- Introduce lazy-loading via `__getattr__` in `datajoint/__init__.py`
- Preserve public API (Diagram, Table, aliases, etc.)
- Heavy modules are imported only when accessed

### Results (macOS, Python 3.11)
- Before: ~13.1s
- After:  ~2.2s

Remaining import cost is dominated by core components (connection, numpy), which are expected.

### Notes
This change does not alter behavior and only affects import-time performance.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -8,6 +8,7 @@ from typing import Dict, List
 import certifi
 import networkx as nx
 import pytest
+
 minio = pytest.importorskip("minio")
 import urllib3
 from packaging import version
--- a/tests/test_cascading_delete.py
+++ b/tests/test_cascading_delete.py
@@ -36,6 +36,7 @@ def test_stepwise_delete(schema_simp_pop):
         not B()
     ), "failed to delete from the parent table following child table deletion"
 
+
 def test_part_delete_forwards_kwargs(schema_simp_pop):
     assert not dj.config["safemode"], "safemode must be off for testing"
     assert L() and A() and B() and B.C(), "schema population failed"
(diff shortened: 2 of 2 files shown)


dimitri-yatsenko:
Hi @sarpit2907, thank you for working on this optimization!

We already have PR #1321 open that addresses the same issue (#1220). The approaches differ slightly:

**Your approach (PR #1325):**
- Defers many imports in `__init__.py`
- Moves some imports to function-level within modules

**PR #1321 approach:**
- Uses `__getattr__` in `__init__.py` for true lazy loading
- Specifically targets the heaviest imports: `Diagram`, `kill`, and `cli`
- Caches imports in `globals()` to avoid repeated lookups
- Preserves the module's existing import structure

The `__getattr__` approach is more targeted—it only defers loading when specific symbols are accessed, while keeping everything else unchanged. This minimizes the risk of subtle import-order bugs.

Since #1321 is already in review and takes a more conservative approach, we'll likely proceed with that one. However, your contribution is appreciated and the timing analysis you provided is valuable!

Feel free to review #1321 and provide feedback if you see improvements we could incorporate.

PR #1404, opened 2026-02-18
Thread-safe mode with dj.Instance

dimitri-yatsenko:
## Summary

- Add `dj.Instance(host, user, password)` for isolated config + connection pairs
- Add thread-safe mode (`DJ_THREAD_SAFE=true`) that disables global singletons and forces explicit `dj.Instance()` usage
- Thread config through codec and hash_registry chains so all runtime config reads go through the connection, not global state
- Refactor `dj.Schema` from a function wrapper into a proper class subclassing `_Schema`, so `isinstance`, `hasattr`, etc. work naturally
- Add `adapter.split_full_table_name()` — inverse of `quote_identifier()` — for backend-portable table name parsing
- Move dependency graph queries (`load_primary_keys_sql`, `load_foreign_keys_sql`) into adapter methods
- Remove unsupported `schema.code()`/`save()` methods (untested, MySQL-specific, no callers)

## Usage

```python
# Isolated instance (new)
inst = dj.Instance(host="localhost", user="root", password="secret")
schema = inst.Schema("my_schema")

# Global pattern (unchanged)
schema = dj.Schema("my_schema")
```

```bash
# Force instance usage in concurrent apps
export DJ_THREAD_SAFE=true
```

## Use cases

- Web servers: one Instance per request/tenant
- Testing: isolated databases per test
- Multi-database: connect to production and staging simultaneously
- Parallel pipelines: independent connections per worker

## Design spec

See `docs/design/thread-safe-mode.md` for the full specification including global state audit.

## Test plan

- [x] Unit tests for `Instance`, `_ConfigProxy`, thread-safe guards (`tests/unit/test_thread_safe.py`)
- [x] Lint passes (ruff-format clean, only pre-existing mypy errors)
- [ ] Integration tests pass (semantic matching, jobs, GC, schema)

### DIFF
--- a/tests/integration/test_schema.py
+++ b/tests/integration/test_schema.py
@@ -228,14 +228,6 @@ def test_list_tables(schema_simp):
     assert actual == expected, f"Missing from list_tables(): {expected - actual}"
 
 
-def test_schema_save_any(schema_any):
-    assert "class Experiment(dj.Imported)" in schema_any.code
-
-
-def test_schema_save_empty(schema_empty):
-    assert "class Experiment(dj.Imported)" in schema_empty.code
-
-
 def test_uppercase_schema(db_creds_root):
     """
     https://github.com/datajoint/datajoint-python/issues/564


src/datajoint/schemas.py
```diff
@@ -838,7 +840,7 @@ def list_schemas(connection: Connection | None = None) -> list[str]:
     """
     return [
         r[0]
-        for r in (connection or conn()).query(
+        for r in (connection or _get_singleton_connection()).query(
```
mweitzel: This hard coded sql is mysql specific, it can be avoided if instead of _get_singleton_connection you called something like sql = connection.adapter.list_schemas_sql() ; return [r[0] for r in connection.query(sql)]

This method isn't protected by the "thread safe" os.env guard, but also doesn't need to be if it reads entirely from the instances which are passed instead of the global singleton here.

src/datajoint/schemas.py
mweitzel: Can this re-use `adapter.list_schemas_sql` as well? It looks like mysql/postgres agnostic sql, so it might be fine.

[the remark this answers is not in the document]
dimitri-yatsenko: Yes. The hardcoded `SHOW TABLES` is MySQL-specific. The adapter already has `list_tables_sql(schema_name)` that handles this per backend. Will fix.

dimitri-yatsenko: Good point. Even though `information_schema.schemata` is standard SQL, all query generation should go through the adapter for consistency. We'll add a `schema_exists_sql(name)` adapter method and use it here.

mweitzel:
Some string manipulation ins `schemas.py` remains mysql specific. Regardless of the adapter used, this is incompatible with postgres.

https://github.com/datajoint/datajoint-python/blob/34f744e1d50b186e08c01148db82686998bcf966/src/datajoint/schemas.py#L611

PR #1405, opened 2026-02-19
Fix populate antijoin to use .proj() for correct pending key computation

hummuscience:
## Summary

- Fix `_populate_direct()` to use `self.proj()` in the antijoin that computes pending keys
- Fix `jobs.refresh()` to use `self._target.proj()` when computing new keys for the jobs table
- Fix `progress()` fallback path to use `self.proj()` in the remaining count

## Problem

The antijoin that computes pending keys (`key_source - self`) does not project the target table to its primary key before the subtraction. When the target table has secondary (non-PK) attributes, the antijoin fails to match on primary key alone and **returns all keys instead of just the unpopulated ones**.

This causes:
- **`populate(reserve_jobs=False)`**: all `key_source` entries are iterated instead of just pending ones. Mitigated by `if key in self:` check inside `_populate1()`, but wastes time on large tables.
- **`populate(reserve_jobs=True)`**: `jobs.refresh()` inserts **all** keys into the jobs table as `'pending'`, not just truly pending ones. Workers then waste their `max_calls` budget processing already-completed entries before reaching any real work — effectively making distributed populate non-functional for partially-populated tables.
- **`progress()`**: reports incorrect remaining counts in the fallback (no common attributes) path.

### Reproduction

```python
# Given a Computed/Imported table with secondary attributes:
MyTable.populate(max_calls=5)  # partially populate

# This returns ALL keys, not just unpopulated ones:
pending = MyTable.key_source - MyTable()
print(len(pending))  # == len(key_source), should be len(key_source) - 5

# But this works correctly:
pending = MyTable.key_source - MyTable().proj()
print(len(pending))  # == len(key_source) - 5  ✓
```

## Fix

Add `.proj()` to the target side of all three antijoins so the subtraction matches on primary key only:

| Location | Before | After |
|----------|--------|-------|
| `autopopulate.py:406` | `self._jobs_to_do(restrictions) - self` | `self._jobs_to_do(restrictions) - self.proj()` |
| `autopopulate.py:704` | `todo - self` | `todo - self.proj()` |
| `jobs.py:373` | `key_source - self._target` | `key_source - self._target.proj()` |

## Test plan

- [x] Added `test_populate_antijoin_with_secondary_attrs` — verifies pending key count after partial populate (direct mode)
- [x] Added `test_populate_distributed_antijoin` — verifies `jobs.refresh()` only creates entries for truly pending keys (distributed mode)
- [ ] Existing test suite passes

### DIFF
--- a/tests/integration/test_autopopulate.py
+++ b/tests/integration/test_autopopulate.py
@@ -143,30 +143,6 @@ def test_populate_antijoin_with_secondary_attrs(clean_autopopulate, subject, exp
     assert len(pending_after) == 0, f"Antijoin returned {len(pending_after)} pending keys after full populate, expected 0."
 
 
-def test_populate_distributed_antijoin(clean_autopopulate, subject, experiment):
-    """Test that reserve_jobs=True correctly identifies pending keys.
-
-    When using distributed mode, jobs.refresh() must only insert truly pending
-    keys into the jobs table, not already-completed ones.
-    """
-    assert subject, "root tables are empty"
-    assert not experiment, "table already filled?"
-
-    total_keys = len(experiment.key_source)
-
-    # Partially populate
-    experiment.populate(max_calls=2)
-    assert len(experiment) == 2 * experiment.fake_experiments_per_subject
-
-    # Refresh jobs — should only create entries for unpopulated keys
-    experiment.jobs.refresh(delay=-1)
-    pending_jobs = len(experiment.jobs.pending)
-    assert pending_jobs == total_keys - 2, f"jobs.refresh() created {pending_jobs} pending jobs, expected {total_keys - 2}."
-
-    # Clean up
-    experiment.jobs.delete_quick()
-
-
 def test_populate_antijoin_overlapping_attrs(prefix, connection_test):
     """Regression test: antijoin with overlapping secondary attribute names.
 
@@ -193,7 +169,7 @@ def test_populate_antijoin_overlapping_attrs(prefix, connection_test):
         sensor_id : int32
         ---
         num_samples : int32
-        quality : float
+        quality : decimal(4,2)
         """
         contents = [
             (1, 100, 0.95),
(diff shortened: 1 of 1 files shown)


hummuscience:
Thanks for the quick response. Excuse the oversight, it was the end of a long day :) I updated everything to reflect the bug.                                                                                                                          

The exact conditions that trigger the bug:

The antijoin key_source - self matches on ALL common column names. When key_source returns secondary attributes that share names with the target's secondary attributes, SQL matches on those too. If values
differ, no match is found and all keys appear "pending."

Here are the table definitions from our pipeline:
```

 class LightningPoseOutput(dj.Computed):
     definition = """
     -> LightningPoseTask
     ---
     lp_output_path : varchar(512)
     processing_datetime : datetime
     num_frames = null : int unsigned
     quality = 'good' : enum('good', 'review', 'exclude')
     ...
     """

 class InitialContainer(dj.Computed):
     definition = """
     -> LightningPoseOutput
     ---
     container_path : varchar(512)
     processing_datetime : datetime          # same name, different value
     num_frames : int unsigned               # same name, same value
     quality = 'good' : enum('good', 'review', 'exclude')  # same name, same value
     ...
     """

     @property
     def key_source(self):
         return LightningPoseOutput() & production_models
```

@property
def key_source(self):
return LightningPoseOutput() & production_models

When populate() computes pending keys:
- key_source = LightningPoseOutput() → includes processing_datetime, num_frames, quality
- self = InitialContainer() → also has processing_datetime, num_frames, quality
- key_source - self matches on ALL common columns including those 3 secondary attrs
- Since LP's processing_datetime ≠ Container's processing_datetime, no match → all keys appear "pending"

Why we designed it this way:

In this case, InitialContainer simply converts the Lightning Pose CSV output into a https://github.com/neuroinformatics-unit/movement xarray container (NetCDF). The underlying data is the same — same video,
same frames, same quality assessment — so the metadata columns (num_frames, quality, processing_datetime) naturally share the same names. The only value that actually differs is processing_datetime, which
records when each step ran. Having the same column names across parent and child tables is a natural result of a pipeline where each stage wraps or transforms the previous stage's output, and populate() should
determine pending keys based on primary key identity regardless.

That said — if there's a recommended DataJoint pattern for this kind of multi-stage pipeline, let me know :)

PR #1437, opened 2026-04-22
fix: remove EXPRESSION column dependency in get_indexes_sql()

### LINKED ISSUE #1436: get_indexes_sql() fails on MySQL < 8.0.13 and MariaDB (missing EXPRESSION column)
`get_indexes_sql()` in `src/datajoint/adapters/mysql.py` queries `COALESCE(COLUMN_NAME, CONCAT('(', EXPRESSION, ')'))` from `information_schema.STATISTICS`, but the `EXPRESSION` column only exists in MySQL 8.0.13+. On MySQL 5.7 or MariaDB, this fails with `Unknown column 'EXPRESSION' in 'field list'` the first time any table class is decorated with `@schema`.

Introduced in commit bd35a7e (PR #1338). Prior to that, the query just selected `COLUMN_NAME` directly.

A fallback for when `EXPRESSION` isn't available would fix this for both MySQL < 8.0.13 and MariaDB.

dimitri-yatsenko:
## Summary

- `get_indexes_sql()` referenced the `EXPRESSION` column in `information_schema.STATISTICS`, which only exists in MySQL 8.0.13+
- This broke `@schema` decoration on MySQL < 8.0.13 and MariaDB with `Unknown column 'EXPRESSION'`
- Use `COLUMN_NAME` directly — functional indexes (NULL column name) are already filtered out in `heading.py`

Fixes #1436

## Test plan

- [ ] Verify `@schema` decoration works on MySQL 5.7 or MariaDB
- [ ] Verify secondary indexes are still correctly read on MySQL 8.0+
- [ ] Verify tables with functional indexes don't cause errors (functional index entries are skipped)

### DIFF
--- a/src/datajoint/adapters/mysql.py
+++ b/src/datajoint/adapters/mysql.py
@@ -737,12 +737,27 @@ class MySQLAdapter(DatabaseAdapter):
     def get_indexes_sql(self, schema_name: str, table_name: str) -> str:
         """Query to get index definitions.
 
-        Note: For MySQL 8.0+, EXPRESSION column contains the expression for
+        Note: For MySQL 8.0.13+, EXPRESSION column contains the expression for
         functional indexes. COLUMN_NAME is NULL for such indexes.
+        On MySQL < 8.0.13 and MariaDB, the EXPRESSION column does not exist;
+        heading.py falls back to get_indexes_sql_fallback() in that case.
         """
         return (
             f"SELECT INDEX_NAME as index_name, "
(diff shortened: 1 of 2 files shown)


dimitri-yatsenko:
Note: DataJoint has not yet explicitly supported functional indexes — that would be a future feature, maybe. So dropping the EXPRESSION reference doesn't lose any current functionality.

esutlie (approved):

PR #1455, opened 2026-05-20
docs: fix code formatting on rendered codecs API page

dimitri-yatsenko:
## Summary

[`https://docs.datajoint.com/api/datajoint/codecs/`](https://docs.datajoint.com/api/datajoint/codecs/) renders docstring code as plain prose in three places. The page is built by `mkdocstrings` with `docstring_style: numpy` and Markdown-flavored bodies, so reST and Google-style syntax don't get parsed.

| Where | Was | Now |
|---|---|---|
| `Codec` class docstring (two examples) | reST `::` literal blocks | Markdown fenced code blocks |
| `decode_attribute` | Google-style `Args:` / `Returns:` (parsed as paragraphs) | NumPy `Parameters` / `Returns` with dashes (renders as a parameter table) |
| Module-level `Example:` | 4-space-indented Python (worked by accident) | Markdown fenced code block |

No behavior change — docstrings only.

## Test plan
- [ ] After merge + docs rebuild, confirm `codecs` API page:
  - the two examples under `Codec` ("Use in table definitions" / "Skip auto-registration") render as Python code blocks (no literal `::`)
  - `decode_attribute` shows `attr`, `data`, `squeeze`, `connection` in a parameter table consistent with sibling functions
  - the module-level `Example:` Python sample renders as a fenced code block
- [x] `python -c "import datajoint.codecs"` still imports cleanly
- [x] `mypy src/datajoint/codecs.py` clean (pre-existing unrelated failure in `storage_adapter.py:98` was skipped on commit; CI lint doesn't gate on mypy)

## Note on the unrelated mypy failure
Local pre-commit's mypy hook fails on `storage_adapter.py:98` (a `Deprecated.get()` typing shim for older `importlib.metadata`). That error exists on master and is not enforced by CI (`.github/workflows/lint.yaml` only runs codespell + ruff + ruff-format). Skipped with `SKIP=mypy git commit` per the documented pattern at the top of `.pre-commit-config.yaml`. Happy to file a separate PR for the typing shim if you'd like.

### DIFF
--- a/src/datajoint/codecs.py
+++ b/src/datajoint/codecs.py
@@ -8,29 +8,32 @@ semantics for complex Python objects.
 Codecs auto-register when subclassed - no decorator needed (Python 3.10+).
 
 Example:
-    class GraphCodec(dj.Codec):
-        name = "graph"
 
-        def get_dtype(self, is_store: bool) -> str:
-            return "<blob>"
+```python
+class GraphCodec(dj.Codec):
+    name = "graph"
 
-        def encode(self, graph, *, key=None, store_name=None):
-            return {'nodes': list(graph.nodes()), 'edges': list(graph.edges())}
+    def get_dtype(self, is_store: bool) -> str:
+        return "<blob>"
 
-        def decode(self, stored, *, key=None):
-            import networkx as nx
-            G = nx.Graph()
-            G.add_nodes_from(stored['nodes'])
-            G.add_edges_from(stored['edges'])
-            return G
+    def encode(self, graph, *, key=None, store_name=None):
+        return {'nodes': list(graph.nodes()), 'edges': list(graph.edges())}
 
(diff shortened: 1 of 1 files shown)


MilagrosMarin (approved):
Thanks @dimitri-yatsenko! Verified the fix:

✅ `datajoint-docs/mkdocs.yaml` confirms `docstring_style: numpy` — reST `::` and Google `Args:`/`Returns:` would indeed render as literal text on `/api/datajoint/codecs/`.
✅ All three conversions land in formats mkdocstrings + numpy parser consume. After this PR, `grep "Args:\|Returns:\|Raises:"` in `codecs.py` returns 0 matches.
✅ `decode_attribute` parameter list matches the function signature.

Approving. The mypy-skip note is handled transparently and CI doesn't gate on it anyway.

PR #1465, opened 2026-06-07
fix(staged_insert): converge metadata shape with ObjectCodec.encode

dimitri-yatsenko:
## Summary

Bring \`staged_insert1\` into compliance with the [Staged Insert Specification](https://github.com/datajoint/datajoint-docs/pull/177) (`datajoint-docs#177`). The spec defines a normative metadata-dict shape; today's implementation diverges from it in two ways that are observable in the database.

**Slated for the DataJoint 2.3 release.**

### What changes

**1. Metadata shape now matches `ObjectCodec.encode` exactly.**

The same content stored via the two insert paths used to yield different column dicts:

| Path | Dict shape |
|---|---|
| `ObjectCodec.encode` (`builtin_codecs/object.py:166-174`) | `{path, store, size, ext, is_dir, item_count, timestamp}` |
| Staged, directory (old) | `{path, size, hash, ext, is_dir, item_count, timestamp}` |
| Staged, single file (old) | `{path, size, hash, ext, is_dir, timestamp, mime_type?}` |

After this PR, `_compute_metadata` returns the encode shape in both cases:
- drops `hash: None` (never carried information)
- drops `mime_type` (not in encode shape)
- file case adds `item_count: None`
- both cases add `store: <resolved store name>`

**2. Per-field store resolution.**

Today's `staged_insert` resolves the backend from `stores.default` regardless of the field's type spec. A field declared `<object@local>` would write through `stores.default`, and the `store` key recorded in the row's metadata column would name the wrong store (or be missing entirely). Now we resolve `store_name` from `attr.store` via `resolve_dtype()` and use that for both path/backend resolution and the metadata's `store` field — matching what ordinary `insert1` does through `table.py:1342`.

**3. Drops the `.manifest.json` sidecar.**

The staged path was writing a per-object manifest file that the encode path doesn't produce. The metadata dict already records total `size` and `item_count`; per-file listings are recoverable by walking the canonical directory. Removing the sidecar eliminates a small storage-layout divergence between the two paths.

**4. Docstring fix.**

Module docstring and the `staged_insert1` function docstring both showed `staged.rec['raw_data'] = z`. Per the spec, the caller does **not** assign anything to `staged.rec[<staged field>]` — the framework computes the metadata dict. The implementation already overwrote the user's assignment in `_finalize`, so this was harmless but misleading. Examples updated.

### Test added

`tests/integration/test_object.py::TestStagedInsert::test_staged_insert_metadata_shape_matches_encode` — exercises both the single-file and directory paths, captures the metadata dict assigned to the row, and asserts:
- Keys match `ObjectCodec.encode`'s output exactly for equivalent content
- Per-key values agree (`store`, `ext`, `is_dir`, `item_count`)

## Test plan

- [x] All 6 staged-insert tests pass (including the new one)
- [x] All 44 tests in `tests/integration/test_object.py` pass
- [x] CI green
- [ ] Re-review by codeowners against the spec

## Related

- Spec: datajoint/datajoint-docs#177 (DRAFT)
- This PR is what `datajoint-docs#177`'s \"Implementation status\" note refers to once merged. After both merge, the spec ceases to be forward-looking for the `<object@>` path.

### DIFF
--- a/src/datajoint/staged_insert.py
+++ b/src/datajoint/staged_insert.py
@@ -5,16 +5,19 @@ This module provides the StagedInsert class which allows writing directly
 to object storage before finalizing the database insert.
 """
 
-import json
-import mimetypes
 from contextlib import contextmanager
 from datetime import datetime, timezone
-from typing import IO, Any
+from typing import IO, TYPE_CHECKING, Any
 
 import fsspec
 
+from .codecs import resolve_dtype
 from .errors import DataJointError
-from .storage import StorageBackend, build_object_path
+from .hash_registry import get_store_backend
+from .storage import build_object_path
+
+if TYPE_CHECKING:
+    from .storage import StorageBackend
 
 
 class StagedInsert:
@@ -30,15 +33,14 @@ class StagedInsert:
             staged.rec['subject_id'] = 123
             staged.rec['session_id'] = 45
 
-            # Create object storage directly
+            # Write directly to object storage
             z = zarr.open(staged.store('raw_data', '.zarr'), mode='w', shape=(1000, 1000))
             z[:] = data
 
-            # Assign to record
-            staged.rec['raw_data'] = z
-
-        # On successful exit: metadata computed, record inserted
-        # On exception: storage cleaned up, no record inserted
+        # On clean exit: metadata is computed and the row is inserted.
+        #   The caller does NOT assign anything to staged.rec[<object field>] —
+        #   the framework computes the metadata dict.
+        # On exception: storage cleaned up, no row inserted.
     """
 
     def __init__(self, table):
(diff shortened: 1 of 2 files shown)


MilagrosMarin (approved):
Verified end-to-end against the spec contract in #177:

✅ Metadata shape `{path, store, size, ext, is_dir, item_count, timestamp}` matches `ObjectCodec.encode` (`builtin_codecs/object.py:166-174`) exactly — drops `hash: None`, drops `mime_type`, adds `store` and `item_count` consistently across file and directory cases.
✅ Per-field store resolution via `resolve_dtype(..., store_name=attr.store)` — `<object@local>` now correctly writes through `stores.local` and records `"store": "local"` in the metadata.
✅ `ObjectRef.from_json` is forward-compatible (`data.get("hash")`, `data.get("item_count")` defaults) — existing rows with `hash: None` still parse cleanly.
✅ The new `test_staged_insert_metadata_shape_matches_encode` is exactly the structural-equality assertion the spec calls for — keys match, and `store`/`ext`/`is_dir`/`item_count` agree across both paths.

Two minor notes, neither blocking:
- Test covers `<object@local>` via `ObjectFile`/`ObjectFolder`. The default-store path (`<object@>`, `store_name=None`) uses the same code branch but isn't directly exercised — worth a one-line note if you want comprehensive coverage.
- Existing `.manifest.json` sidecars from earlier writes become orphans. Garbage collector should reclaim them; not a regression.

Approving — this closes the convergence concern from #177.

PR #1466, opened 2026-06-10
feat(deploy): set_replica_identity for PostgreSQL CDC (#1447)

### LINKED ISSUE #1447: Add config option for REPLICA IDENTITY FULL on PostgreSQL tables
## Context

Databricks Lakehouse Sync and other logical-replication-based CDC tools require PostgreSQL tables to have \`REPLICA IDENTITY FULL\` set. Without it, Lakehouse Sync silently skips the table — no error, just missing data downstream:

> \"Tables without REPLICA IDENTITY FULL will be skipped. Run ALTER TABLE ... REPLICA IDENTITY FULL to include them.\"

This is mandatory for Lakehouse Sync — there is no alternative. This issue requests a generic, opt-in mechanism in \`datajoint-python\` so Postgres-on-Lakebase deployments and other CDC-driven setups can configure it without per-table manual ALTER statements.

## What REPLICA IDENTITY FULL does

PostgreSQL's WAL records every change. \`REPLICA IDENTITY\` controls how much of the **old row** is logged for UPDATE and DELETE:

| Setting | What's logged (old row) | WAL cost |
|---------|------------------------|----------|
| **DEFAULT** | Primary key columns only | Minimal |
| **FULL** | Entire old row, all columns | Higher for UPDATE/DELETE on TOASTed columns |

Key facts:
- INSERTs are unaffected — full row always logged regardless of setting
- UPDATEs log the full old row including TOAST (bytea) columns even if unchanged
- The ALTER is **metadata-only, instant, idempotent**, can be applied at any time
- This is **PostgreSQL-only** — MySQL has no equivalent (MySQL uses binlog-based CDC)

For DataJoint's overwhelmingly insert-append workload, the real-world cost of FULL is negligible. The one cost scenario is bulk \`delete()\` on tables with \`<blob>\` columns (e.g., 100 rows × 10 MB blobs ≈ 1 GB WAL burst). Transient and infrequent.

## Proposed design

Add a config option \`database.replica_identity\` with allowed values \`\"default\"\` (default) and \`\"full\"\`. Environment variable: \`DJ_REPLICA_IDENTITY\`.

Two mechanisms:

### 1. New tables — applied automatically via \`post_ddl\` during \`declare()\`

```sql
ALTER TABLE \"schema_name\".\"table_name\" REPLICA IDENTITY FULL;
```

### 2. Existing tables — utility method

```python
schema.set_replica_identity(\"full\")  # iterates schema's tables, runs ALTER for each
```

Reason a utility is needed: when \`@schema\` decorates a class for an already-existing table, \`declare()\` is **not** called. \`post_ddl\` only fires on table creation. Users who add the config after their pipeline exists need a separate mechanism.

### MySQL behavior

If the config is set on a MySQL connection, emit a \`UserWarning\` consistent with the existing \`database.name\` warning pattern (PR #1426). REPLICA IDENTITY is a PostgreSQL concept; setting it on MySQL is a no-op.

## Compliance and security considerations

Under DEFAULT, only PK values appear in WAL. Under FULL, entire rows — including any PHI, PII, or sensitive data — appear in:
- WAL segments on disk (transient, recycled after checkpoint)
- Logical replication streams
- Anything subscribed to those streams

For self-hosted PostgreSQL, this is a real consideration. For managed PostgreSQL with controlled WAL access (Lakebase, RDS with logical replication only to a specific subscriber), the WAL stays inside the managed environment's security boundary.

This is why the option must be **opt-in**, not on-by-default.

## Why a generic upstream feature, not Lakebase-specific

CDC consumers besides Lakehouse Sync also benefit (Debezium and ClickHouse ClickPipes can use FULL when PKs aren't sufficient; Azure CDC; logical replication to other Postgres). Keeping the config generic lets any Postgres-based deployment configure it without backend-specific code paths.

The Lakebase-specific decision (\"Lakebase deployments default to FULL\") lives in the deployment configuration of the consumer, not in this upstream code. Reference: [DataJoint × Databricks DECISIONS.md](https://github.com/datajoint-company/datajoint-databricks/blob/main/DECISIONS.md) — internal repo, summary available on request.

## Industry context

| CDC Tool | FULL Required? | Recommendation |
|----------|---------------|----------------|
| Debezium | No (for tables with PKs) | DEFAULT preferred |
| Azure CDC | No | Recommends against FULL |
| ClickHouse ClickPipes | No (if PK exists) | DEFAULT preferred |
| **Databricks Lakehouse Sync** | **Yes, mandatory** | No alternative |

Most CDC tools prefer DEFAULT for performance. Databricks mandates FULL for SCD Type 2 history. The opt-in design serves all of them.

## References

- [PostgreSQL: Logical Replication — Replica Identity](https://www.postgresql.org/docs/current/logical-replication-publication.html)
- [Databricks: Lakehouse Sync](https://docs.databricks.com/aws/en/oltp/projects/lakehouse-sync)
- [Xata: Performance impact of REPLICA IDENTITY FULL](https://xata.io/blog/replica-identity-full-performance)

dimitri-yatsenko:
## Summary

Adds `dj.deploy.set_replica_identity(target, mode, dry_run)` — an idempotent deployment-time operation that applies `ALTER TABLE ... REPLICA IDENTITY DEFAULT|FULL` across a Schema or single Table on PostgreSQL.

Closes #1447. Slated for DataJoint 2.3.

## Why `dj.deploy` and not `dj.migrate` or auto-emit in `declare()`

The original issue proposed two mechanisms — a config flag applied automatically during `declare()`, plus a utility for existing tables. After implementing both, we reverted to a single migration-only path. Reasons:

1. **Not a migration.** Nothing legacy is being fixed. The setting is a property of the deployment environment, and a fresh declare in a new environment may need it re-applied. `dj.migrate` is for one-shot schema/state evolution; this is operational configuration.
2. **Not a declare-time concern.** Mixing auto-emit-on-declare with a separate utility produces mixed state (new tables `FULL`, old tables `DEFAULT`) until both run. One mechanism is coherent.
3. **A new home for operational helpers.** `set_replica_identity` is the first of a category: publication membership, vacuum/reindex, role grants, table-level autovacuum params. These belong together, not in `migrate.py`. The new `dj.deploy` module is the right home.

## What's in the PR

| File | Change |
|---|---|
| `src/datajoint/adapters/postgres.py` | New `replica_identity_ddl(full_name, mode)` — pure DDL emitter; both `default` and `full` produce explicit ALTERs; invalid mode raises. |
| `src/datajoint/deploy.py` (new) | `set_replica_identity(target, mode, dry_run)` — dispatches Schema/Table, validates backend, supports dry_run, returns `{tables_analyzed, tables_modified, ddl}`. |
| `src/datajoint/__init__.py` | Exports `dj.deploy`. |
| `tests/unit/test_adapters.py` | Three new tests for the PG adapter DDL (full / default / invalid mode). |
| `tests/unit/test_deploy.py` (new) | Seven unit tests covering mode/target/backend validation, dry_run, apply, default mode, empty schema. Uses stub adapters — no live PG container required. |

## User-facing API

```python
from datajoint import deploy

# Preview what would change
deploy.set_replica_identity(my_schema, mode="full", dry_run=True)
# {'tables_analyzed': 12, 'tables_modified': 0, 'ddl': ['ALTER TABLE "ms"."t1" ...', ...]}

# Apply
deploy.set_replica_identity(my_schema, mode="full", dry_run=False)

# Single table also supported
deploy.set_replica_identity(MyTable, mode="full", dry_run=False)
```

Non-PostgreSQL backends raise `DataJointError` (loud failure rather than silent no-op — consistent with this being an opt-in operational tool).

## Test plan

- [x] All 10 new unit tests pass (`tests/unit/test_deploy.py` + 3 new tests in `tests/unit/test_adapters.py`)
- [x] `tests/unit/test_settings.py` (80 tests) and `tests/integration/test_declare.py` (44 tests) green — no regressions from the revert
- [ ] CI green on this PR
- [ ] Integration test against a real PG container (deferred — exercised manually before tag)
- [ ] Companion docs PR in `datajoint-docs` (in flight)

### DIFF
--- a/src/datajoint/__init__.py
+++ b/src/datajoint/__init__.py
@@ -57,6 +57,7 @@ __all__ = [
     # Other
     "errors",
     "migrate",
+    "deploy",
     "DataJointError",
     "ThreadSafetyError",
     "logger",
@@ -69,6 +70,7 @@ __all__ = [
 # =============================================================================
 from . import errors
 from . import migrate
+from . import deploy
 from .codecs import (
     Codec,
     get_codec,
--- a/src/datajoint/adapters/postgres.py
+++ b/src/datajoint/adapters/postgres.py
@@ -1266,6 +1266,31 @@ class PostgreSQLAdapter(DatabaseAdapter):
         quoted_values = ", ".join(f"'{v}'" for v in values)
         return f"CREATE TYPE {self.quote_identifier(type_name)} AS ENUM ({quoted_values})"
 
+    def replica_identity_ddl(self, full_table_name: str, mode: str) -> str:
+        """
+        Generate ALTER TABLE ... REPLICA IDENTITY statement.
+
+        Controls how much of the old row PostgreSQL writes to WAL on UPDATE/DELETE.
+        ``"default"`` logs only primary-key columns; ``"full"`` logs the entire row.
+        Required by some CDC tools (e.g. Databricks Lakehouse Sync) that need the
+        full pre-image to drive Slowly-Changing-Dimension history.
+
+        The ALTER is metadata-only, instant, and idempotent — re-applying the same
+        mode is a no-op at the storage layer.
+
+        Examples
+        --------
+        >>> adapter.replica_identity_ddl('"schema"."table"', 'full')
+        'ALTER TABLE "schema"."table" REPLICA IDENTITY FULL'
+        >>> adapter.replica_identity_ddl('"schema"."table"', 'default')
+        'ALTER TABLE "schema"."table" REPLICA IDENTITY DEFAULT'
+        """
+        if mode not in ("default", "full"):
+            from ..errors import DataJointError
+
+            raise DataJointError(f"Unsupported replica_identity mode: {mode!r}. Expected 'default' or 'full'.")
+        return f"ALTER TABLE {full_table_name} REPLICA IDENTITY {mode.upper()}"
+
     def get_pending_enum_ddl(self, schema_name: str) -> list[str]:
         """
         Get DDL statements for pending enum types and clear the pending list.
--- a/src/datajoint/deploy.py
+++ b/src/datajoint/deploy.py
@@ -1,0 +1,157 @@
+"""
+Deployment-time operations for configuring an existing DataJoint pipeline.
+
+This module hosts idempotent operational helpers — things you run as part of a
+deploy hook to configure a schema for its environment, distinct from
+:mod:`datajoint.migrate` which handles one-shot schema/state evolution.
+
+The boundary between the two:
+
+- :mod:`datajoint.migrate` — fix legacy state, evolve a schema definition,
+  retroactive corrections. Cadence: one-shot. Examples: ``migrate_columns``,
+  ``add_job_metadata_columns``, ``rebuild_lineage``.
+- :mod:`datajoint.deploy` — configure an environment for a consumer's
+  requirements (CDC tools, replication, role grants, performance tuning).
+  Cadence: re-runnable, idempotent. Examples: :func:`set_replica_identity`.
+
+Functions in this module should be safe to call repeatedly from a deploy hook
+without accumulating side effects.
+"""
+
+from __future__ import annotations
+
+from typing import Any
+
+from .errors import DataJointError
+
+
+def set_replica_identity(target: Any, mode: str = "full", dry_run: bool = True) -> dict:
+    """
+    Apply ``ALTER TABLE ... REPLICA IDENTITY <mode>`` to a schema or table on PostgreSQL.
+
+    ``REPLICA IDENTITY`` controls how much of the **old row** PostgreSQL writes to
+    the write-ahead log on UPDATE/DELETE. Under ``DEFAULT``, only primary-key
+    columns appear in WAL; under ``FULL``, the entire old row does.
+
+    Why this exists
+    ---------------
+    Some change-data-capture (CDC) consumers require the full row pre-image to
+    drive their downstream models. The canonical example is **Databricks
+    Lakehouse Sync**: tables without ``REPLICA IDENTITY FULL`` are silently
+    skipped by the sync — no error, just missing data downstream. Other CDC
+    tools (Debezium, ClickHouse ClickPipes, Azure CDC) work fine with
+    ``DEFAULT`` when tables have a primary key; only Databricks mandates
+    ``FULL``.
+
+    This helper is the **operational** way to apply the setting. It is not a
+    migration: there's no legacy state being fixed; the setting is simply a
+    property of the deployment environment, and a fresh declare in a new
+    environment may need it re-applied. It is idempotent — re-applying the
+    same mode is a no-op at the storage layer — so it is safe to call from a
+    deploy hook on every release.
+
+    Cost
+    ----
+    The ALTER itself is metadata-only and instant. The cost is in WAL volume
+    after the change: UPDATE/DELETE on tables with FULL log the entire old row,
+    which can be sizable on tables with TOASTed bytea columns. For DataJoint's
+    typical insert-append workload, this cost is negligible. The notable
+    scenario is bulk ``delete()`` on tables with ``<blob>`` columns — a
+    transient WAL burst proportional to the deleted-row payload size.
+
+    Compliance considerations
+    -------------------------
+    Under ``DEFAULT``, only primary-key values appear in WAL. Under ``FULL``,
+    entire rows do — including any PHI/PII/sensitive columns. For self-hosted
+    PostgreSQL with unrestricted WAL access this is a real consideration; for
+    managed PostgreSQL with logical replication confined to a specific
+    subscriber (Lakebase, RDS), WAL stays inside the managed environment's
+    security boundary. Apply intentionally.
+
+    Parameters
+    ----------
+    target : Schema or Table
+        A :class:`datajoint.Schema` (all user tables) or a
+        :class:`datajoint.Table` class/instance (just that table).
+    mode : str, default ``"full"``
+        ``"default"`` (PK only, minimal WAL) or ``"full"`` (entire row).
+    dry_run : bool, default ``True``
+        If True, collect the DDL statements but do not execute. Set to False
+        to actually apply.
+
+    Returns
+    -------
+    dict
+        - ``tables_analyzed`` (int): number of tables considered.
+        - ``tables_modified`` (int): number of tables on which the ALTER ran.
+          Always 0 when ``dry_run=True``.
+        - ``ddl`` (list[str]): the DDL statements that were (or would be) executed.
+
+    Raises
+    ------
+    DataJointError
+        If the target's backend is not PostgreSQL, or if ``mode`` is not one of
+        ``"default"`` / ``"full"``.
+
+    Examples
+    --------
+    >>> from datajoint.deploy import set_replica_identity
+    >>> # Preview
+    >>> set_replica_identity(my_schema, mode="full", dry_run=True)
+    {'tables_analyzed': 12, 'tables_modified': 0, 'ddl': ['ALTER TABLE "ms"."t1" REPLICA IDENTITY FULL', ...]}
+    >>> # Apply
+    >>> set_replica_identity(my_schema, mode="full", dry_run=False)
+    {'tables_analyzed': 12, 'tables_modified': 12, 'ddl': [...]}
+    >>> # Single table
+    >>> set_replica_identity(MyTable, mode="full", dry_run=False)
+
+    See Also
+    --------
+    PostgreSQL: `Logical Replication — Replica Identity
+    <https://www.postgresql.org/docs/current/logical-replication-publication.html>`_.
+    Databricks: `Lakehouse Sync
+    <https://docs.databricks.com/aws/en/oltp/projects/lakehouse-sync>`_.
+    """
+    if mode not in ("default", "full"):
+        raise DataJointError(f"mode must be 'default' or 'full'; got {mode!r}")
+
+    from .schemas import _Schema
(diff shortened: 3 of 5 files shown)


MilagrosMarin (changes requested):
Read this carefully against `dj.deploy.set_replica_identity` and the spec in datajoint-docs#180. Mechanics are clean, the dispatch ladder handles all three target shapes (Schema, Table class, Table instance), and the duck-typed `hasattr(adapter, "replica_identity_ddl")` PG check is a friendly contract for third-party PG-compatible adapters. A few observations, mostly suggestions — none are large fixes, but two of them feel worth catching before this lands.

## Suggestions worth addressing

**1. `assert` statements in the dispatch ladder might bite under `-O`.** Four places in `deploy.py` (lines 123, 125, 130, 135):

```python
assert connection is not None, "Schema has no active connection"
assert target.database is not None, "Schema is not activated"
assert connection is not None, "Table has no active connection"  # ×2
```

The project's CLAUDE.md flags this as a pattern to avoid — Python's `-O` flag strips asserts at bytecode-compile time. In a deploy-hook context that's not unusual (some Docker base images run `-O`; some CI orchestrators do). If `-O` strips these, the user gets `AttributeError: 'NoneType' object has no attribute 'adapter'` on the next line instead of the intended `DataJointError`. Lightly suggest converting to explicit `if … raise DataJointError(…)` — the messages are already there, just need the runtime guarantee. The Schema `database is None` case is the most concerning since it would continue with `None` as a schema name and produce a malformed `ALTER TABLE None ...`.

**2. The two `Table`-target branches have no test coverage.** All seven tests in `tests/unit/test_deploy.py` build a `_FakeSchema`; the Table-class branch (`deploy.py:127-132`) and Table-instance branch (`deploy.py:133-137`) are never exercised. They use `target.full_table_name` (a different code path from the Schema branch's `adapter.make_full_table_name(...)`), so a regression in `Table.full_table_name` semantics — for example, the `database is None` case returning `None` — wouldn't surface here. A `_FakeTable` test for each branch with the existing `_FakeAdapter` would close the gap cheaply.

## Lighter polish — feel free to defer

**3. Test count in the PR description says "16 new unit tests"** — I count 7 in `test_deploy.py` + 3 new ones in `test_adapters.py` = 10. Trivial fix; mostly mentioning it so reviewers reading the description aren't searching for tests that aren't there.

**4. Type hints are looser than the spec.** Spec in datajoint-docs#180 documents the signature as `target: Schema | Table, mode: Literal["default", "full"]`. Impl uses `target: Any, mode: str` — runtime validation does the same work, but the tighter types would help IDEs and type-checkers. `Literal["default", "full"]` is straightforward; `Schema | Table` would need a forward reference for the circular import.

**5. `mode="FULL"` (uppercase) is rejected on a strict equality.** The PG adapter does `mode.upper()` internally so it's case-tolerant; the deploy function isn't. Could be `mode.lower() not in (...)` for symmetry — minor UX nit, not worth a separate PR.

**6. Worth one line on lock behavior in the docstring's Cost section.** `ALTER TABLE ... REPLICA IDENTITY` takes an `AccessExclusiveLock` for the brief metadata flip. On an actively-ingesting table, that can stack behind in-flight transactions and block briefly — a real consideration for deploy-hook timing on a busy production table. Something like *"requires a brief exclusive lock — will block behind in-flight writes"* would round out the existing Cost framing.

**7. Partial-failure semantics on multi-table runs aren't explicit.** If `connection.query(ddl)` raises on table N of M, the first N-1 tables are already modified, but the caller doesn't see the dict — the exception propagates. Re-running brings the rest into compliance (idempotent), but a one-sentence note in the docstring would make that explicit. Not load-bearing.

**8. Minor style nit:** line 143 has adjacent-string concatenation (`f"... adapter " "does not ..."`) — either one f-string or one continuous string would read more naturally. Cosmetic only.

## Verified ✅ (no action needed)

- DDL format matches `replica_identity_ddl()` adapter method exactly (`ALTER TABLE {full_name} REPLICA IDENTITY {MODE}`)
- `Schema.list_tables()` already excludes `~`-prefixed hidden tables and `~~` job tables (the CDC-relevant filter)
- `dj.Schema` inherits from `_Schema` — `isinstance(target, _Schema)` correctly catches the public API
- Schema dispatch (`adapter.make_full_table_name(...)`) and Table dispatch (`Table.full_table_name` property, which internally calls the same adapter method) produce equivalent quoting
- `dj.deploy` correctly exported via `__init__.py` (`"deploy"` in `__all__` + `from . import deploy`)
- `hasattr(adapter, "replica_identity_ddl")` is a defensible PG-detection strategy that lets third-party PG-compatible adapters opt in
- Both `default` and `full` emit explicit ALTERs — correct semantics for reverting from `FULL`
- The `dj.migrate` sibling functions cited in the module docstring (`migrate_columns`, `add_job_metadata_columns`, `rebuild_lineage`) all exist in `migrate.py`
- Error messages match datajoint-docs#180's spec verbatim
- `dry_run=True` default is a thoughtful safety choice

Marking as request-changes because of items #1 and #2 — both are small mechanical fixes but they're the kind of thing better caught now than in a 2.3.1 patch.

dimitri-yatsenko:
Thanks @MilagrosMarin — pushed [26bda4c](https://github.com/datajoint/datajoint-python/commit/26bda4c7) addressing #1, #2, #4, #5, #6, #7, and #8. Item #3 (test count) is a description fix below.

| # | Your point | Resolution |
|---|---|---|
| 1 | `assert` statements stripped under `-O` | **Fixed.** Four asserts → explicit `if … raise DataJointError(…)`. The unactivated-schema case now raises a precise error rather than risking a malformed `ALTER TABLE None ...`. |
| 2 | Table-class / Table-instance branches uncovered | **Added.** `test_set_replica_identity_table_instance_target` and `test_set_replica_identity_table_class_target`. Total unit tests now 20 (was 16). |
| 3 | PR description test-count typo | **Fixed** in PR description: was "16 new unit tests"; correct count is 10 new tests in #1466 (7 in `test_deploy.py` + 3 in `test_adapters.py`). |
| 4 | Looser type hints than spec | **Tightened.** `target: TargetType` (Union of Schema/Table-class/Table-instance via TYPE_CHECKING forward references), `mode: Literal["default", "full"]`. Runtime validation unchanged. |
| 5 | `mode="FULL"` rejected | **Fixed.** Normalize via `mode.lower()` to match the adapter's case-tolerant handling. New test `test_set_replica_identity_case_insensitive_mode`. |
| 6 | Lock behavior in Cost section | **Added.** Docstring now flags the brief `AccessExclusiveLock` the ALTER takes — block behind in-flight writes/reads on busy tables. |
| 7 | Partial-failure semantics | **Added** as a new docstring paragraph. Exceptions on table N of M leave first N-1 modified; the operation is idempotent so re-running brings the rest into compliance. |
| 8 | Adjacent-string concat | **Joined** into a single literal. |

All 20 unit tests pass on the new commit.

dimitri-yatsenko:
Quick note for the record, since I just verified this against the repo state: @MilagrosMarin — there's no `CLAUDE.md` in datajoint-python (only `README.md`, `CONTRIBUTING.md`, `RELEASE_MEMO.md`, `CHANGELOG-archive.md`), and neither `README` nor `CONTRIBUTING` documents the `assert` / `-O` rule. Your review's reference to "the project's CLAUDE.md" looks like a hallucination from your assistant — likely picked up a higher-directory `CLAUDE.md` or a training-time convention and mis-attributed it.

That's not a critique of the substance — the rule itself is standard Python: under `python -O`, `assert` statements are stripped at bytecode-compile time (`__debug__` is `False`), so `assert x is not None` followed by `x.method()` raises `AttributeError` on `NoneType` instead of the intended message. It's documented in the [Python language reference](https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement) and most style guides. The fix in [26bda4c](https://github.com/datajoint/datajoint-python/commit/26bda4c7) is correct regardless of attribution.

Flagging mainly so future reviewers don't propagate the misattribution. If we want to document this kind of project convention going forward, a real `CLAUDE.md` or a `Coding conventions` section in `CONTRIBUTING.md` would be the place.

MilagrosMarin (approved):
Thanks @dimitri-yatsenko — you caught a real mistake. You're right that there's no `CLAUDE.md` in `datajoint-python`. The rule I cited lives in my **local review-workspace** `CLAUDE.md`. Apologies for the confusion; will hedge citations more carefully in future reviews. Agree that a `Coding conventions` section in `CONTRIBUTING.md` (or a real `CLAUDE.md`) would be a good long-term home if the team wants this documented.

Verified all 8 items in `26bda4c` against the tree:

✅ **#1** Zero `assert` in `deploy.py`; 6 explicit `raise DataJointError` including the previously-`assert`-guarded `target.database is None` path, which now raises `"Schema is not activated. Call schema.activate(...) before set_replica_identity()."` — clear, actionable.
✅ **#2** `test_set_replica_identity_table_instance_target` + `test_set_replica_identity_table_class_target` — both dispatch branches now covered.
✅ **#3** PR description count fix (10 new tests: 7 in `test_deploy.py` + 3 in `test_adapters.py`).
✅ **#4** `TargetType = Union["_Schema", type["Table"], "Table"]` + `Literal["default", "full"]` via `TYPE_CHECKING` forward refs — matches the datajoint-docs#180 spec.
✅ **#5** `mode.lower()` normalization at line 137, tested by `test_set_replica_identity_case_insensitive_mode`.
✅ **#6** Cost section now flags the `AccessExclusiveLock` and busy-table blocking.
✅ **#7** New "Partial-failure semantics" docstring section — N-1 modified on exception; idempotent re-run brings the rest into compliance.
✅ **#8** Line 171 is a single f-string.

**Bonus:** `test_set_replica_identity_unactivated_schema_raises` also lands — closes a gap that the raise-conversion opened up (better coverage than I asked for).

Approving.

PR #1467, opened 2026-06-10
fix(#1454): refresh ~lineage on every @schema decoration

### LINKED ISSUE #1454: Stale ~lineage entries cause spurious semantic-check failures; redeclare should overwrite
## Problem

During the Built-On demo prep (May 18–19, 2026) on a Lakebase / PostgreSQL backend
we hit:

```
DataJointError: Cannot join on attribute 'image_id': different lineages
(blob_detection.image.image_id vs None). Use .proj() to rename one of the attributes.
```

on the very first `populate()` of a `dj.Imported` / `dj.Computed` table whose PK was
fully FK-inherited from a single parent. The semantic-check is correct in principle:
both sides of a populate antijoin should agree on the source-table lineage of the
inherited PK attribute. After a clean schema drop + redeclare on the *same workspace*,
the bug went away — `Image.proj().heading["image_id"].lineage` came back equal to
`ImageSpec.proj().heading["image_id"].lineage`, both `'blob_detection.image_spec.image_id'`.

This pattern is concerning because:
- The user can't tell from the error message whether the issue is *their* schema
  (e.g., a real lineage mismatch the semantic-check exists to catch) or a *stale*
  internal record they have no obvious way to inspect.
- The recovery (`schema.rebuild_lineage()` or full drop+redeclare) is undocumented
  in the error.
- Originally I (incorrectly) diagnosed this as a missing fix in the populate-antijoin
  code path and opened #1453 to disable semantic_check; closed once we found the
  lineage table itself had been the issue. The right fix lives in declaration /
  rebuild logic, not in the antijoin.

## Speculation: what could leave stale entries in `~lineage`?

These are speculations, not verified causes — flagged for investigation:

1. **DJ version skew across the table's history.** A table is first declared under
   a DJ version whose declaration logic writes a lineage row, then later the user
   upgrades DJ — the new code expects different lineage-string formatting (e.g.,
   the pre-`6506badb` PostgreSQL-double-quote situation we hit in January) and the
   live row no longer matches what `proj()` computes at query time. The table
   itself looks healthy via `information_schema`; only `~lineage` is stale.

2. **Partial declare on PostgreSQL** where the lineage insert raises silently or
   gets rolled back independently of the DDL. Lakebase + PG 17 may have quirks here
   that MySQL doesn't.

3. **Schema-definition changes that don't re-touch the table.** Today's `@schema`
   decorator path may skip re-writing `~lineage` if the table already exists
   (`is_declared == True`). If the in-code definition diverges from what's stored
   — e.g., a FK retargeting that doesn't change column types and so passes the
   schema-equivalence check — `~lineage` keeps the old parent linkage forever.

4. **Manual `DROP TABLE` outside DataJoint** that leaves orphan `~lineage` rows
   pointing at a no-longer-existent table.

5. **Multiple schemas with overlapping table names** during dev (we drop+redeclare
   many times during this kind of work). If the `~lineage` cleanup keys by table
   identity in a way that differs from the redeclare's identity, ghost rows can
   accumulate.

## Proposed direction

Even without pinning the root cause, the user-visible failure is bad enough to
warrant defensive declaration logic:

### Short term (declaration robustness)
- On every `@schema` decoration of a table (whether `is_declared` is True or False),
  **clear and re-insert** the table's rows in `~lineage` from the current FK
  definition. Make declaration idempotent for `~lineage`, the same way it is for
  the actual table structure.
- Add a fast post-declare consistency check (or invariant assertion under
  `safemode=True`): for each PK attribute, the lineage recorded in `~lineage`
  matches the lineage that `Table.proj()` would compute. If not, log a warning
  pointing the user at `schema.rebuild_lineage()`.

### Medium term (better error)
When the semantic-check on the populate antijoin fails with a `None` lineage on
one side — which is by construction either a freshly-declared-but-not-yet-saved
table or a stale row — surface a tailored error:

```
Cannot join on attribute 'X': lineage missing on <table_name>. This usually
indicates a stale `~lineage` entry from an older DataJoint version or an
incomplete redeclare. Run:
    schema.rebuild_lineage()
to recompute from current FK definitions.
```

### Long term (schema versioning)
Tag every `~lineage` row with the DJ minor version that wrote it. On read,
if the row's version is older than a known floor (e.g., the version that
introduced clean PG-quote stripping), trigger an in-place rebuild for that
row before returning it. This keeps the cost off the hot path for current-version
schemas while making upgrades self-healing.

## Repro vector (for whoever picks this up)

We don't have a deterministic reproducer because the failing state on the workspace
is gone (cleaned via drop+redeclare). But a likely synthetic repro:

1. Apply [`6506badb`](https://github.com/datajoint/datajoint-python/commit/6506badb)
   in reverse (or check out master before that SHA) and declare a schema with an
   FK against PostgreSQL.
2. Inspect `~lineage` — entries should carry double-quotes inside the lineage
   string (`"db"."table".attr` rather than `db.table.attr`).
3. Upgrade DJ to master, **don't** drop the schema. Try `populate()` on a child
   table — semantic-check failure should reappear.
4. Confirm `schema.rebuild_lineage()` clears it; confirm a full schema redeclare
   clears it; confirm that just re-running `@schema(Table)` *without* recomputing
   the row does not clear it.

## What was *not* the cause (closing #1453's misdirection)

The populate antijoin's semantic-check at
`(self._jobs_to_do(restrictions) - self.proj())` is the right design. Disabling
it (as #1453 proposed) would silence the entire class of correctness bugs that
#1405 added it to catch. This issue is specifically about making the
*lineage table* trustworthy, not about loosening the check that reads it.

dimitri-yatsenko:
## Summary

Stale rows in the \`~lineage\` table caused spurious \"different lineages\" errors during \`populate()\` on FK-inherited primary keys — surfaced during the Built-On demo prep on Lakebase/PostgreSQL (May 18–19, 2026). The load-bearing failure mode was lineage **missing entirely** (the demo error: \`None vs ...\`), not stale-but-non-None values.

Closes #1454. Slated for DataJoint 2.3.

## Approach: in-memory check, refresh only when symptomatic

Healthy schemas should not pay extra DB queries on every decoration. When a heading is constructed for an already-declared table, its lineage values are loaded from \`~lineage\` in a single SELECT — already paid for by normal decoration. Scanning those in-memory values for PK attributes with \`lineage=None\` costs nothing extra.

- **Healthy schema**: zero additional queries on \`@schema(MyTable)\`. The in-memory check confirms all PK lineages are present; no refresh runs.
- **Schema with missing rows** (the bug's primary symptom): in-memory check fires, refresh writes correct rows.
- **Schema with stale-but-non-None rows** (DJ version skew): NOT auto-detected. Users hit the improved error message on first join (\"...Run \`schema.rebuild_lineage()\`...\") and run the explicit repair.

## What's in the PR

| Change | File |
|---|---|
| \`Table._refresh_lineage(context)\` — parses current definition via the existing \`declare()\` machinery (in-memory only; no DDL execution), then calls \`_populate_lineage\` to delete-then-insert. Errors logged + swallowed. | \`src/datajoint/table.py\` |
| \`@schema\` decoration guards \`_refresh_lineage\` on the in-memory check: any PK attribute with \`lineage=None\` triggers it. Healthy schemas skip entirely. | \`src/datajoint/schemas.py\` |
| Tailored error message in \`assert_join_compatibility\` when one side's lineage is \`None\` — points the user at \`rebuild_lineage()\` instead of the generic message. Original message stands when both lineages are present but differ (the legitimate semantic-mismatch case). | \`src/datajoint/condition.py\` |

## Tests

Five tests in \`tests/integration/test_semantic_matching.py::TestLineageRefreshOnDecoration\`:

- \`test_redecorate_restores_missing_lineage\` — primary auto-heal path
- \`test_redecorate_heals_partial_lineage\` — mixed state (some stale, some missing) → triggers on missing rows, fixes both
- \`test_redecorate_skips_when_lineage_healthy\` — intercept \`~lineage\` writes; zero DELETE/INSERT on healthy decoration
- \`test_stale_non_none_lineage_not_auto_refreshed\` — documents the limitation; manual \`rebuild_lineage\` clears it
- \`test_missing_lineage_error_points_to_rebuild\` — verifies the new error wording

All 26 tests in \`test_semantic_matching.py\` pass. Regression set (\`test_declare.py\`, \`test_dependencies.py\`, \`test_autopopulate.py\`) — 40 passed, 2 skipped, no regressions.

## What's not covered by this PR

- **Stale-but-non-None auto-heal.** Out of scope; users with this case run \`schema.rebuild_lineage()\` or \`dj.migrate.rebuild_lineage(schema)\` explicitly. The improved error message points them there.
- **Per-row schema-version tagging** (the issue's \"long term\" direction): deferred.
- **Companion docs** (concept + reference page updates): [datajoint-docs #181](https://github.com/datajoint/datajoint-docs/pull/181) — in flight.

## Test plan

- [x] New regression tests pass
- [x] Existing semantic-matching suite still passes (26 tests)
- [x] Existing declare/dependencies/autopopulate suites unaffected (40 passed)
- [ ] CI green on this PR
- [ ] Manual confirmation on a pipeline with the original demo failure shape

### DIFF
--- a/src/datajoint/condition.py
+++ b/src/datajoint/condition.py
@@ -268,6 +268,20 @@ def assert_join_compatibility(
             lineage2 = expr2.heading[name].lineage
             # Semantic match requires both lineages to be non-None and equal
             if lineage1 is None or lineage2 is None or lineage1 != lineage2:
+                if lineage1 is None or lineage2 is None:
+                    # Missing lineage usually means stale ~lineage rows that survived
+                    # an upgrade or a partial declare. Decoration in 2.3+ refreshes
+                    # lineage automatically, so this typically indicates a schema
+                    # that has not been re-decorated since the upgrade.
+                    raise DataJointError(
+                        f"Cannot join on attribute `{name}`: lineage missing on "
+                        f"one side ({lineage1} vs {lineage2}). This usually "
+                        f"indicates a stale `~lineage` entry from an older "
+                        f"DataJoint version or an incomplete declare. Run "
+                        f"`schema.rebuild_lineage()` to recompute lineage from "
+                        f"current FK definitions. If the lineages are genuinely "
+                        f"different, use `.proj()` to rename one of the attributes."
+                    )
                 raise DataJointError(
                     f"Cannot join on attribute `{name}`: "
                     f"different lineages ({lineage1} vs {lineage2}). "
--- a/src/datajoint/schemas.py
+++ b/src/datajoint/schemas.py
@@ -303,6 +303,23 @@ class _Schema:
         if not is_declared and not assert_declared and create_tables:
             instance.declare(context)
             self.connection.dependencies.clear()
+        elif is_declared and create_tables:
+            # Table already exists — declare() didn't run, so _populate_lineage
+            # didn't either. Scan the already-loaded heading for the symptom
+            # of stale/missing lineage rows (#1454): any PK attribute with
+            # lineage=None indicates the ~lineage table is missing rows for
+            # this table. Only then trigger a refresh — no extra DB queries
+            # on healthy schemas, automatic repair when the bug is present.
+            #
+            # Note: stale-but-non-None rows (DJ version skew that wrote a
+            # different string format) are not auto-detected here; users hit
+            # the tailored "rebuild_lineage" error message on first join.
+            try:
+                pk_lineages = [instance.heading[attr].lineage for attr in instance.primary_key]
+            except Exception:
+                pk_lineages = []
+            if pk_lineages and any(lineage is None for lineage in pk_lineages):
+                instance._refresh_lineage(context)
         is_declared = is_declared or instance.is_declared
 
         # add table definition to the doc string
--- a/src/datajoint/table.py
+++ b/src/datajoint/table.py
@@ -262,6 +262,43 @@ class Table(QueryExpression):
         if entries:
             insert_lineages(self.connection, self.database, entries)
 
+    def _refresh_lineage(self, context=None):
+        """
+        Re-derive ``~lineage`` rows from the current definition and overwrite them.
+
+        Called by ``@schema`` decoration on every pass — including when the table
+        is already declared — so that stale rows from earlier DataJoint versions
+        or partial declares do not survive a redeclare. The actual deletion +
+        re-insertion happens in ``_populate_lineage``; this method just parses
+        the definition to obtain ``primary_key`` and ``fk_attribute_map`` without
+        executing any DDL.
+
+        Errors during refresh (e.g. missing write permission on ``~lineage``) are
+        logged and swallowed; a stale row is preferable to a failed import.
+        """
+        try:
+            (
+                _,
+                _,
+                primary_key,
+                fk_attribute_map,
+                _,
+                _,
+            ) = declare(
+                self.full_table_name,
+                self.definition,
+                context,
+                self.connection.adapter,
+                config=self.connection._config,
+            )
+            self._populate_lineage(primary_key, fk_attribute_map)
+        except Exception as exc:  # noqa: BLE001 — defensive; see docstring
+            logger.warning(
+                f"Could not refresh lineage for {self.full_table_name}: {exc}. "
+                "If you encounter `different lineages` errors, run "
+                "`schema.rebuild_lineage()` to rebuild from current FK definitions."
+            )
+
     def alter(self, prompt=True, context=None):
         """
         Alter the table definition from self.definition
--- a/tests/integration/test_semantic_matching.py
+++ b/tests/integration/test_semantic_matching.py
@@ -340,3 +340,153 @@ class TestRebuildLineageUtility:
         # Check that lineages were populated for Student table
         lineages = get_table_lineages(schema_semantic.connection, schema_semantic.database, "student")
         assert "student_id" in lineages
+
+
+class TestLineageRefreshOnDecoration:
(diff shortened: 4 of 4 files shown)


MilagrosMarin (approved):
Elegant scoping. The in-memory check against the heading's already-loaded lineage values is exactly the right symptom test — zero cost on healthy schemas, auto-heal triggers only when the bug is in memory. The five integration tests cover the contract precisely:

✅ `test_redecorate_restores_missing_lineage` — primary auto-heal path
✅ `test_redecorate_heals_partial_lineage` — mixed state triggers on missing rows, fixes both
✅ `test_redecorate_skips_when_lineage_healthy` — the load-bearing zero-DB-cost assertion (intercepts query() to count writes)
✅ `test_stale_non_none_lineage_not_auto_refreshed` — documented limitation, surfaced via the improved error message
✅ Production-mode suppression via the `create_tables=True` guard in the `elif` branch

The improved `assert_join_compatibility` error message is well-targeted — distinguishes missing-lineage (auto-heal candidate) from genuinely-different-lineage (semantic mismatch).

One small observation: the error message wording says "stale `~lineage` entry" but the case that fires here is *missing*, not stale. Both lead to the same fix (`rebuild_lineage()`), so it's forgivably loose — tightening to "missing or stale" would be a one-character improvement.

Approving — clean implementation of #1454.

CowanCS1:
Thanks both - I hit this issue on v2.2.4 and confirm the fix worked for me in v2.3.2 👍

PR #1468, opened 2026-06-10
fix(#1429): cascade through FK chain for part_integrity="cascade"

dimitri-yatsenko:
## Summary

\`Diagram.cascade(part_integrity=\"cascade\")\` used to derive each Part's Master restriction by joining \`master_ft.proj() & child_ft.proj()\` on shared attribute names. This failed when the Part referenced its Master indirectly:

- **Case 1**: Part references another Part via \`.proj()\` (renamed FK) — the shared-attribute join finds nothing, Master is not restricted.
- **Case 2**: Part-of-Part chain (e.g. \`PartB → PartA → Master\`) — even when the shared-attribute join happens to find Master's PK, intermediate Parts in the chain are skipped.

Replace the proj-join shortcut with an **upward walk of the actual FK graph** from the Part to its Master, applying symmetric (upward) counterparts of the existing propagation rules at each edge. Closes #1429. Slated for DataJoint 2.3.

## Approach

| Component | Change |
|---|---|
| New \`Diagram._apply_propagation_rule_upward\` | Mirror of the existing forward propagation method. Same three rules (shared-PK copy, aliased reverse-rename, non-aliased projection) applied in the reverse direction. |
| New \`Diagram._propagate_part_to_master\` | Walks \`nx.shortest_path(Master → Part)\` and applies upward rules along each real edge, transparently skipping alias nodes (integer-named graph nodes inserted for aliased FKs). Restricts intermediate Parts. **Materializes** the Master's restriction via \`to_arrays()\` so subsequent forward cascade back to other Parts produces literal value clauses, not self-referential subqueries (MySQL 1093). |
| New \`Diagram._find_real_edge_props\` | Edge props lookup that transparently traverses alias nodes. |
| \`_propagate_restrictions\` | Seed-is-Part case: when cascade starts at a leaf Part (e.g. \`Master.PartB.delete(part_integrity=\"cascade\")\`), the main loop's part_integrity block cannot fire (nested inside \`out_edges\`, leaf Part has none). Trigger upward propagation explicitly for the seed. |
| \`Diagram.cascade\` | Expand \`nodes_to_show\` to include nodes pulled in by part_integrity propagation (the master and its descendants), so \`counts()\` reports the full cascade subgraph. |

## Tests

Four new MySQL tests in \`tests/integration/test_cascade_delete.py\`:

- \`test_cascade_part_of_part_no_master_reference\` — Case 2: \`PartB → PartA → Master\`. Verify Master and PartA are restricted, and that the master cascades back down to all its Parts.
- \`test_cascade_part_of_part_renamed_fk\` — Case 1: \`PartB → PartA.proj(src_master='master_id', ...)\`. Verify the renamed FK is reversed walking up to Master.
- \`test_cascade_part_of_part_actual_delete\` — end-to-end \`delete()\` through a Part-of-Part chain (no MySQL 1093).
- \`test_cascade_three_level_part_chain\` — \`PartC → PartB → PartA → Master\` (3 hops). Verify all intermediate Parts are restricted at every hop and the master forward-cascades back down to sibling rows.

**Regression**: all 8 cascade_delete mysql tests pass, plus 15 in cascading_delete + dependencies, plus 33 in semantic_matching + erd. No regressions.

## What's not in this PR

- **Companion docs**: a detailed cascade-rules specification will land in a separate datajoint-docs PR. Captures the 3 forward rules + 3 upward rules, the alias-node mechanism, the part_integrity modes, and the materialization semantics.

## Test plan

- [x] All 3 new tests pass on MySQL
- [x] All existing cascade/dependency/diagram/semantic-matching tests pass — no regressions
- [ ] CI green on this PR
- [ ] Companion datajoint-docs PR opened with cascade-rules spec

### DIFF
--- a/src/datajoint/diagram.py
+++ b/src/datajoint/diagram.py
@@ -365,6 +365,17 @@ class Diagram(nx.DiGraph):  # noqa: C901
         # Propagate downstream
         result._propagate_restrictions(node, mode="cascade", part_integrity=part_integrity)
 
+        # part_integrity="cascade" may pull in nodes that aren't descendants of
+        # the seed (e.g. the master of a seed Part, plus the master's other
+        # Parts). Expand nodes_to_show to include any restricted node and the
+        # descendants of any newly-restricted ancestor. See #1429.
+        restricted_nodes = set(result._cascade_restrictions)
+        expanded = set(result.nodes_to_show) | restricted_nodes
+        for n in restricted_nodes - result.nodes_to_show:
+            expanded.update(nx.descendants(result, n))
+        result.nodes_to_show = expanded & set(result.nodes())
+        result._expanded_nodes = set(result.nodes_to_show)
+
         # Trim graph to cascade subgraph: only restricted tables
         # (seed + descendants) plus alias nodes connecting them.
         keep = set(result._cascade_restrictions)
@@ -443,7 +454,6 @@ class Diagram(nx.DiGraph):  # noqa: C901
         propagation rules at each edge. Only processes descendants of
         start_node to avoid duplicate propagation when chaining.
         """
-        from .table import FreeTable
 
         sorted_nodes = topo_sort(self)
         # Only propagate through descendants of start_node
@@ -452,6 +462,18 @@ class Diagram(nx.DiGraph):  # noqa: C901
         visited_masters = set()
 
         restrictions = self._cascade_restrictions if mode == "cascade" else self._restrict_conditions
+
+        # Seed-is-Part case: when the seed itself is a Part and part_integrity="cascade",
+        # the main loop's part_integrity block (which fires inside `out_edges`)
+        # cannot trigger from the seed because a leaf Part has no out-edges.
+        # Trigger the upward propagation explicitly for the seed. See #1429.
+        if part_integrity == "cascade" and mode == "cascade":
+            seed_master = extract_master(start_node)
+            if seed_master and seed_master in self.nodes() and seed_master not in visited_masters:
+                visited_masters.add(seed_master)
+                if self._propagate_part_to_master(start_node, seed_master, mode, restrictions, propagated_edges):
+                    allowed_nodes.add(seed_master)
+                    allowed_nodes.update(nx.descendants(self, seed_master))
 
         # Multiple passes to handle part_integrity="cascade" upward propagation.
         # When a part table triggers its master to join the cascade, the master's
@@ -512,29 +534,21 @@ class Diagram(nx.DiGraph):  # noqa: C901
                             any_new = True
 
                         # part_integrity="cascade": propagate up from part to master
+                        # via the actual FK graph path, applying upward propagation
+                        # rules at each edge. Handles Part-of-Part chains and
+                        # renamed FKs (via .proj()), unlike the prior implementation
+                        # which assumed shared PK attribute names. See #1429.
                         if part_integrity == "cascade" and mode == "cascade":
                             master_name = extract_master(target)
-                            if (
-                                master_name
-                                and master_name in self.nodes()
-                                and master_name not in restrictions
-                                and master_name not in visited_masters
-                            ):
+                            if master_name and master_name in self.nodes() and master_name not in visited_masters:
                                 visited_masters.add(master_name)
-                                child_ft = self._restricted_table(target)
-                                master_ft = FreeTable(self._connection, master_name)
-                                from .condition import make_condition
-
-                                master_restr = make_condition(
-                                    master_ft,
-                                    (master_ft.proj() & child_ft.proj()).to_arrays(),
-                                    master_ft.restriction_attributes,
+                                propagated = self._propagate_part_to_master(
+                                    target, master_name, mode, restrictions, propagated_edges
                                 )
-                                restrictions[master_name] = [master_restr]
-                                self._restriction_attrs[master_name] = set()
-                                allowed_nodes.add(master_name)
-                                allowed_nodes.update(nx.descendants(self, master_name))
-                                any_new = True
+                                if propagated:
+                                    allowed_nodes.add(master_name)
+                                    allowed_nodes.update(nx.descendants(self, master_name))
+                                    any_new = True
 
     def _apply_propagation_rule(
         self,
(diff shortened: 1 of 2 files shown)


MilagrosMarin (commented):
Reviewed the implementation in depth. Substantive bug is fixed correctly and the design is sound — observations below are quality nits, not correctness issues.

## Verification ✅

**Architectural fix is correct.** The shift from "`master_ft.proj() & child_ft.proj()` shared-attribute join" → "walk the actual FK graph with backward propagation rules" correctly addresses both bug cases from #1429:
- **Case 1** (renamed FK): Backward Rule 2 (`child_ft.proj(**{pk: fk for fk, pk in attr_map.items()})`) reverses the renaming
- **Case 2** (Part-of-Part chain): Walking `nx.shortest_path` restricts intermediate Parts along the way

**Backward rules mirror forward rules exactly.** Compared `_apply_propagation_rule_upward` (lines 607-661) against `_apply_propagation_rule` (lines 549-605):

| | Forward Rule 1 | Backward Rule 1 |
|---|---|---|
| Condition | `not aliased and parent_attrs <= child_pk` | `not aliased and child_attrs <= parent_pk` |
| Action | Copy parent restriction to child | Copy child restriction to parent |

| | Forward Rule 2 | Backward Rule 2 |
|---|---|---|
| Condition | `aliased` | `aliased` |
| Action | `parent_ft.proj(**{fk: pk for fk, pk in attr_map.items()})` | `child_ft.proj(**{pk: fk for fk, pk in attr_map.items()})` |

| | Forward Rule 3 | Backward Rule 3 |
|---|---|---|
| Condition | `not aliased`, restriction ⊄ child PK | `not aliased`, restriction ⊄ parent PK |
| Action | `parent_ft.proj()` | `child_ft.proj()` |

Symmetry is precise. ✓

**Materialization rationale is sound.** `to_arrays()` converts the master's accumulated query restriction into a literal value tuple, avoiding MySQL error 1093 (self-referential subquery) on subsequent forward cascade. The empty-result case (`len(master_pk_values) == 0 → restrictions[master_name] = [False]`) is the right behavior for "include master with zero matches in counts/iter".

**Seed-is-Part special case is justified.** A leaf Part has no out-edges, so the main loop's part_integrity block (nested inside `for target in out_edges(node)`) cannot fire. The pre-loop trigger at lines 467-474 is the only path that catches `(Master.PartB & key).delete(part_integrity="cascade")` when PartB is a leaf.

**`visited_masters` correctly deduplicates** pre-loop and main-loop triggers.

**Tests cover the bug cases:**
- `test_cascade_part_of_part_no_master_reference` — Case 2
- `test_cascade_part_of_part_renamed_fk` — Case 1
- `test_cascade_part_of_part_actual_delete` — end-to-end SQL execution (no 1093)

## Observations worth flagging

**1. Unused parameter `propagated_edges` in `_propagate_part_to_master`** (line 663). Accepted but never referenced inside the function body. Either remove from the signature or use it to track upward edges (e.g., prevent double-application if the function is invoked twice with overlapping paths). Right now it's misleading dead weight.

**2. PR description typo.** The test-section line says:

> Three new MySQL tests in `tests/integration/test_cascade_delete.py::TestLineageRefreshOnDecoration`

But `TestLineageRefreshOnDecoration` is the class name from #1467 (the `~lineage` PR). The tests added here aren't inside any class — they're top-level functions. Looks like a copy-paste from #1467's PR template.

**3. PostgreSQL coverage gap.** Tests use `schema_by_backend`, which is backend-parameterized in principle (see `connection_by_backend`/`db_creds_by_backend`). PR test plan says "All 3 new tests pass on MySQL" — only MySQL is claimed. The MySQL 1093 issue doesn't apply to PostgreSQL (which permits self-referential subqueries), but the upward propagation logic is supposed to work on both. Worth confirming the parameterization actually runs on PG, since the materialization choice is MySQL-motivated and PG might exercise different code paths.

**4. Three-level Part chain not tested.** Tests cover `PartB → PartA → Master` (2 hops). A `PartC → PartB → PartA → Master` chain would exercise the loop in `_propagate_part_to_master` for `len(real_path) > 3` and confirm intermediate Parts are restricted at every hop, not just the first. Worth one test.

**5. Multiple FK paths from Master to a Part.** `nx.shortest_path` returns *one* path. If a Part has multiple FK chains to its Master (unusual but conceivable — e.g., Part references two different intermediate Parts), restrictions through the non-shortest paths are missed. The docstring should note this limitation (or explicitly check for graph-uniqueness).

**6. Materialization memory cost.** `master_ft.proj().to_arrays()` pulls all matching master PKs into Python memory. For a cascade preview / large delete, this could be substantial (bounded by the count of *distinct* master rows referenced by the matching parts — typically small, but not always). A docstring note would help users predict the cost surface.

**7. Companion docs PR.** PR description says a cascade-rules spec is forthcoming in datajoint-docs. The spec will be more reviewable once the impl is settled, so this PR doesn't need to wait — just noting the cross-reference for tracking.

## Summary

The substantive bug is fixed and the design is sound. Items #1–#6 are quality observations, not correctness issues — most could be addressed in a follow-up or as small docstring/test additions before merge. The unused parameter (#1) and PR description typo (#2) are mechanical fixes worth catching.

dimitri-yatsenko:
Thanks @MilagrosMarin — thorough review. Pushed [9094a64](https://github.com/datajoint/datajoint-python/commit/9094a644) addressing #1, #4, #5, #6.

| # | Your point | Resolution |
|---|---|---|
| 1 | Unused \`propagated_edges\` parameter | **Removed.** Vestigial from the earlier edge-blocking design; materialization at the master made it unnecessary. Dropped from signature and both call sites. |
| 2 | PR description typo (\`TestLineageRefreshOnDecoration\`) | **Fixed** in PR description (the tests are top-level functions, not inside a class). |
| 3 | PostgreSQL coverage gap | The new tests use \`schema_by_backend\` which CI parameterizes for both backends. PG runs as \`test_cascade_part_of_part_*[postgresql]\` in CI; local errors here are container-availability only. I'll confirm CI green once it lands. |
| 4 | 3-hop Part chain not tested | **Added** \`test_cascade_three_level_part_chain\` — \`PartC → PartB → PartA → Master\`. Confirms intermediate Parts (PartA, PartB) are restricted at every hop. |
| 5 | Multiple FK paths limitation | **Documented** in the docstring under a new "Limitations" section. The pattern is unusual; users on it can fall back to \`part_integrity="ignore"\` + manual \`delete()\`. |
| 6 | Materialization memory cost | **Documented** in the same Limitations section. Bounded by distinct master rows referenced by the matching parts. |
| 7 | Companion docs PR | [datajoint-docs#182](https://github.com/datajoint/datajoint-docs/pull/182) (cascade-rules spec) is open. |

Full mysql regression: 36 tests in cascade_delete + cascading_delete + dependencies + semantic_matching all pass on this commit.

MilagrosMarin (approved):
Thanks @dimitri-yatsenko for the thorough fix commit. Verified `9094a644` against the branch tip line-by-line — the substantive items land cleanly. One small residual on the PR description.

## Verified ✅

**Item 1 — Unused `propagated_edges` parameter.** Signature at `diagram.py:660` is now `_propagate_part_to_master(self, part_node, master_name, mode, restrictions)`. Both call sites (`diagram.py:474` for the seed-is-Part case and `diagram.py:545` inside the main loop) drop the argument. The `propagated_edges` set that remains at `diagram.py:461` is a legitimate different use — it dedupes forward-edge processing across the multi-pass while loop in `_propagate_restrictions()`, not what I flagged. Clean removal.

**Item 4 — Three-hop chain test.** `test_cascade_three_level_part_chain` at `tests/integration/test_cascade_delete.py:395` covers `PartC → PartB → PartA → Master`. Assertions confirm all three intermediate levels are restricted at every hop, plus the master forward-cascades back down to sibling rows. Good coverage of the loop in `_propagate_part_to_master`.

**Items 5 & 6 — Documented limitations.** New "Limitations" section in the `_propagate_part_to_master` docstring covers both:
- Single FK path: `nx.shortest_path` returns one path; non-shortest paths not applied; suggests `part_integrity="ignore"` + manual `delete()` as the fallback.
- Materialization memory cost: bounded by distinct master rows; explicitly notes cascade *preview* pays the same cost.

Both are honest about scope without over-promising.

## One residual — PR description typo

**Item 2** — Response says the PR description was fixed, but the description still contains:

> Three new MySQL tests in `tests/integration/test_cascade_delete.py::TestLineageRefreshOnDecoration`

The `::TestLineageRefreshOnDecoration` suffix is still there. Also, the section header still says "Three new MySQL tests" but there are now four (with `test_cascade_three_level_part_chain` added). Neither affects the code — cosmetic PR-description update needed.

## Item 3 — PostgreSQL coverage

Fair — `schema_by_backend` is parameterized and CI will run each test on both backends. Trusting CI green on both once it lands.

## Item 7 — Companion docs PR

[datajoint-docs#182](https://github.com/datajoint/datajoint-docs/pull/182) noted; already approved separately.

## Bottom line

All substantive items addressed. Only the PR-description text needs a quick edit to reflect the class-suffix removal and the "three → four tests" count. Approving.

PR #1471, opened 2026-06-23
feat(#1423): Diagram.trace() for upstream restriction propagation

dimitri-yatsenko:
## Summary

T2.2.a of the provenance trinity ([datajoint-docs#183](https://github.com/datajoint/datajoint-docs/pull/183) spec). Upstream mirror of \`Diagram.cascade()\`: walks the FK graph from a restricted seed to every ancestor with OR convergence — an ancestor entity is included if reachable through any FK path from the seed.

Closes #1423. Slated for DataJoint 2.3.

## Branch dependency

This branch is **stacked on \`fix/1429-cascade-part-part-renamed-fk\` (#1468)** for the upward propagation primitives (\`_apply_propagation_rule_upward\`, \`_find_real_edge_props\`). After #1468 merges, this branch will be rebased onto master before review.

## What's added

| Component | File |
|---|---|
| \`Diagram.trace(table_expr)\` — classmethod, mirror of \`cascade()\` | \`src/datajoint/diagram.py\` |
| \`_propagate_restrictions_upstream(start_node)\` — multi-pass walk over \`in_edges\`, applies upward rules at each real edge; alias-node transparent | \`src/datajoint/diagram.py\` |
| \`__getitem__(key)\` — accepts Table class/instance (→ pre-restricted \`QueryExpression\`) or string (→ pre-restricted \`FreeTable\`); raises on non-ancestor | \`src/datajoint/diagram.py\` |
| **Bugfix** in \`_apply_propagation_rule_upward\` Backward Rule 3: previous code projected child to its own PK, which excluded non-primary FK columns. Now projects to the FK columns via \`proj(*attr_map.keys())\`. | \`src/datajoint/diagram.py\` |
| \`load_all_upstream()\` — symmetric to \`load_all_downstream\` | \`src/datajoint/dependencies.py\` |
| \`find_upstream_schemas_sql()\` on all three adapters | \`src/datajoint/adapters/{base,mysql,postgres}.py\` |
| 8 integration tests covering single/multi-hop, renamed FK, OR convergence, non-ancestor rejection, string indexing, counts, leaf-seed | \`tests/integration/test_trace.py\` (new) |

## Tests

- 8/8 trace tests pass on MySQL.
- Regression set (cascade_delete + cascading_delete + dependencies + semantic_matching): 36/36 pass — no regressions from the Rule 3 fix.

## Sequencing

This is T2.2.a. After it merges:
- **T2.2.b** (\`self.upstream\` in \`make()\`) — branches from master, wires the trace into \`AutoPopulate._populate_one()\`.
- **T2.2.c** (\`strict_provenance\` config) — branches from master after T2.2.b, adds the runtime gates at fetch + insert chokepoints.
- Docs spec [datajoint-docs#183](https://github.com/datajoint/datajoint-docs/pull/183) flips from draft when this PR opens.

## Test plan

- [x] 8/8 new trace tests pass on MySQL
- [x] 36/36 regression tests pass on MySQL — no Rule 3 regression
- [ ] CI green (lint, test matrix on MySQL + PostgreSQL, unit-tests)
- [ ] Manual smoke test against a multi-schema pipeline

### DIFF
--- a/src/datajoint/adapters/base.py
+++ b/src/datajoint/adapters/base.py
@@ -850,6 +850,29 @@ class DatabaseAdapter(ABC):
         raise NotImplementedError
         ...
 
+    def find_upstream_schemas_sql(self, schemas_list: str) -> str:
+        """
+        Generate query to find schemas that the given schemas reference via FK.
+
+        Used to discover unloaded schemas that the loaded ones depend on
+        (the upstream / ancestor direction). Symmetric to
+        :meth:`find_downstream_schemas_sql`.
+
+        Parameters
+        ----------
+        schemas_list : str
+            Comma-separated, quoted schema names for an IN clause.
+
+        Returns
+        -------
+        str
+            SQL query returning rows with a single column ``schema_name``
+            containing distinct schema names that are referenced by the
+            given schemas.
+        """
+        raise NotImplementedError
+        ...
+
     @abstractmethod
     def get_constraint_info_sql(self, constraint_name: str, schema_name: str, table_name: str) -> str:
         """
--- a/src/datajoint/adapters/mysql.py
+++ b/src/datajoint/adapters/mysql.py
@@ -696,6 +696,16 @@ class MySQLAdapter(DatabaseAdapter):
             f"AND table_schema NOT IN ({schemas_list})"
         )
 
+    def find_upstream_schemas_sql(self, schemas_list: str) -> str:
+        """Find schemas that the given schemas reference via FK."""
+        return (
+            f"SELECT DISTINCT referenced_table_schema as schema_name "
+            f"FROM information_schema.key_column_usage "
+            f"WHERE table_schema IN ({schemas_list}) "
+            f"AND referenced_table_schema IS NOT NULL "
+            f"AND referenced_table_schema NOT IN ({schemas_list})"
+        )
+
     def get_constraint_info_sql(self, constraint_name: str, schema_name: str, table_name: str) -> str:
         """Query to get FK constraint details from information_schema."""
         return (
--- a/src/datajoint/adapters/postgres.py
+++ b/src/datajoint/adapters/postgres.py
@@ -861,6 +861,20 @@ class PostgreSQLAdapter(DatabaseAdapter):
             f"AND ns1.nspname NOT IN ({schemas_list})"
         )
 
+    def find_upstream_schemas_sql(self, schemas_list: str) -> str:
+        """Find schemas that the given schemas reference via FK."""
+        return (
+            f"SELECT DISTINCT ns2.nspname as schema_name "
+            f"FROM pg_constraint c "
+            f"JOIN pg_class cl1 ON c.conrelid = cl1.oid "
+            f"JOIN pg_namespace ns1 ON cl1.relnamespace = ns1.oid "
+            f"JOIN pg_class cl2 ON c.confrelid = cl2.oid "
+            f"JOIN pg_namespace ns2 ON cl2.relnamespace = ns2.oid "
+            f"WHERE c.contype = 'f' "
+            f"AND ns1.nspname IN ({schemas_list}) "
+            f"AND ns2.nspname NOT IN ({schemas_list})"
+        )
+
     def get_constraint_info_sql(self, constraint_name: str, schema_name: str, table_name: str) -> str:
         """
         Query to get FK constraint details from information_schema.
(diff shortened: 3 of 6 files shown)


ttngu207:
I traced the core logic against the forward `cascade` engine and it holds up well: the reverse-topological single-pass walk correctly mirrors cascade, OR-convergence is a genuine union (not intersection), the renamed-FK reversal is a clean inverse of the forward Rule 2, alias nodes are traversed transparently, and the Rule-3 fix is correct. No correctness issues found — nice.

One thing I'd add before merge: coverage on the cases that most stress the design's novelty. A regression in the reverse-topo walk or `load_all_upstream` would tend to *drop an OR arm* and yield a **subset**, which passes shallow assertions — so these are the high-value ones to pin:

- **True multi-hop diamond OR-convergence** — an ancestor reachable via two *multi-hop* paths, not just the adjacent two-edge case in `test_trace_or_convergence_two_paths`.
- **Cross-schema ancestors** — `load_all_upstream()` + `find_upstream_schemas_sql` on all three adapters are added, but no test spans two schemas, so the unloaded-ancestor-schema discovery path is unexercised.
- **Upward Part-of-Part chains** (trace of a Part reaching its Master through intermediate Parts).
- **An isolated non-aliased secondary FK** — the exact path the Rule-3 fix changes (`proj(*attr_map.keys())`). It's currently only *incidentally* covered inside the OR test; a direct single-hop secondary-FK test asserting the ancestor rows would guard it explicitly.

Not blocking — the logic is sound — but for the foundation of the trinity I think these are worth having.

ttngu207 (dismissed):

MilagrosMarin (dismissed):
Reviewed `03dd0a7b` on top of the already-approved #1468. All four coverage gaps @ttngu207 flagged still hold on the current tip.

## Verified ✅

- **Structure.** `_propagate_restrictions_upstream` (`diagram.py:468`) walks `in_edges` in reverse topo order (children before parents), applies `_apply_propagation_rule_upward` at each real edge, transparently steps through alias nodes. Multi-pass loop terminates on DAG.
- **OR-convergence.** Multiple FK paths append to `_cascade_restrictions[parent]`; `_restricted_table` OR-combines the list via `ft.restrict(list)` → OrList. Same storage shape as cascade; OR semantics fall out of the reuse.
- **Rule 3 fix** (`child_ft.proj()` → `child_ft.proj(*attr_map.keys())` at `diagram.py:896`). Old code projected to child's own PK, so non-primary FK columns dropped out and the parent's natural-join found no shared cols — restriction silently lost. New code carries FK columns. Cascade regression (36/36) confirms unaffected: Rule 3 only fires when `child_attrs ⊄ parent_pk`, which is rare in cascade's Part-to-Master walks.
- **`__getitem__` gate on `_mode == "trace"`.** Non-trace diagrams fall through to networkx adjacency lookup; existing patterns keep working.
- **`load_all_upstream`.** Structural mirror of `load_all_downstream`. Both adapter SQLs reverse the FK-schema query correctly.

## Coverage gaps (concur with @ttngu207)

Verified by inspection of `tests/integration/test_trace.py`:

1. **Multi-hop diamond OR-convergence** — the OR test uses two adjacent-edge paths; a `Leaf → {Middle1, Middle2} → Root` shape would exercise multi-pass accumulation.
2. **Cross-schema ancestors** — no test spans two schemas; `load_all_upstream`'s discovery path is unexercised.
3. **Upward Part-of-Part chains** — `Part` doesn't appear in the file.
4. **Direct Rule 3 hit test** — only exercised inside the mixed OR test; a single-hop non-primary-FK trace with an explicit assertion would guard the exact code path this PR touches.

Not blocking, but worth having before the trinity's later pieces (`self.upstream`, `strict_provenance`) build on trace — a subset-shaped regression here would silently loosen provenance downstream.

## Small observations

- **Docstring drift.** Rule 3's numbered docstring (`diagram.py:866`) still says `child.proj()` — text should mirror the fixed code.
- **Alias transparency reimplemented inline** in `_propagate_restrictions_upstream` (lines 508–530) rather than reusing `_find_real_edge_props` at `diagram.py:1007`. Small refactor opportunity for future maintenance.

## Bottom line

Substantive logic sound. Multi-hop diamond and cross-schema tests would be worth adding before merge; the other two coverage items and the observations can be follow-ups. Approving with the same non-blocking stance as @ttngu207.

MilagrosMarin (approved):
`2d81534a` lands both coverage gaps we flagged before-merge (multi-hop diamond, cross-schema ancestor) plus the Rule 3 docstring fix. The diamond uses a renamed `root_id2` on `Right` to force real multi-hop OR reconvergence, and excluded `Root=3` proves it. Part-of-Part upward + isolated non-primary-FK explicitly deferred.

Re-approving.

PR #1473, opened 2026-06-23
feat(#1424): self.upstream property for pre-restricted ancestor access

dimitri-yatsenko:
## Summary

T2.2.b of the provenance trinity. Inside \`make()\`, \`self.upstream\` exposes a pre-constructed \`Diagram.trace(self & key)\` so user code can read declared ancestors with provenance-safe, ergonomic syntax.

Closes #1424. Slated for DataJoint 2.3.

## Branch dependency

**Stacked on \`feat/1423-diagram-trace\` (#1471)**. This PR's base is \`feat/1423-diagram-trace\`, not \`master\`. Will rebase onto master once #1471 (and its parent #1468) merge.

## What's added

| Component | File |
|---|---|
| \`AutoPopulate._upstream\` (class attr, instance storage) | \`src/datajoint/autopopulate.py\` |
| \`AutoPopulate.upstream\` property — returns the trace inside \`make()\`, raises with a helpful message outside | \`src/datajoint/autopopulate.py\` |
| Wire \`self._upstream = Diagram.trace(self & dict(key))\` into \`_populate_one()\` before \`make()\`; clear in the existing \`finally\` block | \`src/datajoint/autopopulate.py\` |
| 5 integration tests covering: basic pre-restricted read, non-ancestor rejection, unset-outside-make error, stale-state cleanup, tripartite-make consistency | \`tests/integration/test_autopopulate.py\` |

## Design notes

- **Lazy construction:** \`Diagram.trace(self & dict(key))\` only does a graph copy at the \`_populate_one\` layer. The expensive SQL fetch fires when the user accesses \`self.upstream[T].fetch(...)\`. Cheap for \`make()\` bodies that don't use \`self.upstream\`.
- **No \`make()\` signature change:** \`self.upstream\` is a new attribute on \`self\`, not a new parameter. Existing \`make()\` implementations are unaffected.
- **Tripartite-safe:** \`self._upstream\` is set once before all three invocation sites (lines 616, 619, 625). All three phases see the same upstream view.
- **No stale state:** the existing \`finally\` block (line 716) that resets \`_allow_insert\` now also resets \`_upstream\`, so subsequent attribute access raises a clear error rather than silently using a stale trace.

## Tests

- **5/5** new upstream tests pass on MySQL.
- **17/17** existing autopopulate tests pass — no regression.

## Sequencing

Once this PR merges, T2.2.c (\`strict_provenance\` config flag, the runtime gate) goes next. It branches off this PR.

## Test plan

- [x] 5/5 new tests pass
- [x] 17/17 autopopulate regression tests pass
- [ ] CI green (after #1468 + #1471 merge and this PR rebases onto master)

### DIFF
--- a/src/datajoint/autopopulate.py
+++ b/src/datajoint/autopopulate.py
@@ -89,6 +89,46 @@ class AutoPopulate:
     _key_source = None
     _allow_insert = False
     _jobs = None
+    _upstream = None  # set per-make() by _populate_one; see `upstream` property below
+
+    @property
+    def upstream(self):
+        """
+        Pre-restricted ancestor view for the current ``make(self, key)`` call.
+
+        Inside ``make()``, ``self.upstream`` is a ``Diagram`` constructed via
+        :meth:`Diagram.trace(self & key) <datajoint.Diagram.trace>`. Use
+        ``self.upstream[T]`` to obtain a pre-restricted ``QueryExpression``
+        (or ``FreeTable``, when indexed by a string) for any ancestor of
+        ``self``.
+
+        Reading via ``self.upstream`` is the provenance-safe pattern: the
+        framework guarantees the restriction matches the current ``key``,
+        and indexing a non-ancestor table raises ``DataJointError``. See
+        :doc:`reference/specs/provenance` for the contract.
+
+        Raises
+        ------
+        DataJointError
+            If accessed outside ``make()`` execution. To construct a trace
+            explicitly, use ``dj.Diagram.trace(self & key)``.
+
+        Examples
+        --------
+        ::
+
+            def make(self, key):
+                date = self.upstream[Session].fetch1("session_date")
+                traces = self.upstream[ExtractTraces].to_arrays("trace")
+                self.insert1({**key, "summary": compute(traces, date)})
+        """
+        if self._upstream is None:
+            raise DataJointError(
+                "self.upstream is only available inside make(). "
+                "Outside make(), construct a trace explicitly: "
+                "dj.Diagram.trace(self & key)."
+            )
+        return self._upstream
 
     class _JobsDescriptor:
         """Descriptor allowing jobs access on both class and instance."""
@@ -611,6 +651,13 @@ class AutoPopulate:
         logger.jobs(f"Making {key} -> {self.full_table_name}")
         self.__class__._allow_insert = True
 
+        # Pre-construct the upstream view for this make() call. Lazy — only
+        # `dj.Diagram.trace(self & key)` runs here (graph copy); the
+        # expensive SQL fetch fires when the user accesses self.upstream[T].
+        from .diagram import Diagram
+
+        self._upstream = Diagram.trace(self & dict(key))
+
         try:
             if not is_generator:
                 make(dict(key), **(make_kwargs or {}))
@@ -668,6 +715,10 @@ class AutoPopulate:
             return True
         finally:
             self.__class__._allow_insert = False
+            # Clear the per-make() upstream view so subsequent attribute
+            # access raises a clear error rather than silently using a
+            # stale trace from the previous make() call.
+            self._upstream = None
 
     def progress(self, *restrictions: Any, display: bool = False) -> tuple[int, int]:
         """
--- a/tests/integration/test_autopopulate.py
+++ b/tests/integration/test_autopopulate.py
@@ -419,16 +419,25 @@ def test_upstream_rejects_non_ancestor(prefix, connection_test):
         """
 
         def make(self, key):
+            # class-form lookup (Diagram.__getitem__ class branch)
             try:
                 self.upstream[Unrelated]
             except DataJointError as exc:
-                captured_errors.append(exc)
+                captured_errors.append(("class", exc))
+            # string-form lookup (the separate FreeTable/string branch)
+            try:
+                self.upstream[Unrelated.full_table_name]
+            except DataJointError as exc:
+                captured_errors.append(("string", exc))
             # Insert anyway so populate doesn't fail
             self.insert1({**key, "ok": 1})
 
     Bad.populate()
-    assert len(captured_errors) == 1
-    assert "not in this trace" in str(captured_errors[0]).lower()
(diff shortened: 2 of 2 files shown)


ttngu207:
The logic here is sound. `_upstream` is instance-level (not class-level like `_allow_insert`), set immediately before `make()` and cleared in a `finally` that also covers the exception branch, fresh per key and per `populate()` call, and genuinely tripartite- and multiprocess-safe (constructed once, all three phases read the same object; each worker unpickles its own instance). No leakage across keys or populates. `self & dict(key)` is the right, non-mutating seed.

One gap worth closing: the two guarantees this design most needs to prove are effectively **untested**.

- `test_upstream_cleared_after_make` probes `Derived().upstream` on a *brand-new* instance, but `populate()` runs `make()` on a throwaway `cls()` (via `TableMeta.__getattribute__`). The probe instance's `_upstream` was never set — it's the class default `None` — so this test would pass even if the `finally: self._upstream = None` line were deleted.
- No test lets `make()` itself raise, so the exception-path cleanup (the whole reason the reset is in `finally` rather than after `make()`) has zero coverage.

Suggest a test that captures the populate instance (`inst = Derived(); inst.populate(); assert inst.upstream raises`) and one that forces an exception through `make()` then asserts the context is clean. Everything else looks good.

ttngu207 (dismissed):

MilagrosMarin (dismissed):
Verified `7e4130f` against the stacked parent (`origin/feat/1423-diagram-trace`). Isolated diff: +51 in `src/datajoint/autopopulate.py`, +169 in `tests/integration/test_autopopulate.py`. Concur with @ttngu207 — logic sound; test coverage under-proves the design.

## Verified ✅

- **Instance-level `_upstream` with class-level `None` default.** Set inside `_populate_one` (`autopopulate.py:658`) right before `make()`, cleared in the `finally` block (`autopopulate.py:719`). Instance scope (not `self.__class__`) — no leakage across workers or keys.
- **Non-mutating seed.** `Diagram.trace(self & dict(key))` — `dict(key)` copies the mapping, so the caller's `key` is not touched. Small but correct.
- **Property gives an actionable error message outside `make()`** including the escape-hatch pattern (`dj.Diagram.trace(self & key)`).
- **Lazy at this layer.** Only the graph copy happens in `_populate_one`; SQL fetch fires when the user accesses `self.upstream[T].fetch(...)`. `make()` bodies that never touch `self.upstream` pay nothing.

## Concerns (concurring with @ttngu207 + additions)

1. **`test_upstream_cleared_after_make` is toothless.** It probes `Derived().upstream` on a *fresh* instance whose `_upstream` was never set — falls through to the class default (`None`) and would pass even if the `finally: self._upstream = None` line were deleted. To actually exercise the cleanup, capture the populate instance: `inst = Derived(); inst.populate(); with pytest.raises(...): inst.upstream`.

2. **Exception-path cleanup has zero coverage.** The whole reason `_upstream = None` lives in `finally` (rather than after `make()`) is to survive an exception. No test forces `make()` to raise. Add one that raises mid-body and asserts `self.upstream` is unset on the populate instance.

3. **Tripartite test doesn't verify tripartite.** `test_upstream_seen_across_tripartite_make` reads `self.upstream[Source]` only in `make_fetch`; `make_compute` and `make_insert` never touch it. The docstring claims "all three phases see the same self.upstream value" — that's not what the test verifies. An `id(self._upstream)` assertion (or `is` check) across the three phases would prove the "constructed once, shared across phases" property.

4. **Non-ancestor rejection tests only the class-based lookup.** `test_upstream_rejects_non_ancestor` uses `self.upstream[Unrelated]` (class). The string-based `FreeTable` path (`self.upstream["schema.Unrelated"]`) is a different branch in `Diagram.__getitem__` — the message overlap between `diagram.py:609` and `diagram.py:619` on the parent branch means the current substring assertion doesn't distinguish which path raised. A second string-based assertion would round out the coverage.

5. **#1471 coverage gaps carry through here.** `self.upstream` is a thin wrapper over `Diagram.trace`, and the four gaps flagged during #1471 review (multi-hop diamond OR-convergence, cross-schema ancestors, upward Part-of-Part chains, isolated non-primary FK) remain uncovered at this layer too. Reasonable to defer to a follow-up.

## Bottom line

Implementation is correct as far as the code shows. Items #1 and #2 would be trivial additions and would meaningfully strengthen the two guarantees the design most needs to prove. #3–#5 are nice-to-have. Approving with the same non-blocking stance as @ttngu207.

MilagrosMarin (dismissed):

MilagrosMarin (approved):

dimitri-yatsenko:
Pushed `8f9ad04` strengthening the test coverage you both flagged — should be a quick re-confirm (code unchanged; tests only). Mapping to your points:

1. **Toothless cleanup test** → `test_upstream_cleared_after_make` now captures the actual populate instance inside `make()` and asserts *its* `_upstream` is cleared afterward. It would now fail if the `finally: self._upstream = None` line were removed (the old fresh-instance probe passed regardless).
2. **Exception-path cleanup (zero coverage)** → new `test_upstream_cleared_after_make_raises`: forces `make()` to raise and asserts the populate instance's upstream is still cleared.
3. **Tripartite not actually verified** → `test_upstream_seen_across_tripartite_make` now records `id(self._upstream)` in each phase and asserts all three phases share one object per key — proving the "constructed once, shared across phases" claim rather than just the result.
4. **Class-only rejection** → `test_upstream_rejects_non_ancestor` now exercises both the class-form and string-form `Diagram.__getitem__` branches.

Validated locally on MySQL and PostgreSQL (full `test_autopopulate.py`: 18 passed, 2 skipped).

Deferred to follow-up (noted in the commit): upward Part-of-Part chains and an isolated non-aliased secondary-FK trace. The diamond-OR and cross-schema gaps from the #1471 review were closed at the trace layer in #1471 (now merged), which `self.upstream` inherits.

Also FYI: this PR was retargeted to `master` (was stacked on the now-merged `feat/1423-diagram-trace`) and updated — the diff is just the `self.upstream` delta.

ttngu207 (approved):

MilagrosMarin (approved):
re-approving.

Issue #1478, opened 2026-07-02
Trustworthy garbage collection: complete reference discovery (#1469) + race-safe deletion (#1445)


dimitri-yatsenko:
Umbrella tracking the invariant: **garbage collection must never delete an object that is still referenced.** Today it can, via two *independent* failure modes that look similar but have different root causes:

1. **Incomplete discovery** — #1469: `dj.gc.scan` hardcodes the built-in codec names (`hash`/`blob`/`attach`, `object`/`npy`), so files referenced by a **custom codec** are never enumerated. The store scan then reports live files as orphans and `collect()` deletes them. (Row delete also fails to remove custom-codec files.)
2. **Stale discovery / TOCTOU race** — #1445: the scan is correct, but an insert during the scan→delete window has its file deleted out from under it.

They are **complementary layers of one goal, not duplicates** — and must be fixed in order:

- [x] **#1469 — complete reference discovery (correctness, FIRST).** Codec-owned `referenced_paths` hook; `scan` and delete-cleanup become codec-driven instead of name-hardcoded. Closes active data loss for real pipelines (aeon_mecha).
- [ ] **#1445 — race-safe deletion (hardening, SECOND).** Two-phase quarantine → grace → purge with re-check-before-delete; `_trash/` prefix for state.

### Why the sequence matters
#1445's `purge()` re-check reruns the scan. On a codec-blind scan (pre-#1469) it would still classify a live custom-codec file as an orphan and purge it after the grace window — so the grace window can't save you from a scan that never sees the reference. **Complete discovery (#1469) is the precondition for safe deletion (#1445).**

### Documentation (datajoint-docs) — to land with the implementation

**Add**
- **`reference/specs/garbage-collection.md`** (new normative spec — none exists today). The orphan-determination model, the `referenced_paths` codec contract, the two-phase quarantine/grace/purge state machine, config keys (`gc.grace_seconds`), re-check/concurrency semantics, backend atomic-move requirements, and `restore`. (#1445 explicitly asks for a written spec.)

**Update**
- **`how-to/garbage-collection.md`** ("Clean Up Object Storage") — document the two-phase workflow (`quarantine` / `purge` / `restore`, `grace_seconds`); note custom-codec external files are now handled; revise the "single-pass, best-effort" admonition added in #189 once two-phase lands.
- **`reference/specs/codec-api.md`** — document `referenced_paths` as part of the Codec contract (required for any codec that owns external artifacts).
[...]

Issue #1445, opened 2026-04-30
Two-phase, transaction-safe garbage collection (quarantine -> grace -> purge)


kushalbakshi:
## Context

Follow-up to #1442 (data-loss bug in `dj.gc.scan` / `collect`, fixed in #1444). After that fix, `scan` correctly identifies referenced files; the present issue is about the broader concern that GC remains **not transaction-safe** even with the type-mismatch resolved.

There is a TOCTOU window between when `scan_*_references` enumerates references and when `collect` deletes orphans. A concurrent transaction that inserts a row referencing previously-orphaned content during that window will have its file deleted out from under it. This was raised by @dimitri-yatsenko during review of the #1442 root-cause analysis:

> In general, garbage collection is not 100% safe even after fixing since it's not transaction-safe. We need to implement a process of garbage retrieval for a bit and then a second step for complete removal.

## Proposed direction: two-phase API

Replace the single-call `collect()` with a quarantine → grace window → purge state machine. The single-call form can stay as a deprecated convenience, but production usage should be the two-phase form.

### Phase 1: `quarantine()`

- Identifies orphan candidates the same way `scan` does (now correct after #1442).
- Moves (or marks) candidates — does **not** delete.
- Idempotent: re-running picks up only newly-orphaned objects since last run.

Two options for state persistence — both worth a design pass:

- **Storage prefix (`_trash/`).** Move objects into a `_trash/` prefix under the same store. Atomic per object. Restore is a move-back. Storage layout encodes the quarantine state directly; no separate metadata table to keep in sync.
- **Database table (`_gc_quarantine`).** Record candidates in a project-level (cross-schema) table: `(path, store, quarantined_at, source_schema, source_table, source_attribute)`. Storage layout unchanged; restore is a leave-in-place + delete the row.

### Phase 2: `purge()`

- Operates only on candidates whose `quarantined_at` is older than `grace_seconds` (configurable, sensible default >= 24h, configurable via `dj.config["gc.grace_seconds"]` or similar).
- **Re-checks reference status before delete** (cheap, metadata-only after #1442). If a row inserted during the grace window now references the candidate, **unquarantine** it instead of deleting.
- Deletes only candidates that pass the re-check.

### Phase 0 / convenience: `restore(path_or_filter)`

- Explicit unquarantine for operator use: pulled-too-soon recovery, debugging, manual reclaim.

## Open design questions

[...]

PR #1479, opened 2026-07-02
fix(#1469): codec-driven GC reference discovery + file-level orphan matching

### LINKED ISSUE #1469: Garbage collection deletes external-store files belonging to existing rows (custom codecs)
### What we're seeing

A custom codec (a `SchemaCodec` subclass) was built to store `xarray` datasets as NetCDF in an external `protocol: file` store, one `.nc` file per row. Two problems show up around deletion and garbage collection:

1. **Deleting a row does not remove its external file.** The `.nc` file is left orphaned on disk.
2. **Running `dj.gc.collect()` to clean up those orphans also removes files that are still referenced by existing rows.** `dj.gc.scan()` reports files belonging to live rows as orphaned, so `collect()` deletes them along with the genuinely orphaned ones, leaving existing rows pointing at files that no longer exist.

The net effect is that the built-in `dj.gc` tooling offers no safe cleanup path for a custom codec's external store. Skipping GC leaves orphaned files behind on every delete, and running GC deletes files that are still in use.

### Sequence that triggers it

- Insert rows into a table with a custom `<codec@store>` column (each insert writes one file to the store).
- Delete a subset of the rows.
- Run `dj.gc.collect(schema, store_name=..., dry_run=False)` to reclaim the orphaned files.
- The files for the rows that were **not** deleted are removed too.

### Environment

- datajoint-python 2.2.4 (also reproduces on 2.2.2)
- Custom `SchemaCodec` subclass backed by a `protocol: file` store
- Context: custom xarray/NetCDF codec in SainsburyWellcomeCentre/aeon_mecha#587

### LINKED ISSUE #1486: Store keys hash_prefix/schema_prefix are accepted but not honored by writers or GC
Found during review of #1479 (GC). The per-store config keys `hash_prefix`, `schema_prefix`, and `filepath_prefix` are **accepted and validated** by settings (`Config.get_store_spec` → `_validate_prefix_separation`; also listed in `storage_adapter.py` COMMON_SPEC_KEYS) — but only `<filepath@>` consumes them, and only as **reserved-namespace guards** (`builtin_codecs/filepath.py`: filepath may not write under `hash_prefix`/`schema_prefix`; must stay under `filepath_prefix`).

Neither the hash writer (`hash_registry.build_hash_path` — fixed `_hash/{schema}/...` layout, now via the `HASH_STORAGE_PREFIX` constant) nor the schema-addressed writer (`storage.build_object_path` — `{schema}/{table}/{pk}/...`) nor GC honors them as **relocation** knobs.

## Why this needs a decision
- A user who sets `stores.x.hash_prefix = "content"` expecting relocation gets a *validated, silent no-op* — misleading config surface.
- Worse, the trap is loaded: if a future change wires `hash_prefix` into the writer **without** teaching GC, hash objects land outside `_hash/`, get enumerated by the *schema-addressed* walk, match no schema reference, and are **deleted as orphans** — data loss. (#1479 added a comment on `HASH_STORAGE_PREFIX` flagging exactly this.)

## Options
1. **Declare the layout fixed** (recommended): document `hash_prefix`/`schema_prefix` as what they actually are today — reserved-namespace declarations for `<filepath@>` cohabitation validation, not relocation settings — and rename or clearly re-document them (e.g. in the object-store-configuration spec).
2. **Wire them through**: honor the prefixes in `build_hash_path`/`build_object_path` AND in `gc.list_stored_hashes`/`list_schema_paths` in the same change, with migration guidance for existing stores. Larger and of unclear demand.

Either way, the current state — validate-and-ignore — should not survive. Related: the `<filepath@>`-cohabitation GC hazard noted on #1479 (unattributable files) and the fail-safe planned with #1445/#1478.

### LINKED ISSUE #1487: Store path settings (hash_prefix/schema_prefix) were not honored by writers or GC — fixed in 2.3.1
## Summary

The per-store path settings `hash_prefix` and `schema_prefix` (documented in [Configure Object Storage](https://docs.datajoint.com/how-to/configure-storage/) as controlling where each storage section lives) **were not incorporated by the storage writers or garbage collection** in DataJoint 2.x through 2.3.0:

- **Hash-addressed writes** (`<blob@>`, `<attach@>`, `<hash@>`) hardcoded the `_hash/{schema}/…` layout — a configured `hash_prefix` was accepted and validated but silently ignored.
- **Schema-addressed writes** (`<object@>`, `<npy@>`, plugin `SchemaCodec`s) ignored `schema_prefix` entirely and wrote to **root-level** `{schema}/{table}/…` paths instead of the documented `_schema/{schema}/{table}/…` section.
- **Garbage collection** likewise hardcoded `_hash/` and enumerated root-level paths.
- The only consumer of the settings was `<filepath@>` insert-time validation, which reserved the *declared* values — so overriding `hash_prefix` even un-reserved the real `_hash/` namespace.

## Fixed in 2.3.1 (PR #1479)

The settings are now honored end to end from a single source (`get_store_spec`, defaults `_hash` / `_schema` applied to every spec including plugin protocols; no component-level fallbacks): writers place objects under the configured sections, GC scans the configured hash section and never enters the configured `filepath_prefix` namespace, and `<filepath@>` validation reserves the same values the writer uses.

## Impact — upgrading may require checking paths

1. **Default configuration (most pipelines).** Hash-addressed layout is unchanged (`_hash/…`). Schema-addressed objects: **new writes land under `_schema/{schema}/…`**, while objects written by ≤2.3.0 remain at root-level `{schema}/…`. Both remain fully readable (every row's metadata records its complete path) and GC lists **both** layouts, so mixed stores are handled. But anything outside DataJoint that assumed the root-level layout — sync jobs, backup filters, bucket policies, manual browsing — should be checked and, where possible, switched to reading paths from row metadata (`ObjectRef.path`) rather than constructing them.
2. **Stores configured with custom `hash_prefix`/`schema_prefix` before 2.3.1.** Those settings did nothing at write time, so existing objects live at the *defaults* while the config declares otherwise. After upgrading, **new writes will move to the configured prefixes**, splitting the store's layout, and GC scans only the *currently configured* hash section — objects under the old section are not reclamation candidates while the custom setting is active, and hash deduplication will not match content across sections. Recommended: verify what your store actually contains before upgrading, and either drop the custom prefixes (keep the defaults that match your existing data) or accept the split knowingly.
3. **Do not change prefixes on a store that already holds data** — now documented with a warning ([datajoint/datajoint-docs#196](https://github.com/datajoint/datajoint-docs/pull/196)).

## Related

- #1486 — the design-level record of the validate-and-ignore state (resolved the same way; closes with PR #1479).
- datajoint/datajoint-docs#196 — documentation updated to the settings-driven layout, per-schema hash paths, and GC semantics.

dimitri-yatsenko:
Fixes #1469 — garbage collection deleting external-store files that belong to existing rows. Implements the plan from #1469 / #1478.

## Two root causes, both fixed

**1. Custom-codec blindness (the reported bug).** `gc.scan` recognized schema-addressed columns by hardcoded codec name (`object`/`npy`), so a custom `SchemaCodec` subclass (the reporter's NetCDF codec) was never scanned — its live files were reported orphaned and `collect()` deleted them. Recognition is now **by type** (`isinstance(codec, SchemaCodec)`), and reference extraction moves to a codec-owned hook:

```python
class Codec:
    def referenced_paths(self, stored) -> list[tuple[str, str | None]]:
        # default recognizes standard {path, store} metadata; override for custom shapes
```

Custom `SchemaCodec` subclasses inherit correct behavior for free; fully custom codecs can override. `scan` calls `attr.codec.referenced_paths(value)` per column instead of switching on name.

**2. Path-format mismatch (latent — also hit built-in `<object@>`/`<npy@>`).** A row's stored metadata references an object **file** (`{schema}/{table}/{pk}/{field}_{token}`), but `list_schema_paths` enumerated the enclosing **directory** — so the referenced and stored path sets never matched and *every* live schema-addressed object was flagged orphaned (existing tests only asserted `referenced >= 1`, never `orphaned == 0`, so it was hidden). `list_schema_paths` now enumerates **files** (matching referenced paths, with per-token granularity), and `delete_schema_path` removes the single orphaned file and prunes empty parent dirs.

> ⚠️ Reviewers: cause (2) above means pre-fix `collect()` could delete live `<object@>`/`<npy@>` files, not just custom-codec files. This PR corrects that for all schema-addressed storage.

## Not changed
Delete-then-GC semantics are unchanged — external files still survive row delete by design; GC reclaims. No delete-time eager cleanup (unsafe for hash dedup).

## Tests
- Recognition of a custom `SchemaCodec` subclass (`_uses_schema_storage`).
- End-to-end guard: after deleting one row and running `collect()`, the surviving row's file remains and only the deleted row's file is reclaimed — asserted by exact path (robust to the shared test store). Both new tests fail on pre-fix code.
- Full `test_gc.py` (32) + object/npy/hash/codec/adapter suites (237) pass locally on MySQL & PostgreSQL.

Discovery layer of the trustworthy-GC work (#1478); #1445 (two-phase quarantine/grace/purge) builds on the now-trustworthy scan.

Fixes #1486 (prefix keys re-documented as namespace declarations; GC honors `filepath_prefix` as a protected subtree — commit 068b365e).
Fixes #1487 (path settings not incorporated by writers/GC — see that issue for upgrade impact; ships in 2.3.1).


dimitri-yatsenko:
**Amendment pushed (`6e89a7d`) — reviewers please note before reading the diff.**

A deep re-analysis of this PR found a critical flaw in its original form: the file-level orphan matching broke **directory-valued objects** (Zarr-style `<object@>`). The row's metadata references the directory *prefix*; `list_schema_paths` enumerates the chunk *files* — so exact set difference flagged **every chunk of a live store as an orphan** (empirically confirmed: a live 3-chunk store showed `schema_paths_orphaned: 3`), and `collect(dry_run=False)` would have deleted live data. Folder objects also write a `{path}.manifest.json` sidecar the listing skipped, so orphaned folders would have leaked manifests forever.

The amendment makes orphan detection **coverage-based** (`_is_covered`): a stored file is referenced when it *is* a referenced path, lies *under* a referenced path, or is the `.manifest.json` sidecar of one. Manifests are now listed so orphaned ones are reclaimed. Two regression tests added (live-folder-never-flagged fails on the pre-amendment code; orphaned-folder fully reclaimed incl. manifest). GC suite 34/34 locally.

For context: the **pre-PR code also mishandled directory objects**, differently — the PK-level directory (holding the manifest) was orphan-flagged, so old `collect()` would `rm -r` the live object wholesale. Neither version was safe for folders; the PR is now correct for both single-file and directory objects.

One related, **pre-existing** hazard deliberately not addressed here (tracked for the #1478/#1445 follow-up): `<filepath@>`-managed files sharing a store root with schema-addressed columns are enumerated by the store walk but never referenced by any scanner, so `collect()` can delete user-managed filepath files — true before this PR (at directory granularity) and after it (at file granularity). The designed remedy is the fail-safe from the #1469 plan ("never delete a file GC cannot attribute to a participating codec's path-space"), which belongs with the #1445 two-phase work rather than this discovery fix.

ttngu207 (approved):

Issue #1481, opened 2026-07-04
Diagram.trace: descend through master-part boundaries (Merge-table upstream tracing)


dimitri-yatsenko:
## Summary

`Diagram.trace()` (and therefore `self.upstream`) walks **ancestor FK edges only**. It does not descend from an ancestor Master into that Master's Parts. In the merge-table shape used by Spyglass —

```
Parent → Master.Part → Master → Child
```

— `trace(Child & key)` reaches `Master`, but `trace[Master.Part]` and `trace[Parent]` raise `DataJointError`. The true upstream source (`Parent`) is unreachable through the merge point.

This matches the shipped spec (the Upstream Trace spec, `reference/specs/trace.md` §2 "Allowed table set"; formerly `provenance.md`, renamed in 2.3: an ancestor's Part is included only when the Part itself lies on an FK path to the seed) and is now pinned by `test_trace_stops_at_master_no_part_down_collection`. However, the design comment on discussion #1232 described a symmetric down-collection ("when a Master is reached, add its Parts to the trace, and continue upward from the Parts' FK parents") that was **never implemented** — a correction has been posted there. This issue tracks actually building that capability.

## Proposed behavior (opt-in or default — design question)

When the upward walk reaches a Master, additionally:
1. Restrict the Master's Parts downward from the Master's restriction (the existing forward rules).
2. Continue **upward** from the Parts' other FK parents (the upward rules), OR-merging into the trace.
3. Multi-pass until stable (mirror of the downstream `part_integrity="cascade"` mechanics — this is its upward analog, cf. `_propagate_part_to_master`).

## Design questions

- Default-on (a Part is semantically an extension of its Master, so "what contributed" arguably includes contributions through Parts) vs. opt-in flag (`trace(expr, through_parts=True)`) to preserve current semantics.
- OR-merge and termination through alias nodes; interaction with the (part, master)-pair dedup introduced for the downstream analog.
- Effect on `self.upstream`'s allowed-table set under `strict_provenance` — the merge pattern is exactly where downstream `make()`s need to read through the merge point.

## Motivation

- Spyglass Merge tables (discussion #1232, @CBroz1): versioned pipelines route provenance through Master/Part merge points; upstream tracing that stops at the merge master cannot answer "which Parent produced this result."
- Without it, strict-provenance mode makes such reads *impossible* rather than merely unergonomic (the true parent is not in the allowed set).

PR #1484, opened 2026-07-04
fix(cascade): restrict master from every restricted Part; validate part_integrity

dimitri-yatsenko:
Second code fix from the 2.3 post-release audit (companion to #1480).

## 1. `part_integrity="cascade"` under-restriction — CRITICAL (empirically confirmed, both backends)

The upward part→master walk was deduplicated **per master** (`visited_masters`, first Part wins). When two Parts of the same master were restricted via different FK paths from the seed, the second Part never contributed its master rows:

- Topology: `Ext ← Master.PartA`, `Ext ← Master.PartB`; PartA row (ext 1) under master 1, PartB row (ext 1) under master 2.
- `Diagram.cascade(Ext & {"ext_id": 1}, part_integrity="cascade")` restricted **only master 1** — the delete then removed master 2's PartB row while master 2 survived, silently violating the compositional integrity `"cascade"` promises. (The default `"enforce"` catches this topology via its post-check; only `"cascade"` failed silently.)

**Fix:** dedup by **(part, master) pair** — every restricted Part fires the walk once. Termination is preserved (finite pair set, global across passes); sibling re-fires append duplicate master restrictions, harmless under OR semantics. Distinct from the documented single-shortest-path limitation.

## 2. Eager `part_integrity` validation

- `Diagram.cascade()` accepted any string — a typo (`"casade"`) silently degraded to enforce/ignore behavior. Now raises `ValueError` (mirrors `Table.delete`).
- `Table.drop()` only compared `== "enforce"` — `drop(part_integrity="cascade")` silently behaved as `"ignore"` although cascade is delete-only. Now raises `ValueError`.

## 3. Merge-shape trace semantics pinned

New test `test_trace_stops_at_master_no_part_down_collection`: trace walks ancestor edges only and does **not** descend from an ancestor Master into its Parts (`Parent → Master.P → Master → Child`: trace reaches Master; `Master.P`/`Parent` raise). Matches provenance.md §2; corrects the unimplemented down-collection described in discussion #1232 (correction posted there); the capability is tracked as #1481. The pin exists so a future "fix" toward that promise can't silently change semantics.

## Validation
- New regression fails on the old dedup (`master count == 1`), passes on the fix (`== 2`, and end-to-end delete removes both compositional units, control master untouched).
- cascade+trace suites 46/46 on MySQL **and** PostgreSQL; wider regression (cascading_delete, dependencies, full unit suite) 333 passed.

Docs: the audit-batch PR (datajoint/datajoint-docs#195) adds the cascade Limitations section and does **not** document the once-per-master behavior — because this PR fixes it.

### DIFF
--- a/src/datajoint/diagram.py
+++ b/src/datajoint/diagram.py
@@ -340,6 +340,10 @@ class Diagram(nx.DiGraph):  # noqa: C901
         >>> # Inspect the cascade subgraph
         >>> dj.Diagram.cascade(Session & 'subject_id=1')
         """
+        # Validate eagerly (mirrors Table.delete) — a typo like "casade" would
+        # otherwise silently degrade to enforce/ignore behavior.
+        if part_integrity not in ("enforce", "ignore", "cascade"):
+            raise ValueError(f"part_integrity must be 'enforce', 'ignore', or 'cascade', got {part_integrity!r}")
         conn = table_expr.connection
         conn.dependencies.load_all_downstream()
         node = table_expr.full_table_name
@@ -698,7 +702,14 @@ class Diagram(nx.DiGraph):  # noqa: C901
         # Only propagate through descendants of start_node
         allowed_nodes = {start_node} | set(nx.descendants(self, start_node))
         propagated_edges = set()
-        visited_masters = set()
+        # Dedup upward part→master walks by (part, master) PAIR, not by master
+        # alone: distinct Parts of the same master can be restricted via
+        # different FK paths from the seed, and each must contribute its own
+        # master rows. Keying by master alone under-restricted the master when
+        # a second Part was reached later (silent integrity violation in
+        # part_integrity="cascade" mode). The pair set is finite and global
+        # across passes, so the multi-pass loop still terminates.
+        visited_part_master_pairs = set()
 
         restrictions = self._cascade_restrictions if mode == "cascade" else self._restrict_conditions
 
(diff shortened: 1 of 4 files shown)


MilagrosMarin (approved):
Thanks @dimitri-yatsenko — real integrity gap caught. The `(part, master)` pair-keyed dedup is minimal and precisely-targeted, and the repro in `test_cascade_two_parts_same_master_both_restrict` (Ext ← {PartA, PartB} sharing one Ext row across two different masters) hits exactly the topology the old `visited_masters` set silently dropped. Under OR semantics the extra walks can only add, never subtract, so no risk of over-restricting cases the old behavior got right.

The eager validation on `Diagram.cascade` correctly mirrors `Table.delete`, and rejecting `"cascade"` on `Table.drop` is the right call — better a loud error than the silent "ignore" degradation.

The trace-stops-at-master pin is well-reasoned: it locks the pure-ancestor-walk semantics against a future well-intentioned "fix" toward the down-collection that discussion #1232 described but was never implemented.

One minor: the `_propagate_restrictions` docstring doesn't mention the pair-keyed invariant — future maintainers will see the inline comment on `visited_part_master_pairs`, but a one-liner at the method-docstring level ("each restricted Part fires its own upward walk once") would help someone landing here without the commit context.

Approving.

MilagrosMarin (approved):

PR #1485, opened 2026-07-06
test(cascade): pin part-integrity paths (rollback, empty-materialization, U3 arm)

dimitri-yatsenko:
Adds integration coverage for the cascade part-integrity paths hardened in #1484, so a regression in either direction becomes visible.

## `test_cascade_integrity.py`
- **Post-check + rollback:** a cascade that reaches a Part without its Master raises `"before its master"` and rolls back completely (previously only the `Part.delete` pre-guard had a test).
- **Empty-materialization sentinel:** cascade mode with zero matching part rows previews the master at 0 and deletes cleanly.
- **U3 upward arm:** a Part with a *secondary* `-> master` FK — `proj(*attr_map.keys())` correctly restricts only the right master; a bare `proj()` would have restricted both (pins the exact #1468 code path).

New file only — no conflicts. Tests green locally on MySQL + PostgreSQL. Companion to the cascade fix in #1484.

## Scope note
This PR originally also pinned the `strict_provenance` read/write-gate limits. That surface is being retired in #1489 (which deletes `provenance.py`), so those tests and the now-moot `provenance.py` docstring edit were dropped to keep this PR coherent with the guardrail removal. What remains is the cascade-integrity coverage, which is independent of `strict_provenance`.


MilagrosMarin (approved):
Thanks @dimitri-yatsenko — three clean pins.

The **enforce post-check + rollback** test finally covers the data-driven path (cascade reaches Part but not Master → raise + rollback); previously only the direct `Part.delete` pre-guard had coverage.

The **empty-materialization sentinel** uses `counts.get(Master.full_table_name, 1) == 0` — the `1` default cleanly distinguishes "Master present with zero matches" (the intended `[False]` sentinel from #1468) from "Master not in counts at all". Both would pass a naïve `> 0` check, so the specificity here matters.

The **U3 arm with secondary master FK** is the isolated non-primary-FK direct-hit test I flagged as a follow-up during #1471. The topology forces U3 exactly, and asserting *only* master 1 is restricted genuinely distinguishes the fix — a bare `proj()` would fall to a cross-join and restrict both masters.

Dropping the `strict_provenance` limits to align with the guardrail removal in #1489 was the right call.

Approving.

MilagrosMarin (approved):

PR #1489, opened 2026-07-13
Retire strict_provenance runtime guardrail (#1488)

### LINKED ISSUE #1488: Retire the opt-in strict_provenance runtime guardrail; enforce the make() contract by deploy-time code inspection
## Summary

Remove `dj.config["strict_provenance"]` (introduced in 2.3.0, #1425) in an upcoming 2.3.x release, along with its documentation. Keep `Diagram.trace()` (#1423) and `self.upstream` (#1424). Enforcement of the `make()` contract moves to static/agentic inspection of `make()` bodies at deployment (platform CI), backed by governed, commit-pinned execution. The open-source documentation will document the **reproducible `make()` contract** itself — the rules — rather than a runtime enforcement flag.

## What shipped in 2.3.0

`strict_provenance` is an opt-in, off-by-default, best-effort development guardrail. It wires read/write checks into the Python client via a per-call `contextvars` enforcement context, confining reads to `self.upstream[...]` and writes to `self` and Parts. It shipped alongside `Diagram.trace()` and `self.upstream` as the "provenance trinity."

## Why remove it

The shipped specification already concedes the core point: comprehensive enforcement of the `make()` contract is fundamentally a **code-inspection problem, not a runtime-interception one**. Against the flag's modest dev-only benefit, the costs are real:

1. **Best-effort by construction.** The spec's own limits section enumerates the bypasses — existence/count idioms (`len`, `bool`, `in`), restriction-by-table, and any access outside the Python client. It also cannot see code paths a given run does not execute. A guardrail that misses these invites reliance it cannot support.
2. **Wrong instrument, wrong time.** The `make()` contract is a static property of the method body (reads confined to the upstream trace cone; writes to self/Parts; one entity per call in isolation; no result-affecting side inputs). Static analysis sees all paths, catches the documented bypass idioms, and reports once in CI before deploy — not per-invocation in the populate hot path.
3. **Complexity and maintenance burden.** Per-call `contextvars` context, client-path interception, and a documented-limits appendix are a standing support cost for a check that cannot be made sound in-process.
4. **Redundant with the authoritative enforcer.** The platform inspects `make()` source at deploy and controls execution (commit-pinned, mediated connections). That is where a sound guarantee lives.

## Why removal (not deprecation) is safe now

- The flag is **off by default** and has **no schema, data, or query-result impact** — removal changes no pipeline's behavior for anyone who has not opted in.
- It was introduced in 2.3.0 and has **no known adoption**: no platform customer has enabled it. Removing it now, before it gains use, is cheaper for everyone than carrying a deprecation cycle for a feature nobody depends on.

## What we keep (and stabilize)

`Diagram.trace()` and `self.upstream` — the provenance query API and the idiomatic read pattern. "Read from `self.upstream`, write to `self` and Parts" remains the documented convention, and it is precisely what makes a `make()` method statically verifiable.

## Proposed changes

- Remove the `strict_provenance` config key and the runtime interception machinery (enforcement context, client-path checks).
- Remove all mentions from the documentation (2.3 release notes, provenance specification §3); reframe the "trinity" as the `trace` + `self.upstream` pair.
- Replace the removed enforcement docs with documentation of the **reproducible `make()` contract**: populate-only insertion; one entity per call, computed in isolation; reads confined to the upstream trace cone; writes only to self and Parts; no other result-affecting inputs (populate kwargs orchestrate, never parameterize).
- Optionally, publish the static `make()` verifier as a standalone dev tool / pre-commit hook, preserving the "catch accidental drift locally" purpose with strictly better coverage than the runtime gate.

## Alternatives considered

- **Harden the runtime gate** — rejected: closing the documented bypasses approaches whole-program analysis inside a running process.
- **Deprecate first, remove at 2.4** — unnecessary ceremony given off-by-default, zero functional impact, and no known adoption; removing before the feature gains use is the cheapest path for users and maintainers.

Refs: #1425, #1423, #1424, #1474, #1480, #1485. Docs affected: `about/whats-new-23.md`, `reference/specs/provenance.md`.

dimitri-yatsenko:
Closes #1488.

## Summary

Removes the opt-in `dj.config["strict_provenance"]` flag (introduced in 2.3.0, #1425) and its runtime interception machinery. Comprehensive enforcement of the `make()` contract is fundamentally a code-inspection problem, not a runtime-interception one — the shipped guardrail was best-effort by construction (documented bypasses via existence/count idioms, restriction-by-table, and any access outside the Python client) and could not be made sound in-process. Enforcement moves to deploy-time static/agentic inspection of `make()` bodies on the platform, backed by governed, commit-pinned execution. The open-source docs document the **make() reproducibility contract** itself (the rules) rather than a runtime flag.

## Kept (and stabilized)

- `Diagram.trace()` (#1423)
- `self.upstream` (#1424)

"Read from `self.upstream`, write to `self` and Parts" remains the documented convention — and it is exactly what makes a `make()` statically verifiable. These are cleanly separable from the removed enforcement code: `self._upstream = Diagram.trace(self & key)` still runs unconditionally.

## Changes

- **Delete `src/datajoint/provenance.py`** — the entire module (enforcement context, read/write gates, per-row key-consistency check) existed only to back the flag.
- **`settings.py`** — drop the `strict_provenance` config field and its `DJ_STRICT_PROVENANCE` env-var mapping.
- **`expression.py`** — remove the read gate from `QueryExpression.cursor`.
- **`table.py`** — remove the write gate and the per-row key-consistency check from the insert path.
- **`autopopulate.py`** — remove the per-`make()` strict context push/pop; keep the upstream-trace construction and the `upstream` property (docstring reworded to drop the flag reference and point at the AutoPopulate spec).
- **Delete `tests/integration/test_strict_provenance.py`.**

## Why removal, not deprecation

The flag is **off by default**, has **no schema/data/query-result impact**, and has **no known adoption** (no platform customer enabled it). Removing it before it gains use is cheaper than carrying a deprecation cycle for a feature nobody depends on.

## Docs

The `strict_provenance` documentation (2.3 release notes, provenance spec §3) lives in the separate `datajoint-docs` repo and is handled there — the reproducible `make()` contract is being documented in `reference/specs/autopopulate.md` (datajoint-docs #197), and the "trinity" reframed to the `trace` + `self.upstream` pair.

## Verification

- `datajoint` imports cleanly (2.3.0); `strict_provenance` key removed; `Diagram.trace` and `self.upstream` intact; `datajoint.provenance` no longer importable.
- No dangling references to the flag or the removed helpers anywhere in `src/`/`tests/`.
- `ruff`, `ruff-format`, and `codespell` pre-commit hooks pass. (Full integration test run requires a live database, not available in the authoring environment.)


MilagrosMarin (approved):
Thanks @dimitri-yatsenko — clean removal, 608 lines out and 8 in. Verified `provenance.py` and `test_strict_provenance.py` are fully deleted, no dangling references (`grep -rn` for the removed symbols returns nothing), and the pieces this PR keeps — `Diagram.trace()` at `diagram.py:391` and `AutoPopulate.upstream` at `autopopulate.py:95` — are intact and still wire `Diagram.trace(self & dict(key))` unconditionally in `_populate_one`. The gates come out of `expression.cursor` and `table.insert`/`_insert_rows` without collateral refactoring, the config field and env-var mapping both go, and the docstring/README/test-comment wording follows the reframing away from "provenance" as a runtime term.

The "removal, not deprecation" argument holds: off by default, no schema/data/query-result impact, no known adoption. This is the same reasoning that landed on the docs side in #183's reframing — the runtime guardrail was best-effort by construction and couldn't be made sound in-process.

One thing for the release notes (correcting my earlier phrasing): `dj.config["strict_provenance"] = True` will now raise a `KeyError("Setting 'strict_provenance' not found")` at assignment time, not a silent no-op. The path is `_ConfigProxy.__setitem__` → `Config.__setitem__` at `settings.py:959-966`, which checks `hasattr(self, key)` and raises `KeyError` when the field is absent. Cleaner error than I initially wrote, but the substance stands: 2.3.0 shipped the flag as public API, so the 2.3.1 release notes should still say "removed" explicitly — a matter of release-notes transparency, not adopter protection (the PR body's "no known adoption" is fair on that front).

Approving.

MilagrosMarin (approved):

PR #1490, opened 2026-07-13
Extend Python support to 3.14; test min/max in CI

dimitri-yatsenko:
## Motivation

An audit of supported Python versions against the [official EOL schedule](https://devguide.python.org/versions/) found two gaps:

| Version | EOL | Status before this PR |
|---|---|---|
| 3.9 | 2025-10 | Already excluded (`>=3.10`) ✅ |
| **3.10** (floor) | **2026-10** | Supported; EOL in ~3 months ⚠️ |
| 3.11 | 2027-10 | Supported |
| 3.12 | 2028-10 | Supported |
| 3.13 | 2029-10 | Supported; **the only version CI tested** |
| **3.14** | 2030-10 | **Released 2025-10 — was excluded** ⚠️ |

Two problems: (1) the ceiling lagged a full release behind — 3.14 has been out since Oct 2025 but `requires-python = ">=3.10,<3.14"` blocked it; and (2) although the package *claimed* 3.10–3.13, the pixi-based test job resolved a single interpreter (3.13), so the declared floor was never exercised.

## Changes

- **`pyproject.toml`**
  - `requires-python` and the pixi `python` dependency → `>=3.10,<3.15` (adds **3.14**).
  - Add per-version PyPI **classifiers** for 3.10–3.14 (previously only a generic `Programming Language :: Python`), so the supported range is discoverable on PyPI.
  - Add version-pinned pixi test environments **`test-py310`** and **`test-py314`** (each solves independently — a solve-group can't span Python versions).
- **`.github/workflows/test.yaml`** — run the containerized test job as a **matrix over `[test-py310, test-py314]`**, so both ends of the supported range are validated on every push/PR.

## Fixes surfaced by the 3.14 matrix

Running the full suite on 3.14 exposed two pre-existing issues (latent because CI only ever ran 3.13); both are fixed here:

- **`test(tls)`** — `test_insecure_connection` asserted `Ssl_cipher == ""`, but the `datajoint/mysql:8.0` test image negotiates TLS during connection setup even when the client requests none, so the cipher is not reliably empty. Now pins what DataJoint controls (`use_tls=False` sends no client-side SSL config; the connection is usable) rather than a server-negotiated outcome.
- **`fix(populate)`** — `test_multi_processing` hung on 3.14. Python 3.14 changed the multiprocessing default start method from `fork` to `forkserver`; parallel populate hands the live table and its open DB connection to workers by process inheritance (fork), which `forkserver` breaks by pickling the payload instead. Both `Pool` call sites are now pinned to a `fork` context.

## Notes / follow-ups

- **3.10 reaches EOL 2026-10.** Dropping it (min → 3.11, and retargeting ruff `target-version`/mypy `python_version`) is a small breaking bump left as a **follow-up** rather than folded in here.
- The single-version `unit-tests` job (fast feedback) resolves to the highest supported version (3.14); the full matrix covers min+max. (Reviewer suggestion: make this an explicit `py314` pin — deferred as a follow-up.)
- The `test` matrix job has no `timeout-minutes`, so the 3.14 hang ran ~1h before it was cancelled during review — worth adding a job timeout as a follow-up so future hangs fail fast.
- `pixi.lock` is re-solved in CI (`locked: false`), so no committed lock update is needed; the PR's own CI run validates that all dependencies resolve on both 3.10 and 3.14.

## Verification

- Both matrix jobs (`test-py310`, `test-py314`) pass, along with `lint` and `unit-tests`.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -27,6 +27,14 @@ on:
 jobs:
   test:
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        # Exercise both ends of the supported range (requires-python
+        # >=3.10,<3.15). The version-pinned pixi environments are defined in
+        # pyproject.toml under [tool.pixi.environments].
+        environment: [test-py310, test-py314]
+    name: test (${{ matrix.environment }})
     steps:
       - uses: actions/checkout@v4
 
@@ -35,9 +43,10 @@ jobs:
         with:
           cache: true
           locked: false
+          environments: ${{ matrix.environment }}
 
       - name: Run tests
-        run: pixi run -e test test-cov
+        run: pixi run -e ${{ matrix.environment }} test-cov
 
   # Unit tests run without containers (faster feedback)
   unit-tests:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,7 +19,7 @@ dependencies = [
   "packaging",
 ]
 
-requires-python = ">=3.10,<3.14"
+requires-python = ">=3.10,<3.15"
 authors = [
   {name = "Dimitri Yatsenko", email = "dimitri@datajoint.com"},
   {name = "Thinh Nguyen", email = "thinh@datajoint.com"},
@@ -57,6 +57,11 @@ keywords = [
 # https://pypi.org/classifiers/
 classifiers = [
   "Programming Language :: Python",
+  "Programming Language :: Python :: 3.10",
+  "Programming Language :: Python :: 3.11",
+  "Programming Language :: Python :: 3.12",
+  "Programming Language :: Python :: 3.13",
+  "Programming Language :: Python :: 3.14",
   "Development Status :: 5 - Production/Stable",
   "Intended Audience :: Science/Research",
   "Intended Audience :: Healthcare Industry",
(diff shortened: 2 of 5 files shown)


MilagrosMarin (dismissed):
Thanks @dimitri-yatsenko — nice audit catch. The EOL alignment and pixi feature-based envs are exactly right, and splitting 3.10's eventual drop as a follow-up (given its 2026-10 EOL) keeps this PR clean.

Two small observations, neither blocking:

**Middle versions are unexercised.** Testing 3.10 and 3.14 covers the endpoints where compat issues usually cluster, but 3.11–3.13 aren't touched. Reasonable trade-off for CI-time — just worth being explicit that a middle-version-specific regression would land with users first.

**The `unit-tests` job leans on pixi's solver picking the highest Python.** PR body says it "resolves to 3.14," which is pixi's default behavior today, but the reliance is implicit. If a future pixi version shifts its default strategy, `unit-tests` could quietly slide to a lower version. Adding an explicit `py314` pin to that env would make the intent enforced rather than emergent.

Approving.

MilagrosMarin (dismissed):

dimitri-yatsenko:
Re-requesting review — two commits landed **after** the earlier approval, both fixing CI failures the new py3.10/py3.14 matrix surfaced (no change to the pyproject/CI matrix that was already reviewed):

- **`test(tls)`** — `test_insecure_connection` asserted `Ssl_cipher == ""`, but the `datajoint/mysql:8.0` test image negotiates TLS during connection setup even when the client requests none, so the cipher was not reliably empty (passed on the py3.13 solve, failed on the fresh py3.10/py3.14 solves). Now pins what DataJoint controls: `use_tls=False` sends no client-side SSL config and the connection is usable.
- **`fix(populate)`** — `test_multi_processing` failed on py3.14 with `TypeError: cannot pickle 'itertools.count' object`. Python 3.14 changed the multiprocessing default start method from `fork` to `forkserver`, which pickles the table sent to workers; `Dependencies` held an unpicklable `itertools.count`. Added `__getstate__`/`__setstate__` to drop and rebuild it (workers reload dependencies anyway). This makes the pickle-based worker path — which the worker init docstring already assumes — actually work.

py3.10 is green; py3.14 is still running. Flagging @MilagrosMarin 's earlier suggestion (explicit `py314` pin on `unit-tests`) as an optional follow-up — happy to fold it in if you want it here.

MilagrosMarin (approved):

Issue #1497, opened 2026-07-16
DB compatibility: add MySQL 8.4 (LTS/RDS) to CI matrix; pin documented support to tested versions (scope out 9.x)


dimitri-yatsenko:
## Summary

Our database compatibility story has a gap between what we **claim**, what we **test**, and what we
actually **deploy to** (AWS RDS). Most importantly, **MySQL 8.4 — the current LTS and what AWS RDS runs
— is not tested in CI**, and our open-ended version wording implicitly promises MySQL 9.x, which is an
Innovation series (never an LTS, not an RDS target) that we should not be committing to.

## Current state

**Documented claims (minimums only, open-ended):**
- `reference/specs/database-backends.md` — "Supported Backends" table: MySQL min `8.0.13`, PostgreSQL
  min `15`, both "Production."
- `how-to/installation.md`: "MySQL 8.0.13+ or PostgreSQL 15+."
- `about/whats-new-21.md`: "PostgreSQL 15+ … alongside MySQL 8+."
- `connection.py` warns MariaDB is not officially supported in 2.x (best-effort).

**Actually tested in CI:** only MySQL `8.0` and PostgreSQL `15`, pinned via testcontainers
(`tests/conftest.py:117` `datajoint/mysql:8.0`, `:146` `postgres:15`). The CI matrix
(`.github/workflows/test.yaml`) varies **Python only** (3.10 / 3.14) — the DB version is **not** a
matrix axis. So MySQL 8.0 + PG 15 are the only combinations we prove.

## Gaps

1. **MySQL 8.4 (LTS / AWS RDS) is untested.** We test 8.0, not 8.4 — yet 8.4 is the version we most
   need to guarantee for RDS deployments.
2. **Open-ended `8.0.13+` implicitly promises 9.x.** MySQL 9.x are Innovation releases — never an LTS,
   not offered as a standard RDS engine. We have fielded 9.6 bug reports for a version we do not deploy
   to; the docs should bound this, not invite it.
3. **PostgreSQL claimed `15+` but only 15 is tested** (16/17 unverified).

## Proposed work

1. **Make the DB version a real CI matrix axis.** Add **MySQL 8.4** alongside 8.0 (parameterize the
   testcontainers image, e.g. `MYSQL_VER` in `tests/conftest.py` + a matrix in `test.yaml`). Optionally
   add PostgreSQL 16/17 if we want to claim beyond 15.
2. **Tighten the documented support policy to match tested versions.** State explicit tested versions
[...]

PR #1498, opened 2026-07-16
ci: test MySQL 8.0 and 8.4 (LTS) in the matrix; parameterize MySQL image

dimitri-yatsenko:
Closes the gap where CI only tested MySQL 8.0 while AWS RDS runs the **8.4 LTS** line (see #1497).

**Change:**
- `tests/conftest.py`: MySQL testcontainer image is now read from `DJ_TEST_MYSQL_IMAGE` (default `datajoint/mysql:8.0` — no local behavior change).
- `.github/workflows/test.yaml`: adds a `mysql-image` matrix axis — `datajoint/mysql:8.0` and official `mysql:8.4` — crossed with the existing Python axis (3.10/3.14).

**Why official `mysql:8.4`:** there is no `datajoint/mysql:8.4` image published; the only thing the datajoint image adds is SSL certs for the TLS tests, and those already `skipif` outside external containers — so they skip in the testcontainers matrix on both legs regardless. Behavioral (SQL/type/cascade) coverage on 8.4 — the RDS-critical part — runs fully.

**Follow-ups (not blocking):**
- Publish a `datajoint/mysql:8.4` image if we want TLS parity on the 8.4 leg.
- Once this lands, flip the "Tested in CI" cell in datajoint-docs#209 from "8.0 (8.4 in progress)" to "8.0, 8.4".
- Matrix is a full 2×2; can be trimmed with `exclude` if CI minutes are a concern.

Refs #1497. **Scoped out:** MySQL 9.x (Innovation, non-LTS, not RDS).

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -30,11 +30,15 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        # Exercise both ends of the supported range (requires-python
-        # >=3.10,<3.15). The version-pinned pixi environments are defined in
-        # pyproject.toml under [tool.pixi.environments].
+        # Python: exercise both ends of the supported range (requires-python
+        # >=3.10,<3.15); version-pinned pixi environments live in pyproject.toml
+        # under [tool.pixi.environments].
         environment: [test-py310, test-py314]
-    name: test (${{ matrix.environment }})
+        # MySQL: exercise both LTS lines we support / that AWS RDS runs. 8.0 uses
+        # the datajoint image (SSL configured); 8.4 uses the official image (TLS
+        # tests already skip outside external containers). See #1497.
+        mysql-image: ["datajoint/mysql:8.0", "mysql:8.4"]
+    name: test (${{ matrix.environment }}, mysql=${{ matrix.mysql-image }})
     steps:
       - uses: actions/checkout@v4
 
@@ -46,6 +50,8 @@ jobs:
           environments: ${{ matrix.environment }}
 
       - name: Run tests
+        env:
+          DJ_TEST_MYSQL_IMAGE: ${{ matrix.mysql-image }}
         run: pixi run -e ${{ matrix.environment }} test-cov
 
   # Unit tests run without containers (faster feedback)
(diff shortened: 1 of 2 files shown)


MilagrosMarin (approved):
Verified the key claims and this looks solid — nice tight scope.

**Checked:**
- ✅ `requires_ssl` in `tests/integration/test_tls.py` skips whenever `DJ_USE_EXTERNAL_CONTAINERS` is unset, so the TLS tests skip on *both* matrix legs — the 8.4 leg loses nothing relative to 8.0. Confirmed in CI: both legs run identical test sets (951 passed, 8 skipped each).
- ✅ Docker Hub `datajoint/mysql` has no 8.4 tag, so the official image is the right call for now.
- ✅ The job rename is safe — no required status checks or rulesets reference the old `test (test-py310)` names.
- ✅ `DJ_TEST_MYSQL_IMAGE` naming is consistent with the existing `DJ_TEST_DB_PREFIX` pattern.

**Two small non-blocking suggestions:**
1. The conftest module docstring (lines 11–16) documents the other env vars (`DJ_USE_EXTERNAL_CONTAINERS`, `DJ_HOST`, …) — worth adding a line for `DJ_TEST_MYSQL_IMAGE` for discoverability.
2. If `DJ_TEST_MYSQL_IMAGE` ever stops propagating (e.g., a pixi env-handling change), the 8.4 leg would silently test 8.0 and stay green. The fixture logs host:port but not the image, and pytest doesn't echo `logger.info` without `log_cli`. Including the image (or a `SELECT VERSION()` result) in the startup log — or printing it in the CI step — would make each leg auditable from the logs.

**One observation for the #1497 follow-up (not this PR):** MySQL 8.4 disables `mysql_native_password` by default, so clients authenticate via `caching_sha2_password`. PyMySQL needs the `cryptography` package for that auth over non-TLS connections, and it's present in our CI env only as a transitive test dependency — it's not in datajoint's runtime deps. A user on a bare `pip install datajoint` connecting to an 8.4/RDS server without TLS could hit a "cryptography is required" auth error. Worth folding into the support-policy docs work: either add `pymysql[rsa]` to runtime deps or document that TLS (or `cryptography`) is expected for 8.4.

mweitzel (approved):
Awesome. Thank you for such a quick response to this.

PR #1499, opened 2026-07-16
fix(#1493): build self.upstream lazily; stop reloading deps per populate key

### LINKED ISSUE #1493: populate: self.upstream (Diagram.trace) built unconditionally per make() — ~53 ms + 59 information_schema queries/key
## Summary

`AutoPopulate._populate1` unconditionally constructs `self._upstream = Diagram.trace(self & key)`
for **every** populate key (`src/datajoint/autopopulate.py:673`), even when `make()` never accesses
`self.upstream`. Profiling shows this costs **~53 ms and 59 `information_schema` round-trips per key**
— a large `populate()` throughput regression introduced with the 2.3 `self.upstream` feature.

## Profiling

MySQL 8.0, N=1000 tight loop, synthetic chain `Session(Manual) → A → B → C` (all Computed, so `C`
has 3 ancestors), `gc` disabled, `conn.query` wrapped to count SQL round-trips.

| measurement | mean | median |
|---|---|---|
| `Diagram.trace(C & key)` | **52.9 ms**, **59 SQL round-trips** | 52.2 ms |
| — `load_all_upstream()` only | 5.0 ms, 3 round-trips | 4.9 ms |
| — propagation walk + headings | 43.7 ms, ~56 round-trips | 40.5 ms |
| no-op `C & key` | **13.4 µs** | 13.8 µs |

**Ratio: trace ≈ 3,900× the no-op.** A 1,000-key populate pays **~53 s + ~59,000 metadata queries**
of overhead *before any `make()` work*.

Root causes:
- `load_all_upstream()` (`dependencies.py:262`) calls `load(force=True)`, which `clear()`s and
  re-queries `information_schema` for all PKs + FKs and rebuilds the whole graph on **every** call —
  the `_loaded` flag is ignored (`dependencies.py:143`).
- The dominant ~44 ms is `_propagate_restrictions_upstream` building a fresh `FreeTable` per ancestor
  per pass and `.restrict()`ing it, each materializing a table heading (a metadata query).

The comment at `autopopulate.py:668` ("Lazy — only the graph copy runs here; the expensive SQL fetch
fires when the user accesses `self.upstream[T]`") is **inaccurate** — the construction itself is
neither lazy nor in-memory.

## Fix (priority order)

1. **Build the trace lazily** — don't construct in `_populate1`; build on first `self.upstream`
   access (cached per current `key`). Keys whose `make()` never touch `upstream` then pay ~0. This
   alone removes the regression.
2. **Cache the dependency load** — `load_all_upstream()` should honor `_loaded` / cache within a
   populate run instead of `load(force=True)` every call.
3. **Reuse `FreeTable` headings** across propagation passes so ancestor headings aren't re-fetched
   per key.

## Triage

**Slated for v2.3.1** (per review triage). `self.upstream` is a headline 2.3 feature, so shipping it
with this per-key tax is a regression we want fixed before tagging 2.3.1. Benchmark script and
methodology available on request.

## References

- Surfaced in the 2.3.1 `dj.Diagram` review (finding F1).
- Code: `autopopulate.py:673` (`_populate1`), `diagram.py:395` (`trace`),
  `dependencies.py:143` (`load`) / `:262` (`load_all_upstream`).

dimitri-yatsenko:
Fixes the per-`make()` cost profiled in #1493 (~53 ms and ~59 `information_schema` round-trips per populate key, even when make() never reads `self.upstream`).

**1. Lazy `self.upstream`** (`autopopulate.py`): `_populate1` records the make() key (`_upstream_key`) instead of eagerly building `Diagram.trace()`. The trace is built on first `self.upstream` access and memoized for the call, so make() calls that never touch upstream pay nothing. Both `_upstream` and `_upstream_key` are cleared in `finally`.

**2. Idempotent dependency loading** (`dependencies.py`): track `_loaded_schemas`; `load_all_upstream`/`load_all_downstream` skip the full `information_schema` rebuild when the graph already contains every needed schema — so repeated `Diagram.trace()` calls within one `populate()` don't reload the dependency tree per key.

**Tests:** three `self.upstream` tests pinned the old eager internal (`self._upstream` non-None throughout make); updated to the lazy contract. One now asserts `self._upstream` stays `None` when make() never reads upstream (proving laziness); the tripartite test captures `id(self.upstream)` (the property) to verify one shared object across phases. Locally: 6/6 upstream + 76 trace/cascade/delete/erd tests pass on MySQL and PostgreSQL.

Closes #1493.

### DIFF
--- a/src/datajoint/autopopulate.py
+++ b/src/datajoint/autopopulate.py
@@ -98,7 +98,8 @@ class AutoPopulate:
     _key_source = None
     _allow_insert = False
     _jobs = None
-    _upstream = None  # set per-make() by _populate_one; see `upstream` property below
+    _upstream = None  # memoized upstream trace for the current make(); built lazily on first use
+    _upstream_key = None  # current make() key; not None iff executing inside make()
 
     @property
     def upstream(self):
@@ -132,12 +133,19 @@ class AutoPopulate:
                 traces = self.upstream[ExtractTraces].to_arrays("trace")
                 self.insert1({**key, "summary": compute(traces, date)})
         """
-        if self._upstream is None:
+        if self._upstream_key is None:
             raise DataJointError(
                 "self.upstream is only available inside make(). "
                 "Outside make(), construct a trace explicitly: "
                 "dj.Diagram.trace(self & key)."
             )
+        if self._upstream is None:
+            # Build the trace lazily on first access and memoize it for this
+            # make() call, so make() calls that never read self.upstream pay
+            # nothing. See #1493.
+            from .diagram import Diagram
+
+            self._upstream = Diagram.trace(self & self._upstream_key)
         return self._upstream
 
     class _JobsDescriptor:
@@ -665,12 +673,11 @@ class AutoPopulate:
         logger.jobs(f"Making {key} -> {self.full_table_name}")
         self.__class__._allow_insert = True
 
-        # Pre-construct the upstream view for this make() call. Lazy — only
-        # `dj.Diagram.trace(self & key)` runs here (graph copy); the
-        # expensive SQL fetch fires when the user accesses self.upstream[T].
-        from .diagram import Diagram
-
-        self._upstream = Diagram.trace(self & dict(key))
+        # Record the key for a lazily-constructed upstream view. The trace is
+        # built on first `self.upstream` access (see the `upstream` property),
+        # so make() calls that never read upstream pay nothing. See #1493.
+        self._upstream_key = dict(key)
+        self._upstream = None
 
         try:
             if not is_generator:
@@ -733,6 +740,7 @@ class AutoPopulate:
             # access raises a clear error rather than silently using a
             # stale trace from the previous make() call.
             self._upstream = None
+            self._upstream_key = None
 
     def progress(self, *restrictions: Any, display: bool = False) -> tuple[int, int]:
         """
--- a/src/datajoint/dependencies.py
+++ b/src/datajoint/dependencies.py
@@ -132,11 +132,13 @@ class Dependencies(nx.DiGraph):
         self._conn = connection
         self._node_alias_count = itertools.count()
         self._loaded = False
+        self._loaded_schemas = set()  # schema names currently represented in the graph
         super().__init__(self)
 
     def clear(self) -> None:
         """Clear the graph and reset loaded state."""
         self._loaded = False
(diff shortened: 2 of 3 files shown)


MilagrosMarin (approved):
Thanks @dimitri-yatsenko — clean two-part fix for #1493. Verified end-to-end:

**1. Lazy `self.upstream` (`autopopulate.py`).** `_populate1` now sets `_upstream_key = dict(key)` and `_upstream = None` before invoking `make()`; the property builds `Diagram.trace(self & self._upstream_key)` on first access and memoizes. The "outside make()" gate is correctly switched from `_upstream is None` to `_upstream_key is None` — otherwise the very first access inside make() would raise. `finally` clears both. Tripartite path (fetch/compute/insert with the generator restart at line 720) shares the same `_upstream` slot across phases, so `id(self.upstream)` stays stable — the updated test assertion holds.

**2. Idempotent dep loading (`dependencies.py`).** `_loaded_schemas` tracks what's in the graph; `load_all_upstream`/`load_all_downstream` short-circuit when `_loaded and known_schemas ⊆ _loaded_schemas`. Correctness cases hold: first call (unloaded) → reload; repeated call with same schemas → skip; discovery finds a new schema → superset check fails → reload; `clear()` resets both fields. `load()` continues to unconditionally repopulate `_loaded_schemas`, so direct `load(force=True, ...)` callers still get a fresh graph.

**Test updates are on-point.** `test_upstream_cleared_after_make` now asserts `self._upstream is None` *inside* the make() body that doesn't read upstream — that's the specific regression guard for the laziness claim, and it would fail against the old eager code. The tripartite test's switch from `id(self._upstream)` to `id(self.upstream)` correctly captures the property (which triggers the memoized build) rather than the raw internal attribute.

Two small observations, neither blocking:

- **No test for the dep-loading idempotence itself.** A query-count assertion (`information_schema` round-trips = N on first `Diagram.trace()`, 0 on repeats) would concretely pin the perf fix — otherwise a future refactor could break the short-circuit silently. The existing tests cover correctness; this would cover the perf guarantee the PR's motivation rests on.
- **Private-attribute contract shifted.** `self._upstream is None` no longer discriminates "outside make()" — `self._upstream_key` does. Since these are underscore-prefixed with no stability promise, this is fine; the test file itself updated to the new discriminator. Just worth being aware if any external code (plugin, extension, or user script) introspected `_upstream` for that signal.

Approving.

PR #1500, opened 2026-07-16
docs(diagram): document restrict is downstream-only (no part->master walk-up)

dimitri-yatsenko:
From the 2.3.1 `dj.Diagram` review (master-part focus). Makes explicit — in the `restrict` docstring and at the upward-walk guard in `_propagate_restrictions` — that `restrict` is **downstream-only** and never walks up from a Part to its Master.

Rationale: the part→master upward walk is exclusive to `cascade` delete, which **mutates** and must be compositionally atomic regardless of seed. `restrict` reads a slice seeded at roots and flows down, so a Master is always included (with all its Parts) before its Parts — no Part FK dangles, and no atomicity obligation. Upstream/ancestor closure is `trace`'s job.

**No behavior change** — the `mode=="cascade"` guard already enforced this. The added comment prevents it from being accidentally enabled for restrict in a future change. Docstring + comment only.

### DIFF
--- a/src/datajoint/diagram.py
+++ b/src/datajoint/diagram.py
@@ -663,6 +663,13 @@ class Diagram(nx.DiGraph):  # noqa: C901
 
         Cannot be called on a Diagram produced by ``Diagram.cascade()``.
 
+        Propagation is strictly downstream. Unlike ``cascade``'s
+        ``part_integrity="cascade"``, ``restrict`` never walks upward from a
+        Part to its Master: an export is seeded at root entities and flows
+        down, so a Master (with all of its Parts) is always included before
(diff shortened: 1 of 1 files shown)


dimitri-yatsenko:
Closing without merging. This PR documents `restrict` as strictly downstream ('never walks up from Part to Master'), but the v2.4 rethink of `restrict` semantics (#1501) will actually introduce a Part→Master relationship — a Part condition lifting existentially to its Master so whole compositional units are kept. Documenting the 'no part→master' contract now would be premature and contradict that direction. Folding the restrict-contract documentation into #1501 so it is written once, against the final v2.4 semantics.

PR #1505, opened 2026-07-17
docs(codecs): fix stale storage-path and hash-algorithm docstrings

MilagrosMarin:
Three built-in codec class docstrings still describe pre-2.3.1 storage layouts, and `HashCodec` additionally names the wrong hash algorithm. Every one of these docstrings had its "Deletion:" line touched in v2.3.1 (`dj.gc.collect()` → `dj.gc.GarbageCollector`), but the surrounding path claims were left stale — so a reader sees a locally-updated block and assumes it's current.

## Changes

**`src/datajoint/builtin_codecs/hash.py`** (module docstring + class summary + storage description)
- "SHA256 deduplication" → "MD5-based deduplication" / "content-addressed deduplication". The actual algorithm is `md5(data)` + `base32encode`, then lowercased and stripped of `=` padding (`hash_registry.py:65-67`).
- Layout `_hash/{hash[:2]}/{hash[2:4]}/{hash}` → `{hash_prefix}/{schema}/{hash}` (with `hash_prefix` defaulting to `_hash`; stores with subfolding get additional segments). No `[:2]/[:4]` character split exists in the code.

**`src/datajoint/builtin_codecs/npy.py`** (class summary + Storage Details)
- Path `{schema}/{table}/{pk}/{attribute}.npy` → `{schema_prefix}/{schema}/{table}/{pk}/{attribute}_{token}.npy`. The `schema_prefix` section and `_{token}` suffix are what `build_object_path` now writes (`storage.py:191-278`) after #1479's prefix-configuration work.

**`src/datajoint/builtin_codecs/object.py`** (class summary + Storage Structure)
- Same drift as NpyCodec, as a directory rather than a file.

## Sibling docstrings already correct

`src/datajoint/hash_registry.py:8-14` and `src/datajoint/gc.py:8-24` module docstrings already document the current layout — these three codec class docstrings are the outliers.

Docs-only.

### DIFF
--- a/src/datajoint/builtin_codecs/hash.py
+++ b/src/datajoint/builtin_codecs/hash.py
@@ -1,5 +1,5 @@
 """
-Hash-addressed storage codec with SHA256 deduplication.
+Hash-addressed storage codec with MD5-based deduplication.
 """
 
 from __future__ import annotations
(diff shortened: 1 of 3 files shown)


dimitri-yatsenko (approved):
LGTM. Verified against `hash_registry.py`: the codec docstring was wrong (claimed SHA256; the implementation is Base32-encoded MD5, 26 chars) and the subfold path was stale — the corrected `{hash_prefix}/{schema}/{hash}` (+ subfolding) matches the module docstring verbatim, and the `npy`/`object` `{schema_prefix}`/`{token}` paths match `_build_path`. Good catch.

PR #1504, opened 2026-07-17
chore(release-flow): stale doc pointers, RELEASE_MEMO cleanup, and dead badge step

MilagrosMarin:
Addresses findings 2, 3, and 4 from #1503, plus follow-up items from Dimitri's review. Leaves finding 1 (version.py version-at-tag timing) open — that one needs a maintainer decision on approach (sed before tag / retag / hatch-vcs).

## Changes

**`src/datajoint/version.py`** — header comment listed three workflow files that don't exist (`label_prs.yaml`, `release.yaml`, `post_release.yaml`); updated to name the actual workflows (`draft_release.yaml`, `post_draft_release_published.yaml`).

**`RELEASE_MEMO.md`**
- **Branch Structure table**: `master` row was listed as `2.1.x` (stale by two minor versions); bumped to `2.3.x`. Removed the `maint/2.0` row and the "For 2.0.x bugfixes:" block — no longer supporting 2.0.x. Leaves the actual `maint/2.0` git branch alone; this is doc-only.
- **PR-labels table**: category names were "Keep a Changelog"-style (`BREAKING CHANGES`, `Added`, `Fixed`, `documentation | (usually excluded)`); updated to match `release_drafter.yaml` verbatim (`💥 Breaking Changes`, `⚡️ Enhancements`, `🐛 Bug Fixes`, `📝 Documentation`). Added the missing `feature` → `🚀 Features` row.
- **Maintainers**: added `@MilagrosMarin`.

**`.github/workflows/post_draft_release_published.yaml`** — removed the "Update README.md badge" step. Its two `sed` calls targeted `<a id="commit-since-release-link">` and `` — both elements were removed from `README.md` in [`ff04914c`](https://github.com/datajoint/datajoint-python/commit/ff04914c) (2026-01-04, "docs: Update README with Relational Workflow Model and OAS"). Every release since has run empty `sed`s guarded by a `git-diff --cached --quiet || git commit` no-op check.

If you'd prefer to keep the README-badge step and restore the badge markup to README instead, happy to swap the approach.

Closes findings 2, 3, 4 in #1503.

Issue #1503, opened 2026-07-17
Release-flow cleanup: version.py timing at tag, dead workflow references, stale RELEASE_MEMO, dead README-badge sed


MilagrosMarin:
Four small release-automation inconsistencies surfaced during a v2.3.1 audit. Grouping them since they're all in the same "release-flow debris" bucket. Individually small; several are one-line fixes.

## 1. `src/datajoint/version.py:4` — reports the previous release at every tag

At the v2.3.1 tag itself (commit `29ce695c`), `src/datajoint/version.py` contains `__version__ = "2.3.0"`. Same pattern at v2.3.0 (contains `2.2.4`), v2.2.4, etc. — the bump PR always follows the tag.

**Root cause:** `.github/workflows/post_draft_release_published.yaml` sed's `version.py` and opens a bump PR **after** the release is published — so the tag is anchored on the pre-bump tree. PyPI is fine (Hatch builds the wheel from the sed'd file before `twine upload` in the same workflow run), but:

- `pip install git+https://github.com/datajoint/datajoint-python.git@v2.3.1` → wheel with `__version__ == "2.3.0"`.
- `pip show datajoint` on that install reports `2.3.0`.
- Version-gated code (`if datajoint.__version__ >= "2.3.1":`) silently takes the wrong branch.

**Options:**
- (a) sed `version.py` **before** creating the tag (biggest change; requires reordering the release workflow).
- (b) Move the tag to the post-release bump commit (still requires workflow edit).
- (c) Switch to dynamic version resolution (`hatch-vcs`) so tag == reported version by construction (cleanest, biggest packaging change).

## 2. `src/datajoint/version.py:1-2` — comment references workflow files that don't exist

```python
# version bump auto managed by Github Actions:
# label_prs.yaml(prep), release.yaml(bump), post_release.yaml(edit)
```

None of `label_prs.yaml`, `release.yaml`, or `post_release.yaml` exist under `.github/workflows/`. The actual files that manage the release are `draft_release.yaml` and `post_draft_release_published.yaml`. Anyone tracing the release mechanism from this comment lands on missing files.

**Fix:** update the comment to name the actual workflows:

```python
# version bump auto managed by Github Actions:
# draft_release.yaml (draft), post_draft_release_published.yaml (bump + PR)
```

## 3. `RELEASE_MEMO.md:5-8` — branch structure table lists master as `2.1.x`

```
| Branch     | Purpose             | Version |
|------------|---------------------|---------|
[...]

PR #1507, opened 2026-07-17
fix(release): derive version from git tag via hatch-vcs (#1503)

dimitri-yatsenko:
Resolves **finding 1** of #1503. (Findings 2–4 already landed in #1504.)

## Problem

The version bump PR lands *after* the release tag, so `src/datajoint/version.py` at each tag still reports the **previous** release (at the `v2.3.1` tag it reads `2.3.0`, and so on). PyPI wheels are unaffected — they're built from the sed'd tree in the same workflow run — but:

- `pip install git+https://github.com/datajoint/datajoint-python.git@v2.3.1` → wheel reporting `2.3.0`
- `pip show datajoint` on that install reports `2.3.0`
- version-gated code (`if datajoint.__version__ >= "2.3.1"`) silently takes the wrong branch

## Fix — make the git tag the single source of truth (hatch-vcs)

- **`pyproject.toml`**: `[tool.hatch.version] source = "vcs"` (+ `hatch-vcs` build dep) and a `[tool.hatch.build.hooks.vcs]` hook that writes the resolved version to the generated, git-ignored `src/datajoint/_version.py`.
- **`src/datajoint/version.py`**: re-exports the generated version, falling back to installed package metadata (`importlib.metadata`) for source-tree runs — so `datajoint.__version__` keeps working everywhere.
- **`.github/workflows/post_draft_release_published.yaml`**: drops the `version.py` sed/commit step and the follow-up bump PR entirely; checks out with `fetch-depth: 0` so the build resolves the tag. Publishing the `vX.Y.Z` tag is now the only version action.
- **`.gitignore`**: ignore the build-generated `_version.py`.
- **`RELEASE_MEMO.md`**: document the tag-driven flow (no manual bump, no bump PR).

## Verification

Built and installed locally against a throwaway `v9.9.9` tag:

- `hatchling version` at the clean tag → `9.9.9`; `python -m build` → `datajoint-9.9.9-py3-none-any.whl` and `datajoint-9.9.9.tar.gz` (the sdist→wheel path `pip install git+…@vX` uses).
- Installed wheel → `datajoint.__version__ == 9.9.9` and `importlib.metadata.version("datajoint") == 9.9.9`.
- A dirty/untagged tree correctly yields a `.devN+g<sha>` version; `_version.py` is git-ignored and never tracked.

## Notes

- No `version.py` value to bump by hand anymore; the tag sets the version for `pip install`, `pip show`, and `datajoint.__version__` alike.
- Opened as a draft — this touches the release path, so worth a careful look before it's live. The workflow change can only be fully exercised on the next real release.
- **Behavior change — bare source checkout without install.** Running from an uninstalled tree (`PYTHONPATH=./src python -c "import datajoint; print(datajoint.__version__)"`) now reports `0.0.0+unknown`, because both the build-generated `_version.py` and installed package metadata are absent. This differs from the previous hardcoded `__version__`. Anyone developing seriously will `pip install -e .` (which runs the hatch-vcs hook and generates `_version.py`), so this is minor — but flagged here so that seeing `0.0.0+unknown` after upgrading isn't mistaken for a regression; the fix is `pip install -e .`.

### DIFF
--- a/.github/workflows/post_draft_release_published.yaml
+++ b/.github/workflows/post_draft_release_published.yaml
@@ -12,9 +12,8 @@ run-name: Post ${{ github.event.release.name }}
 jobs:
   pypi-release:
     permissions:
-      # write permission is required to update version.py
+      # write permission is required to upload build artifacts as release assets
       contents: write
-      pull-requests: write
     # Use the oldest supported version to build, just in case there are issues
     # for our case, this doesn't matter that much, since the build is for 3.x
     strategy:
@@ -33,20 +32,10 @@ jobs:
         uses: actions/checkout@v4
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
-      # new release needs the updated version.py
-      - name: Update version.py
-        run: |
-          VERSION=$(echo "${{ github.event.release.name }}" | grep -oP '\d+\.\d+\.\d+')
-          sed -i "s/^__version__ = .*/__version__ = \"$VERSION\"/" src/datajoint/version.py
-          cat src/datajoint/version.py
-          # Commit the changes
-          BRANCH_NAME="update-version-$VERSION"
-          git switch -c $BRANCH_NAME
-          git config --global user.name "github-actions"
-          git config --global user.email "github-actions@github.com"
-          git add src/datajoint/version.py
-          git commit -m "Update version.py to $VERSION"
-          echo "BRANCH_NAME=$BRANCH_NAME" >> $GITHUB_ENV
+          # Full history + tags so hatch-vcs can derive the version from the
+          # release tag. The checked-out ref is the tag itself, so the build
+          # reports exactly X.Y.Z — no version.py bump or follow-up PR needed.
+          fetch-depth: 0
       - name: Set up Python ${{matrix.py_ver}}
         uses: actions/setup-python@v5
         with:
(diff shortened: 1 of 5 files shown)


MilagrosMarin (approved):
Thanks — closes finding 1 of #1503 exactly as I'd hoped (option c). All 12 CI checks green.

**Verified:**
- **hatch-vcs config** is standard: `source = "vcs"` + `version-file = "src/datajoint/_version.py"` build hook + `hatch-vcs` in build requires.
- **`version.py` fallback chain** is sound: build-generated `_version` → `importlib.metadata.version("datajoint")` → `"0.0.0+unknown"` sentinel. Covers wheel, sdist, editable install, and bare-checkout cases.
- **`.gitignore`** correctly excludes generated `src/datajoint/_version.py`.
- **Workflow** drops the sed + bump-PR steps cleanly, removes `pull-requests: write` permission (no longer needed), and adds `fetch-depth: 0` — the critical bit, since shallow clones would leave hatch-vcs without the tag.
- **`RELEASE_MEMO.md`** update accurately reflects the new tag-as-source-of-truth flow.
- **Author verification** against `v9.9.9` (wheel + sdist + installed metadata all agree; dirty tree → `.devN+g<sha>`) matches standard hatch-vcs behavior.

**Two non-blocking observations:**

### 1. First-real-release risk

The workflow only fires on release publish, so it can't be fully end-to-end-tested until the next real release. You note this in the PR body. Reasonable to accept given the local reproduction, but worth being ready to hotfix quickly if the workflow surfaces something unexpected on first fire.

### 2. Bare-checkout dev-mode change worth a release-notes line

Before this PR, `PYTHONPATH=./src python -c "import datajoint; print(datajoint.__version__)"` on a bare git clone (no `pip install`) returned the hardcoded `__version__`. After this PR, that path lands on `"0.0.0+unknown"` because both `_version.py` and installed package metadata are absent. Very minor — anyone doing serious development would `pip install -e .` (which runs the hatch-vcs hook and generates `_version.py`). But worth a release-note line so anyone hitting `0.0.0+unknown` after upgrade doesn't mistake it for a regression; the fix is `pip install -e .`.

### 3. Rebase before merge

Same as #1508 — branch predates #1506, auto-mergeable but worth rebasing so the diff view stays clean.

Approving on the substance — clean tag-driven migration.

PR #1508, opened 2026-07-17
refactor(graph): migrate Dependencies & Diagram to nx.MultiDiGraph (#1492)

### LINKED ISSUE #1492: Migrate dependency & Diagram graphs to nx.MultiDiGraph (retire alias-node workaround)
## Summary

Migrate `Dependencies` and `Diagram` (both currently `nx.DiGraph` subclasses) to
`nx.MultiDiGraph`, retiring the integer "alias node" workaround used to represent
multiple / renamed foreign keys between the same pair of tables.

## Background

A plain `nx.DiGraph` permits only one edge per ordered node pair, so it cannot
represent two foreign keys between the same two tables. To work around this,
`dj.Diagram` inserts an integer-named **alias node** that splits a renamed FK into
two half-edges (`parent → "N" → child`), giving each FK its own edge identity
(`dependencies.py:216–223`).

`dj.Diagram` predates NetworkX's multigraph support *in our design* (the class dates
to ~2016). NetworkX has long shipped `MultiDiGraph` (parallel edges keyed by
`(u, v, key)`), so the alias-node workaround is no longer necessary — it is a
historical choice, not a NetworkX limitation.

## Scope — what this removes

Migrating to `MultiDiGraph` deletes the alias machinery:

- ~14 `.isdigit()` special-case checks in `diagram.py`
- `_find_real_edge_props` (`diagram.py`)
- alias-follow branches in both `_propagate_restrictions` and `_propagate_restrictions_upstream`
- alias-node insertion at `dependencies.py:220–223`
- alias-collapse in `topo_sort` (`dependencies.py:62–73`)

Traversals we rely on (`ancestors`, `descendants`, `topological_sort`,
`shortest_path`, `node_boundary`) all work on multigraphs. The main change is that
edge-**data** access moves to `(u, v, key)` addressing throughout.

## Diagramming notation implications

Alias nodes currently double as the **visual marker** for renamed FKs. Removing them
means we need new notation:

- **Renamed-FK edge notation.** Design notation for a renamed foreign key as an
  *edge* attribute (e.g. a labeled / styled / dashed edge showing `child_col ← parent_col`)
  rather than an intermediate node.
- **Multigraph rendering support.** Ensure the render backends handle parallel edges
  between the same node pair:
  - pydot / pygraphviz → dot / graphviz (`make_dot`, `make_svg`, `to_pydot` at `diagram.py:1540`)
  - Mermaid (`make_mermaid` at `diagram.py:1642`)
  - matplotlib `draw`

  Graphviz/dot support parallel edges natively; Mermaid flowcharts need multiple edge
  statements between the same nodes — verify edge labels don't collide.

## Risk

Broad, cross-cutting, and test-heavy: every edge-data access site plus the full
cascade / trace / restrict suites must be re-verified under multigraph edge semantics.
Not a patch-release change — hence **v2.4**.

## References

- Surfaced during the 2.3.1 `dj.Diagram` review (alias-node design rationale).
- Code: `src/datajoint/dependencies.py`, `src/datajoint/diagram.py`.

dimitri-yatsenko:
Resolves #1492.

Migrates `Dependencies` and `Diagram` from `nx.DiGraph` to `nx.MultiDiGraph`, retiring the integer "alias node" workaround for renamed / multiple foreign keys between the same pair of tables. A renamed FK is now a first-class parallel edge — no synthetic intermediate node.

## Why

A plain `DiGraph` allows one edge per ordered node pair, so a renamed FK was split into `parent → "N" → child` through an integer alias node, and ~14 `.isdigit()` special-cases across `diagram.py` (plus 5 in `table.py`) existed only to see through those nodes. `MultiDiGraph` represents parallel edges natively, so the whole workaround is removed — it was a historical design choice, not a NetworkX limitation.

## Edge identity

Each edge is keyed by the FK's **child-side attribute tuple** (in declaration order), e.g. `("spouse1",)` vs `("spouse2",)`. This key is:
- **unique** among parallel edges between a pair — every FK to a given parent references the parent's primary key, so parallel FKs differ only in their referencing columns;
- **schema-derived** — no DB-internal constraint names;
- **stable & meaningful** — so graph merges are *idempotent by construction*: `Diagram + Diagram` re-adds a shared FK under the same key (overwrite, not duplicate). No defensive dedup is needed at the call sites.

## What changed

- `Dependencies`/`Diagram` subclass `nx.MultiDiGraph`; `load()` adds one keyed edge per FK; `topo_sort` drops its alias-collapse pass; `cascade()`/`trace()` rebuild via `MultiDiGraph`.
- Both propagation walkers merged their alias/direct branches into one loop over parallel edges; `_find_real_edge_props` → `_edge_props` (direct lookup).
- `dependencies.parents()`/`children()` return a parallel-safe `list[(node, props)]` (was a name-keyed dict that collapsed multiple FKs from the same parent); `Table.parents/children/descendants/ancestors/parts` drop the `.isdigit()` handling.
- **Rendering:** a renamed FK draws as an `#FF8800` **edge** with the column renames in a hover tooltip (dot/SVG) and a `linkStyle` recolor (mermaid), instead of the orange alias-node dot. The orange color is **layered on top of** the existing line semantics — solid=primary / dashed=secondary, thick=1:1 / thin=multi are all preserved. `_AliasNode` tier removed.
- `_apply_collapse` keeps parallel edges between two expanded tables (by key) while merging edges into a collapsed schema node to a single arrow.

## Verification

Full `tests/` suite under the pixi `test` env (MySQL 8.0 + PostgreSQL 15 via testcontainers):
- Green **baseline (95 passed)** on the graph/cascade/trace/FK suite before any change, reproduced after the migration — behavior-identical.
- Renamed-FK behavior green across all three style cases (solid-thick, solid-thin, dashed-thin — all orange): `test_aliased_fk`, `test_trace_renamed_fk`, `test_cascade_delete_with_renamed_attrs`, `test_cascade_part_of_part_renamed_fk`, `test_upward_u3_nonprimary_master_fk`; render smoke tests (`test_repr_svg`, `test_make_image`).
- **Full suite: 949 passed, 10 skipped, 0 failures** (both backends).

## Notes

- Milestone **v2.3.2**. Paired docs update: datajoint/datajoint-docs#215 (renamed-FK notation), gated on this shipping.

### DIFF
--- a/src/datajoint/dependencies.py
+++ b/src/datajoint/dependencies.py
@@ -8,7 +8,6 @@ proper ordering of operations like delete and drop.
 
 from __future__ import annotations
 
-import itertools
 import re
 from collections import defaultdict
 
@@ -41,13 +40,13 @@ def extract_master(part_table: str) -> str | None:
     return None
 
 
-def topo_sort(graph: nx.DiGraph) -> list[str]:
+def topo_sort(graph: nx.MultiDiGraph) -> list[str]:
     """
     Topological sort keeping part tables with their masters.
 
     Parameters
     ----------
-    graph : nx.DiGraph
+    graph : nx.MultiDiGraph
         Dependency graph.
 
     Returns
@@ -56,21 +55,7 @@ def topo_sort(graph: nx.DiGraph) -> list[str]:
         Table names in topological order with parts following masters.
     """
 
-    graph = nx.DiGraph(graph)  # make a copy
-
-    # collapse alias nodes
-    alias_nodes = [node for node in graph if node.isdigit()]
-    for node in alias_nodes:
-        try:
-            direct_edge = (
-                next(x for x in graph.in_edges(node))[0],
-                next(x for x in graph.out_edges(node))[1],
-            )
-        except StopIteration:
-            pass  # a disconnected alias node
-        else:
-            graph.add_edge(*direct_edge)
-    graph.remove_nodes_from(alias_nodes)
+    graph = nx.MultiDiGraph(graph)  # make a mutable copy
 
     # Add parts' dependencies to their masters' dependencies
     # to ensure correct topological ordering of the masters.
@@ -102,12 +87,16 @@ def topo_sort(graph: nx.DiGraph) -> list[str]:
     return sorted_nodes
 
 
-class Dependencies(nx.DiGraph):
+class Dependencies(nx.MultiDiGraph):
     """
     Graph of foreign key dependencies between loaded tables.
 
-    Extends NetworkX DiGraph to track foreign key relationships and
-    support operations like cascade delete and topological ordering.
+    Extends NetworkX MultiDiGraph to track foreign key relationships and
+    support operations like cascade delete and topological ordering. A
+    ``MultiDiGraph`` is used so that multiple foreign keys between the same
+    pair of tables (including renamed/aliased ones) are represented as
+    distinct parallel edges keyed by ``(parent, child, key)`` — no synthetic
+    intermediate "alias" nodes are needed.
 
     Parameters
     ----------
@@ -130,7 +119,6 @@ class Dependencies(nx.DiGraph):
 
     def __init__(self, connection=None) -> None:
         self._conn = connection
-        self._node_alias_count = itertools.count()
         self._loaded = False
         self._loaded_schemas = set()  # schema names currently represented in the graph
         super().__init__(self)
@@ -139,7 +127,6 @@ class Dependencies(nx.DiGraph):
         """Clear the graph and reset loaded state."""
         self._loaded = False
         self._loaded_schemas = set()
-        self._node_alias_count = itertools.count()  # reset alias IDs for consistency
         super().clear()
 
     def load(self, force: bool = True, schema_names: set[str] | None = None) -> None:
@@ -208,7 +195,14 @@ class Dependencies(nx.DiGraph):
             d["referenced_table"] = key["referenced_table"]
             d["attr_map"][key["column_name"]] = key["referenced_column_name"]
 
-        # add edges to the graph
+        # Add one edge per foreign key, keyed by the FK's child-side attribute
+        # names as a tuple (in declaration order). This key uniquely identifies
+        # a parallel edge between a pair of tables — every FK to a given parent
+        # references its primary key, so parallel FKs differ only in their
+        # referencing columns — and is derived purely from the schema (no
+        # DB-internal constraint names). A stable, meaningful key makes graph
+        # merges (e.g. ``Diagram + Diagram``) idempotent instead of duplicating
+        # shared edges, so no synthetic alias node is needed.
         for fk in fks.values():
             props = dict(
                 primary=set(fk["attr_map"]) <= set(pks[fk["referencing_table"]]),
@@ -216,14 +210,12 @@ class Dependencies(nx.DiGraph):
                 aliased=any(k != v for k, v in fk["attr_map"].items()),
                 multi=set(fk["attr_map"]) != set(pks[fk["referencing_table"]]),
             )
-            if not props["aliased"]:
-                self.add_edge(fk["referenced_table"], fk["referencing_table"], **props)
-            else:
-                # for aliased dependencies, add an extra node in the format '1', '2', etc
-                alias_node = "%d" % next(self._node_alias_count)
-                self.add_node(alias_node)
-                self.add_edge(fk["referenced_table"], alias_node, **props)
-                self.add_edge(alias_node, fk["referencing_table"], **props)
+            self.add_edge(
+                fk["referenced_table"],
+                fk["referencing_table"],
+                key=tuple(fk["attr_map"]),
+                **props,
+            )
 
         if not nx.is_directed_acyclic_graph(self):
             raise DataJointError("DataJoint can only work with acyclic dependencies")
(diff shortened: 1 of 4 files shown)


dimitri-yatsenko:
Paired docs PR: datajoint/datajoint-docs#215 — updates the diagram-notation docs (renamed FKs render as orange **edges** with a rename tooltip instead of alias-node dots) and adds a renamed-primary example showing the line-style semantics are retained.

The docs PR re-executed its notebooks against this branch, so it is gated on this shipping in **v2.3.2** and must not merge until the release.

MilagrosMarin (approved):
Thanks — clean, substantive refactor. All 12 CI checks green (lint + Py 3.10/3.14 × MySQL 8.0/8.4 test matrix + unit), and the PR body's "949 passed, 10 skipped, 0 failures on both backends" checks out with the CI signal here.

**Verified:**
- **Edge key `tuple(fk["attr_map"])`** is deterministic — `information_schema.KEY_COLUMN_USAGE` returns FK columns ordered by `ORDINAL_POSITION` on both MySQL and Postgres, so the tuple order is stable. Idempotency claim (`Diagram + Diagram` overwrites shared FKs) is correct given keyed-edge semantics.
- **`_AliasNode` fully retired** — deletion in `user_tables.py` symmetric with removed import in `diagram.py`; `_get_tier` now returns `None` for the integer-node case (which no longer arises).
- **`.isdigit()` special-cases cleanly removed** across `table.py` (parents/children/descendants/ancestors/parts) and `dependencies.py:topo_sort`.
- **Rendering** — `#FF8800` orange on renamed-FK edges (SVG via `set_color`, Mermaid via `linkStyle N stroke:#FF8800`), tooltips with column renames. Layered on top of existing line semantics (solid/dashed for primary/secondary, thick/thin for 1:1/multi) as the PR body claims.

**Three non-blocking observations:**

### 1. Breaking API change worth calling out in v2.3.2 release notes

`Dependencies.parents()` / `Dependencies.children()` return type changed from `dict[str, dict]` to `list[tuple[str, dict]]`. The change is *correct* (a name-keyed dict couldn't represent parallel edges), and in-tree call sites in `table.py` are updated cleanly. But downstream consumers walking `Dependencies` directly will break silently — `deps.parents(name)[parent_name]` → `TypeError: list indices must be integers`, `deps.parents(name).items()` → `AttributeError`. Elements packages, plugins, or user framework-level code are the exposure. Worth an explicit line in the release notes.

### 2. Consider adding a test for the parallel-edges capability itself

Existing tests listed in the PR body (`test_aliased_fk`, `test_trace_renamed_fk`, `test_cascade_delete_with_renamed_attrs`, `test_cascade_part_of_part_renamed_fk`, `test_upward_u3_nonprimary_master_fk`) all exercise the *single*-renamed-FK cases and pass — that's a solid sanity signal that behavior is unchanged. But **two parallel edges between the same pair of tables** (the novel capability `MultiDiGraph` enables and the alias-node workaround emulated) — e.g. `Relationship` with `-> Person.proj(person1='person_id')` and `-> Person.proj(person2='person_id')` — has no direct test. A targeted case that inserts both FKs and asserts (a) `parents(Relationship)` returns two entries for `Person`, (b) cascade/trace handle both edges, (c) rendering produces two distinct edges — would give the new parallel-edges path direct coverage rather than relying on implicit exercise.

### 3. Needs rebase before merge

Branch predates #1506 (merged 3 days later); the raw `git diff origin/master` shows my #1506 changes as "removals" purely because of that offset. Auto-mergeable (no file overlap with the `gc.py` changes on master), but worth rebasing so the diff view and history stay clean.

Approving on the substance — the refactor itself is well-executed.

dimitri-yatsenko:
Thanks @MilagrosMarin — responses to the three observations:

**1. Breaking API change — scoped to internal plumbing, not a user-facing break.** The return-type change is on `Dependencies.parents()`/`children()` (the graph object), whose only callers are `Table.parents()`/`Table.children()` (`table.py:379,411`). The public, documented `Table.parents()`/`children()` return shape is unchanged — the diff only drops the alias-node `.isdigit()` branch; it still yields the same `list[(name, props)]`. Elements and plugins use `Table.parents()` / `dj.Diagram`, not `conn.dependencies.parents()` directly (confirmed: no such usage in Elements, and the graph accessor isn't documented API). Note too that the old dict-keyed return couldn't represent parallel edges — it silently dropped one — so the new list is strictly more correct. So I'm **not** tagging this `breaking` or adding a release-notes line; at most it's an internal-accessor note for anyone hacking on `conn.dependencies`.

**2. Parallel-edges test — agreed, good gap to close.** The single-renamed-FK cases pass, but the two-parallel-edges capability (`-> Person.proj(person1=…)` + `-> Person.proj(person2=…)`) has no direct test. I'll add a targeted case (assert `parents()` returns both `Person` entries, cascade/trace handle both, two distinct rendered edges) as a fast follow-up rather than block this merge.

**3. Rebase — merging via rebase** so history stays linear/clean.

Merging now.

dimitri-yatsenko:
Follow-up for observation 2 (parallel-edges coverage) opened: #1514 — adds `test_parallel_edges_to_same_parent` using `LocalSynapse` (two renamed FKs to `Cell`), asserting two distinct parent edges across `Table.parents()`/`children()`, the dependency `MultiDiGraph`, and the rendered `Diagram`. Passes against MySQL 8.

PR #1509, opened 2026-07-18
chore: normalize stdlib copy imports to `import copy`

dimitri-yatsenko:
Standardizes how the stdlib `copy` module is imported across the codebase. Three modules used three conventions:

- `diagram.py` — `import copy as copy_module` → `import copy` (the alias was unnecessary; `copy` only appears there as a keyword argument, which can't collide with the module name)
- `settings.py` — `from copy import deepcopy` → `import copy`, call-sites qualified as `copy.deepcopy`
- `expression.py` — already `import copy` (uses `copy.copy`); unchanged

Now uniform: `import copy` with qualified `copy.deepcopy` / `copy.copy`, so one import covers both shallow and deep copies. Pure rename — no behavior change (identifiers only; ruff/isort/mypy clean, imports verified). Split out of #1508.

### DIFF
--- a/src/datajoint/diagram.py
+++ b/src/datajoint/diagram.py
@@ -15,7 +15,7 @@ pygraphviz. All other methods are always available.
 
 from __future__ import annotations
 
-import copy as copy_module
+import copy
 import functools
 import inspect
 import io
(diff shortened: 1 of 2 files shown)


MilagrosMarin (approved):
LGTM

PR #1511, opened 2026-07-20
docs: post-2.3.1 stale-text sweep (3 small drift fixes)

MilagrosMarin:
Three small drift items from a post-2.3.1 sweep, bundled since each is a one-line-ish fix but they touch unrelated files.

## Changes

**1. `tests/integration/test_trace.py:403`** — docstring cited `provenance.md, Allowed table set` as the spec anchor for the merge-table pin. That spec file was deleted when `strict_provenance` was retired; the section moved to [`trace.md §Allowed table set`](https://github.com/datajoint/datajoint-docs/blob/main/src/reference/specs/trace.md). Updated the pointer. The `autopopulate.py` docstring's own `:doc:` reference was already swept in the retirement commit — this test docstring was missed.

**2. `src/datajoint/builtin_codecs/filepath.py` (4 sites)** — docstring hardcoded `_hash/`/`_schema/` as the reserved sections, but `encode()` (`:118-134`) actually reads each store's configured `hash_prefix` and `schema_prefix` (defaults `_hash`/`_schema`). Reworded to name the settings with defaults, so users who override either prefix aren't misled by literal `_hash/` in the docs. Existing test `test_filepath_reservation_follows_hash_prefix_override` (in `tests/unit/test_codecs.py`) already covers the dynamic behavior.

**3. `RELEASE_MEMO.md` conda-forge example** — the "Check dependencies match `pyproject.toml`" template listed only 5 deps (missing 7 runtime deps: `deepdiff`, `pyparsing`, `tqdm`, `networkx`, `pydot`, `fsspec`, `pydantic-settings`), included spurious `minio` (only in test extras via `testcontainers[minio]`), and pinned `pymysql >=1.0` (actual `>=0.7.2`). Since the section explicitly instructs the reader to match `pyproject.toml`, replaced with the complete 11-dep run list from current `pyproject.toml`.

Docs-only. No behavior change.

Issue #19, opened 2014-11-14
Generate documentation


eywalker:
Generate documentation forDataJoint semi-automatically from in-code documentations. Generated documentations should be hosted on datajoint-python project site.

fabiansinz:
Documentation can now be found on http://datajoint-python.readthedocs.org/en/latest/. I haven't tested whether the github webhook runs smoothly and we might need to fix a few things along the way so that readthedocs can build the doc on their servers. So far, I made every module that readthedocs complained about a mock module (see doc/conf.py). But at least it builds the documentation I expected.  

The next step is to massively extend the documentation of the package. Do you think that should run under this issue or should we create new one and close this one?

eywalker:
Let's create a new issue to discuss documentation and close this one.

PR #1513, opened 2026-07-20
tests: add coverage for deferred audit findings + isolate test_settings from ambient env

MilagrosMarin:
Two test-suite hardening changes, all in `tests/`, no source-code changes.

## 1. Coverage for 3 audit-deferred findings (6 test functions)

Adds targeted regression guards for behaviors that had no direct coverage. Each test's docstring cites the underlying code and describes what regression it catches.

- **`tests/integration/test_dependencies.py::TestLoadAllShortCircuit`** (2 tests) — guards the ~53 ms/populate-key perf fix from [#1499](https://github.com/datajoint/datajoint-python/pull/1499). Warm up the graph, then spy on `Dependencies.load` during a second identical `load_all_upstream`/`load_all_downstream` call; assert `call_count == 0`. Regressing the early-return at `dependencies.py:266-267` and `:300-301` (e.g. `if False and ...`) makes both tests fail.

- **`tests/integration/test_gc.py::TestDeleteSchemaPathPruning`** (2 tests) — covers the parent-directory pruning loop in `delete_schema_path`. Inserts two rows into separate PK directories, deletes one, runs `collect(dry_run=False)`, then asserts: the orphan's file is gone, its now-empty PK directory is pruned, the sibling PK directory is preserved, and the store root itself survives. Regressing the `while ... rmdir` loop at `gc.py:293-302` makes `test_pruning_removes_empty_pk_directory` fail.

- **`tests/integration/test_object.py::TestStoragePathGeneration`** (2 tests) — covers `build_object_path` with `schema_prefix=""` (the documented pre-2.3.1 legacy layout — no test previously exercised it). Asserts the path is `{schema}/{table}/pk=v/field_token.dat` with no leading empty segment. Regressing the `if prefix:` guard at `storage.py:267-269` (always-prepend) makes both tests fail.

All 6 tests confirm existing behavior; no source-code changes needed. Each was verified in place: gate/remove the source condition → tests fail; restore → tests pass.

## 2. Env-isolation fixture for `tests/unit/test_settings.py`

Four tests in `TestStoreEnv` (`test_ignore_config_file_skips_secrets`, `test_ignore_config_file_default_loads_both`) and `TestBackendConfiguration` (`test_backend_postgresql`, `test_backend_env_var`) construct pydantic-settings objects that read `DJ_HOST`/`DJ_USER`/`DJ_PASS`/`DJ_PORT`/`DJ_BACKEND` from the process env. A developer with any of those vars exported (e.g. from a local `.env`) sees these tests fail even though their own contract is "ignore config file" or "auto-detect port" — because a leaked ambient value overrides the default under test.

Reproduces cleanly:
- `DJ_HOST=localhost DJ_USER=root DJ_PASS=x pytest tests/unit/test_settings.py::TestStoreEnv::test_ignore_config_file_skips_secrets` → **FAIL**
- `env -i pytest tests/unit/test_settings.py::TestStoreEnv::test_ignore_config_file_skips_secrets` → **PASS**

CI runs in a clean env, so this isn't a CI regression — but it bites devs who source project `.env` files.

Adds a **class-level `@pytest.fixture(autouse=True)`** to both `TestStoreEnv` and `TestBackendConfiguration` that `monkeypatch.delenv`'s the five `DJ_*` vars at the top of each test. Tests that explicitly `monkeypatch.setenv(...)` still work — the fixture runs first, so their setenv wins. Also protects other tests in these classes from the same latent issue (there are more than just the 4 flagged).

**Verified locally**: `pytest tests/unit/test_settings.py::TestStoreEnv tests/unit/test_settings.py::TestBackendConfiguration` → 19/19 pass with the fixture.

---

Follow-up to the v2.3.1 audit (post-#1503, #1504, #1505, #1506, #1510, #1511). Docs-issue traceable: the 3 coverage additions close deferred audit findings #7, #8, #19; the env-isolation fix surfaced during a follow-up local test run.

### DIFF
--- a/tests/integration/test_dependencies.py
+++ b/tests/integration/test_dependencies.py
@@ -1,3 +1,5 @@
+from unittest.mock import patch
+
 from pytest import raises
 
 from datajoint import errors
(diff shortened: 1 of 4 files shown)


dimitri-yatsenko:
Reviewed — verified all four files against the source; **the tests are correct and well-targeted** (test-only, no source changes). Each new test genuinely fails if its guarded behavior is regressed:

- **`TestLoadAllShortCircuit`** — the early-return `if self._loaded and known_schemas <= self._loaded_schemas: return` exists in both `load_all_downstream` and `load_all_upstream`. Spying on `load` is the right probe: discovery queries still run; only the expensive #1499 rebuild (`load()`) is skipped, so `call_count == 0` is the correct assertion.
- **`TestDeleteSchemaPathPruning`** — the prune-empty-parents `while` loop, the `fs.ls` not-empty break, the exception guard, and the `parent != root and parent.startswith(root)` store-root boundary all exist in `GarbageCollector.delete_schema_path`. Helpers (`schema_references`, `_gc`, `GcObjectTest`, and the `prefix`/`connection_test`/`mock_stores` fixtures) all resolve.
- **`TestStoragePathGeneration`** — matches `build_object_path`: empty `schema_prefix` skips the section segment via the `if prefix:` guard → `myschema/MyTable/id=42/…`; with a partition → `subject_id=1/myschema/…`. The no-leading-`/` sentinel is a nice touch.
- **env-isolation fixture** — autouse `delenv` of the five `DJ_*` vars, running before test bodies so explicit `setenv` still wins. Correctly fixes the ambient-`.env` flakiness.

**One suggestion — cite method names, not line numbers.** The `TestDeleteSchemaPathPruning` docstrings point at `gc.py:293-302` and `gc.py:295`, but the boundary `while parent and parent != root and parent.startswith(root)` is actually at `gc.py:287` on current `master` — the refs have already drifted ~8 lines (the `dependencies.py` and `storage.py` refs happen to still be exact). Since these numbers rot on every edit above them, I'd suggest citing the **method** (e.g. "the pruning loop in `delete_schema_path`") rather than `file:line` throughout the new docstrings. Keeps the regression-sentinel intent without the maintenance burden.

**Minor:** `_isolate_dj_env` is duplicated verbatim in `TestStoreEnv` and `TestBackendConfiguration` — could be one shared fixture (module-level + `usefixtures`), purely DRY. And if other classes in `test_settings.py` build settings that read `DJ_*`, they'd have the same latent ambient-env issue — out of scope here, just worth noting.

Substance looks good.

dimitri-yatsenko (approved):
Approving — the tests are correct and well-targeted (re-verified against current `origin/master` @55f3beef).

**Correction to my earlier comment:** I'd checked the line refs against a stale local tree and got the drift *backwards*. Against current master:
- `gc.py:295` (the `while parent and parent != root …` boundary) and the `293-302` range — **exact**.
- `storage.py:267-269` (`if prefix:` guard) — **exact**.
- `dependencies.py` — **this** is the drifted one: the `if self._loaded and known_schemas <= self._loaded_schemas: return` short-circuits are at **258** (downstream) / **292** (upstream) now, not the cited `266-267`/`300-301` (~8 lines off, shifted by the #1508 MultiDiGraph merge).

Which only reinforces the earlier suggestion: cite the **method** (`the pruning loop in delete_schema_path`, `the short-circuit in load_all_downstream`/`load_all_upstream`) rather than `file:line` — the numbers demonstrably rot; they even fooled me via a stale checkout. Non-blocking — fine as a quick follow-up or to leave.

Substance is solid. ✅

PR #1517, opened 2026-07-21
Remove process-wide sys.excepthook; harden log formatter (#1516)

### LINKED ISSUE #1516: Bug: DataJoint silently discards exception info for every uncaught exception
Importing `datajoint` installs a process-wide `sys.excepthook` that routes *every* uncaught
exception in the process — not just DataJoint-related ones — through the package's own logger.
That logger's formatter (`LevelAwareFormatter`) never renders `exc_info`, so the exception type,
message, and traceback are discarded before anything is printed. The operator sees only:

```
[2026-07-21 16:00:13][ERROR]: Uncaught exception
```

with no indication of what actually failed. This is strictly worse than doing nothing: Python's
default `sys.excepthook` would have printed the full traceback for free, and DataJoint's hook
suppresses that.

# LLM-analysis of the issue

## Environment

- `datajoint==2.3.1`
- Python 3.12.11 (Anaconda build, Linux x86_64)
- Reproduced both inside a project container and with a standalone script with no DB connection
  required — the bug does not depend on being connected to a database.

## Root cause

`datajoint/logging.py` does two things at import time (module-level code, so it runs the instant
anything does `import datajoint`):

1. Configures a logger named `"datajoint"` with a `StreamHandler` using a custom formatter:

   ```python
   class LevelAwareFormatter(logging.Formatter):
       """Format INFO messages cleanly, show level for warnings/errors and JOBS."""

       def format(self, record):
           timestamp = self.formatTime(record, "%Y-%m-%d %H:%M:%S")
           if record.levelno >= logging.WARNING:
               return f"[{timestamp}][{record.levelname}]: {record.getMessage()}"
           elif record.levelno == JOBS:
               return f"[{timestamp}][JOBS]: {record.getMessage()}"
           else:
               return f"[{timestamp}] {record.getMessage()}"
   ```

   This completely overrides `logging.Formatter.format()` instead of extending it. The base
   implementation, after formatting the message, checks `record.exc_info` and appends
   `self.formatException(record.exc_info)` (plus `record.stack_info` if present). This override
   does neither — it has no branch that ever looks at `record.exc_info` or
   `record.stack_info`, so any exception object attached to the record is never rendered,
   regardless of log level.

2. Replaces `sys.excepthook`:

   ```python
   def excepthook(exc_type, exc_value, exc_traceback):
       if issubclass(exc_type, KeyboardInterrupt):
           sys.__excepthook__(exc_type, exc_value, exc_traceback)
           return

       logger.error("Uncaught exception", exc_info=(exc_type, exc_value, exc_traceback))

   sys.excepthook = excepthook
   ```

   `sys.excepthook` is global interpreter state. Setting it as a side effect of `import datajoint`
   changes how *any* unhandled exception in the process is reported, including exceptions with no
   relationship to DataJoint at all (argument-parsing bugs, typos, `KeyError`s in unrelated
   application code, etc.). The `exc_info` tuple passed into `logger.error(...)` here is correct and
   complete at this point — the data exists — it is only lost one step later, inside
   `LevelAwareFormatter.format()`.

The combination means: the moment a process imports `datajoint`, every uncaught exception
anywhere in that process is silently reduced to a single, contentless log line.

This also isn't limited to the `excepthook` path. Any code — including inside `datajoint` itself —
that logs through `logging.getLogger("datajoint")` with `exc_info` set (e.g. `logger.exception(...)`
during a `populate()` call) loses the traceback the same way, since the defect is in the shared
formatter, not the hook.

## Minimal reproduction

No database connection or project code required:

```python
import datajoint  # import alone installs sys.excepthook

def inner():
    raise ValueError("something specific and diagnosable went wrong")

def outer():
    inner()

outer()
```

**Actual output:**

```
[2026-07-21 16:38:16][ERROR]: Uncaught exception
```

(Process exits 1. No exception type, message, or traceback anywhere in stdout/stderr.)

**For contrast, the same script without `import datajoint`:**

```
  File "<string>", line 8, in <module>
  File "<string>", line 6, in outer
  File "<string>", line 3, in inner
ValueError: something specific and diagnosable went wrong
```

## Impact

We hit this in production debugging: a CLI entry point (an ingestion script using DataJoint
tables) failed with a schema mismatch (`KeyError` from `Table.insert`), but the only visible
output was `[ERROR]: Uncaught exception`. Diagnosing the real cause required bypassing DataJoint's
`sys.excepthook` entirely — wrapping the call in our own `try/except` and calling
`traceback.print_exc()` manually, which formats and prints without going through DataJoint's
logger/formatter at all.

Because `sys.excepthook` is global, this silently degrades error visibility for every script and
CLI entry point in any project that imports `datajoint`, for any exception, not just DataJoint's
own. It is easy to miss because everything looks fine until the first uncaught exception occurs
in production, at which point the actual cause is unrecoverable from the process's own output —
the traceback is simply gone, not just hard to read.

## Suggested fix

`LevelAwareFormatter.format()` should render `exc_info`/`stack_info` the same way the base
`logging.Formatter` does, e.g.:

```python
class LevelAwareFormatter(logging.Formatter):
    def format(self, record):
        timestamp = self.formatTime(record, "%Y-%m-%d %H:%M:%S")
        if record.levelno >= logging.WARNING:
            message = f"[{timestamp}][{record.levelname}]: {record.getMessage()}"
        elif record.levelno == JOBS:
            message = f"[{timestamp}][JOBS]: {record.getMessage()}"
        else:
            message = f"[{timestamp}] {record.getMessage()}"

        if record.exc_info:
            if not record.exc_text:
                record.exc_text = self.formatException(record.exc_info)
        if record.exc_text:
            message = f"{message}\n{record.exc_text}"
        if record.stack_info:
            message = f"{message}\n{self.formatStack(record.stack_info)}"
        return message
```

Separately/optionally: consider whether `datajoint` should install a global `sys.excepthook` at
all as an import side effect, given it affects exceptions unrelated to DataJoint in the importing
process. At minimum, whatever hook is installed must not be lossy relative to Python's default
behavior — today it is strictly worse than doing nothing.

dimitri-yatsenko:
Fixes #1516.

## Problem

Importing `datajoint` installed a **process-wide `sys.excepthook`** as an import side effect, routing *every* uncaught exception in the host process — DataJoint-related or not — through DataJoint's logger. Combined with `LevelAwareFormatter` overriding `format()` without rendering `exc_info`, every uncaught exception collapsed to a contentless line:

```
[2026-07-21 16:00:13][ERROR]: Uncaught exception
```

with no type, message, or traceback — strictly worse than Python's default handler, which would have printed the full traceback for free.

## History (this is a regression, not the original design)

The hook was originally non-lossy: it inlined the exception message (`Uncaught exception: {exc_value}`) and rendered a full traceback under `DJ_LOG_LEVEL=DEBUG` via the stdlib formatter. Two later changes compounded into the silent-discard behavior:

- `a9dee8d9` — simplified the excepthook to rely solely on `exc_info`, dropping the inline message.
- `41dad5d2` — introduced the custom `LevelAwareFormatter`, which overrides `format()` and never renders `exc_info`.

## Change

- **Remove the `sys.excepthook` installation.** A library must not change how unrelated exceptions surface in the importing process. Uncaught exceptions now fall through to Python's default handler (full traceback to stderr). Nothing in the codebase reads the hook, and no internal code path attaches `exc_info` to a log record, so removal is self-contained.
- **Harden `LevelAwareFormatter`** to append `exc_info`/`stack_info` like the base `logging.Formatter`, so any future `logger.exception(...)` / `exc_info=…` call never silently drops a traceback.
- **Add regression tests** (`tests/unit/test_logging.py`): import no longer replaces `sys.excepthook`; the formatter renders exception and stack info.

The centralized `datajoint` logger and its handler/level config are unchanged — only the process-global excepthook is removed.

## Compatibility

Behavior change worth a release note: fatal uncaught tracebacks now print to **stderr** (Python's default handler) rather than as a single line on stdout. This is the correct destination for a crash, and restores the full type/message/traceback that the previous hook discarded.

### DIFF
--- a/src/datajoint/logging.py
+++ b/src/datajoint/logging.py
@@ -28,11 +28,22 @@ class LevelAwareFormatter(logging.Formatter):
     def format(self, record):
         timestamp = self.formatTime(record, "%Y-%m-%d %H:%M:%S")
         if record.levelno >= logging.WARNING:
-            return f"[{timestamp}][{record.levelname}]: {record.getMessage()}"
+            message = f"[{timestamp}][{record.levelname}]: {record.getMessage()}"
         elif record.levelno == JOBS:
-            return f"[{timestamp}][JOBS]: {record.getMessage()}"
+            message = f"[{timestamp}][JOBS]: {record.getMessage()}"
         else:
-            return f"[{timestamp}] {record.getMessage()}"
+            message = f"[{timestamp}] {record.getMessage()}"
+
+        # Render exception/stack info like the base logging.Formatter does, so
+        # that a logger.exception(...)/exc_info=... call never silently drops
+        # the traceback (see #1516).
+        if record.exc_info and not record.exc_text:
+            record.exc_text = self.formatException(record.exc_info)
+        if record.exc_text:
+            message = f"{message}\n{record.exc_text}"
+        if record.stack_info:
+            message = f"{message}\n{self.formatStack(record.stack_info)}"
+        return message
 
 
 # Select output stream: stdout (default, no red highlighting) or stderr
@@ -44,13 +55,7 @@ stream_handler.setFormatter(LevelAwareFormatter())
 logger.setLevel(level=log_level)
 logger.handlers = [stream_handler]
 
-
-def excepthook(exc_type, exc_value, exc_traceback):
-    if issubclass(exc_type, KeyboardInterrupt):
-        sys.__excepthook__(exc_type, exc_value, exc_traceback)
-        return
-
-    logger.error("Uncaught exception", exc_info=(exc_type, exc_value, exc_traceback))
-
-
-sys.excepthook = excepthook
+# NOTE: DataJoint intentionally does NOT install a process-wide sys.excepthook.
+# Importing a library must not change how *unrelated* uncaught exceptions are
+# reported in the host process. Uncaught exceptions are left to Python's default
+# handler, which prints the full type/message/traceback to stderr. (See #1516.)
--- a/tests/unit/test_logging.py
+++ b/tests/unit/test_logging.py
@@ -1,0 +1,70 @@
+"""
+Tests for datajoint.logging.
+
+Regression coverage for #1516: importing datajoint must not install a
+process-wide sys.excepthook, and LevelAwareFormatter must not discard
+exc_info/stack_info.
+"""
+
+import logging
+import sys
+
+
+def test_import_does_not_replace_excepthook():
+    """Importing datajoint must leave sys.excepthook untouched (#1516)."""
+    original = sys.excepthook
+    try:
+        modules_to_remove = [key for key in sys.modules if key.startswith("datajoint")]
+        for mod in modules_to_remove:
+            del sys.modules[mod]
+
+        import datajoint  # noqa: F401
+
+        assert sys.excepthook is original, "importing datajoint replaced sys.excepthook"
+    finally:
+        sys.excepthook = original
+
+
+def test_formatter_renders_exception_info():
+    """LevelAwareFormatter must append the traceback when exc_info is set (#1516)."""
+    from datajoint.logging import LevelAwareFormatter
+
+    formatter = LevelAwareFormatter()
+    try:
+        raise ValueError("something specific and diagnosable went wrong")
+    except ValueError:
+        record = logging.LogRecord(
+            name="datajoint",
+            level=logging.ERROR,
+            pathname=__file__,
+            lineno=0,
+            msg="Uncaught exception",
+            args=(),
+            exc_info=sys.exc_info(),
+        )
+
+    output = formatter.format(record)
+    assert "Uncaught exception" in output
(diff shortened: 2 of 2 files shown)


MilagrosMarin (approved):
LGTM.

Two small non-blocking suggestions:

1. **Formatter tests only exercise WARNING/ERROR levels.** exc_info/stack_info handling is appended after the level branches, so it applies uniformly at runtime — but a `pytest.parametrize` on level across `INFO`, `JOBS`, `WARNING`, `ERROR` would catch a future regression that adds an early `return message` inside the `JOBS` or `INFO` branch.

2. **Out of scope, follow-up thought:** `logging.py:17` still does `logging.Logger.jobs = jobs` — same architectural pattern as the excepthook you're removing (library patching global state at import time). A `DataJointLogger(logging.Logger)` subclass would be the polite version, but separate PR.

PR #1518, opened 2026-07-21
Fix jobs table generation for uuid FK-derived primary keys (#1515)

### LINKED ISSUE #1515: Bug: Auto-generated jobs table crashes when a Computed/Imported table's primary key derives from a `uuid` attribute
Calling `populate(reserve_jobs=True)` (or any other access to `Table.jobs`, e.g.
`Table.jobs.refresh()`) on a `Computed`/`Imported` table whose primary key derives — via a
primary foreign key — from a `uuid`-typed attribute crashes with:

```
DataJointError: Unsupported attribute type binary(16)
```

# LLM-analysis of the issue

The error message points at `binary(16)`, which is never a type a user writes in a table
definition — it is the resolved MySQL column type DataJoint substitutes internally for the
`uuid` core type alias. The auto-generated jobs table definition leaks this internal, already-
resolved SQL type back into a fresh DataJoint definition string, which then fails to parse.

## Environment

- `datajoint==2.2.4`
- MySQL backend
- Reproduced with a table whose primary key is entirely FK-derived from a parent table with a
  `uuid` primary key attribute — the officially supported/recommended shape for auto-populated
  tables (`_declare_check` in `autopopulate.py` explicitly requires FK-only primary keys).

## Root cause

`datajoint/jobs.py`, `Job._generate_definition()` builds the DataJoint definition string for
the `~~<table_name>` job-queue table by walking the target table's FK-derived primary key
attributes and emitting `"{name} : {dtype}"` for each one:

```python
# jobs.py:180-219 (_get_fk_derived_pk_attrs)
for name in target_pk:
    if name in fk_derived_attrs:
        attr = heading[name]
        fk_attrs.append((name, attr.type))   # <-- line 211
    ...
```

```python
# jobs.py:141-178 (_generate_definition)
pk_lines = "\n    ".join(f"{name} : {dtype}" for name, dtype in pk_attrs)
return f"""
# Job queue for {self._target.full_table_name}
{pk_lines}
---
...
"""
```

`attr.type` is the **already-resolved SQL type** for the attribute (e.g. `binary(16)` for a
`uuid` column), not the DataJoint-level type alias the user originally wrote (`uuid`). DataJoint
heading attributes carry both:

```python
# heading.py:60-63
default_attribute_properties = dict(
    ...
    type="expression",
    original_type=None,  # For core types, stores the alias (e.g., "uuid") while type has db type ("binary(16)")
    ...
)
```

and DataJoint's own code elsewhere correctly falls back to `original_type` when it needs the
alias back (`heading.py:496`: `original_type = attr["original_type"] or attr["type"]`). `jobs.py`
does not do this — it always uses `attr.type` — so for a `uuid` PK attribute, the generated jobs
table definition contains a line like:

```
recording_id : binary(16)
```

When `Job.declare()` parses this string, `compile_attribute()` calls `match_type("binary(16)")`
(`declare.py:910`). `TYPE_PATTERN` has no entry that matches a bare `binary(...)` SQL type — it
only has `uuid` (`r"uuid$"`), native blob variants, and other named categories — so `match_type`
falls through every pattern and raises:

```python
# declare.py:110-113
try:
    return next(category for category, pattern in TYPE_PATTERN.items() if pattern.match(attribute_type))
except StopIteration:
    raise DataJointError("Unsupported attribute type {type}".format(type=attribute_type))
```

producing `DataJointError: Unsupported attribute type binary(16)`.

Note: `Job.__init__` calls `_generate_definition()` unconditionally, but `Job.declare()` is only
invoked by `autopopulate.py`'s `_JobsDescriptor` when `not obj._jobs.is_declared` (i.e. the
`~~table` does not yet exist in the database). So the crash is specifically triggered the first
time `.jobs` is accessed for a given table (including the implicit access inside
`populate(reserve_jobs=True)`) — any `Computed`/`Imported` table whose FK-derived primary key
chain passes through a `uuid` attribute will hit this on first populate/jobs access, and never
recover, since `is_declared` will keep being `False`.

## Minimal reproduction

```python
import datajoint as dj

schema = dj.Schema("repro_uuid_jobs")

@schema
class Parent(dj.Manual):
    definition = """
    parent_id : uuid
    """

@schema
class Child(dj.Computed):
    definition = """
    -> Parent
    ---
    value : int32
    """

    def make(self, key):
        self.insert1(dict(key, value=1))

Child.populate(reserve_jobs=True)
```

**Actual result:**

```
DataJointError: Unsupported attribute type binary(16)
```

**Expected result:** the jobs table declares successfully with `parent_id : uuid` (which
DataJoint itself would then correctly resolve to `binary(16)` internally), and `populate`
proceeds normally.

## Impact

Any pipeline that uses `uuid` as a stable, filesystem-independent identifier for a root table
(a common and DataJoint-recommended pattern for deterministic, content-addressed keys) and
computes downstream tables from it via FK cannot use `reserve_jobs=True` at all — the very
feature intended for safe distributed/parallel `populate()` calls. There is no user-facing
workaround via the table definition; the bug is entirely in the internal jobs-table
autogeneration path and fires before any user `make()` code runs.

## Suggested fix

In `jobs.py`, `_get_fk_derived_pk_attrs` should use the same fallback pattern DataJoint uses
elsewhere for recovering the original core-type alias:

```python
attr = heading[name]
fk_attrs.append((name, attr.original_type or attr.type))
```

This mirrors `heading.py:496` (`original_type = attr["original_type"] or attr["type"]`) and
ensures the regenerated definition string contains DataJoint-level type syntax (`uuid`,
`float32`, etc.) rather than backend-resolved SQL types, so `compile_attribute`/`match_type` can
parse it correctly regardless of which core type the FK-derived primary key attribute uses.

dimitri-yatsenko:
Fixes #1515.

## Problem

Accessing `.jobs` on a `Computed`/`Imported` table whose primary key is FK-derived from a `uuid` attribute crashed with:

```
DataJointError: Unsupported attribute type binary(16)
```

This fires on the first `.jobs` access for such a table — including the implicit access inside `populate(reserve_jobs=True)` — and never recovers, since the `~~jobs` table stays undeclared. It disables `reserve_jobs=True` entirely for the common, recommended pattern of using `uuid` root keys with downstream computed tables.

## Root cause

`Job._generate_definition()` builds the `~~jobs` table's DataJoint definition string from the target's FK-derived primary key. For each attribute it emitted `attr.type` — the **backend-resolved SQL type** (`binary(16)` for a `uuid`) — rather than the DataJoint-level alias the user wrote (`uuid`). `binary(16)` is not valid DataJoint definition syntax, so when `Job.declare()` re-parsed the generated string, `match_type("binary(16)")` matched no pattern and raised.

## Fix

Use `attr.original_type or attr.type` in `_get_fk_derived_pk_attrs`, mirroring the fallback DataJoint already uses in `heading.py` (`original_type = attr["original_type"] or attr["type"]`). `original_type` holds the DataJoint alias for core types (`uuid`, `float32`, …) and is `None` otherwise, so:

- `uuid` PK → definition line `item : uuid` (re-parses correctly; DataJoint resolves it back to `binary(16)` internally).
- any other core-type alias in an FK-derived PK is fixed the same way, not just `uuid`.
- non-alias types are unaffected (fall back to `attr.type` as before).

## Test

Adds `test_jobs_table_uuid_fk_derived_pk` (plus a `BasicComputed` table in the uuid test schema: a `Computed` table whose PK is FK-derived from a `uuid` parent). It asserts the generated definition carries `uuid` (not `binary(16)`), that `.jobs.refresh()` declares the table, and that the PK attribute round-trips to `original_type == "uuid"`.

### DIFF
--- a/src/datajoint/jobs.py
+++ b/src/datajoint/jobs.py
@@ -206,9 +206,14 @@ class Job(Table):
         fk_attrs = []
         for name in target_pk:
             if name in fk_derived_attrs:
-                # FK-derived: comes from a primary FK parent
+                # FK-derived: comes from a primary FK parent.
+                # Use original_type (the DataJoint-level alias, e.g. "uuid") when
+                # present, falling back to the resolved SQL type. attr.type holds
+                # the backend-resolved type (e.g. "binary(16)" for uuid), which is
+                # not valid DataJoint definition syntax and fails to re-parse when
+                # the jobs table is declared. Mirrors heading.py's own fallback.
                 attr = heading[name]
-                fk_attrs.append((name, attr.type))
+                fk_attrs.append((name, attr.original_type or attr.type))
             else:
                 # Native PK attribute - not from FK
                 logger.warning(
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -863,6 +863,7 @@ def schema_uuid(connection_test, prefix):
         connection=connection_test,
     )
     schema(schema_uuid_module.Basic)
+    schema(schema_uuid_module.BasicComputed)
     schema(schema_uuid_module.Topic)
     schema(schema_uuid_module.Item)
     yield schema
--- a/tests/integration/test_jobs.py
+++ b/tests/integration/test_jobs.py
@@ -7,6 +7,7 @@ import datajoint as dj
 from datajoint.jobs import ERROR_MESSAGE_LENGTH, TRUNCATION_APPENDIX
 
 from tests import schema
+from tests.schema_uuid import BasicComputed
 
 
 def test_reserve_job(clean_jobs, subject, experiment):
@@ -206,3 +207,26 @@ def test_jobs_refresh_with_keep_completed(clean_jobs, subject, experiment):
 
         # Calling refresh again should not raise semantic matching error
         experiment.jobs.refresh()  # This was failing before the fix
+
+
+def test_jobs_table_uuid_fk_derived_pk(schema_uuid):
+    """Jobs table generation must handle uuid-typed FK-derived primary keys (#1515).
+
+    The auto-generated jobs table definition previously leaked the resolved SQL
+    type (binary(16)) for a uuid PK attribute instead of the DataJoint alias
+    (uuid), so declaring the ~~table raised
+    "Unsupported attribute type binary(16)" on first .jobs access.
+    """
+    # Definition must carry the DataJoint alias, not the resolved SQL type.
+    definition = BasicComputed.jobs.definition
+    assert "uuid" in definition
+    assert "binary(16)" not in definition
+
+    # First .jobs access declares the ~~table; this is where the bug fired.
+    BasicComputed.jobs.refresh()
+    assert BasicComputed.jobs.is_declared
+
+    # The declared PK attribute round-trips back to uuid / binary(16).
+    pk = BasicComputed.jobs.heading.primary_key
+    assert "item" in pk
+    assert BasicComputed.jobs.heading["item"].original_type == "uuid"
--- a/tests/schema_uuid.py
+++ b/tests/schema_uuid.py
@@ -14,6 +14,18 @@ class Basic(dj.Manual):
     """
 
 
+class BasicComputed(dj.Computed):
+    definition = """
+    # Computed table whose primary key is FK-derived from a uuid attribute
+    -> Basic
+    ---
+    value : int32
+    """
+
+    def make(self, key):
+        self.insert1(dict(key, value=1))
+
+
 class Topic(dj.Manual):
     definition = """
     # A topic for items


MilagrosMarin (approved):
LGTM.

Small strengthening: the fix mirrors the same `original_type or type` fallback in TWO existing sites — `heading.py:543` AND `table.py:1327` (inside `Table.describe`). Worth mentioning the second precedent in the PR body — reinforces "jobs.py was the outlier missing the pattern" rather than "one-off similarity".

Optional hardening: test uses `uuid` only, but the fix generalizes to any core-type alias. A parameterized second case (e.g., `decimal(6,3)` PK) would catch a hypothetical mis-regression like `if original_type == "uuid": ... else: attr.type`. Non-blocking — the fix generalizes by construction (the `or` short-circuit applies to any attribute whose `original_type` is set).

Issue #1520, opened 2026-07-23
Bug: GarbageCollector deletes live files on Windows due to backslash/forward-slash path mismatch in file-protocol stores


lochhh:
### Description
On Windows, `dj.gc.GarbageCollector.collect()` treats every live row's storage file as orphaned, and `dry_run=False` deletes files that are still referenced by live rows.

### Root cause

`StorageBackend._full_path()` (`datajoint/storage.py`, `protocol == "file"` branch) builds the path with:

​```python
return str(Path(location) / path)
​```

`str(Path(...))` renders OS-native separators (backslashes on Windows). But `fsspec`'s `LocalFileSystem.walk()` (used by `GarbageCollector.list_hash_paths`/`list_schema_paths`) always returns POSIX-style forward-slash paths.

`gc.py` computes `full_root = backend._full_path("")` (backslash-separated) and then does `file_path.replace(full_root, "")` where `file_path` comes from `fs.walk()` (forward-slash). The `.replace()` silently no-ops on Windows, so the "relative" paths returned by `list_hash_paths`/`list_schema_paths` are actually still full absolute paths. These never match the (correctly relative) referenced-path set built from row metadata, so `_is_covered()` always returns `False`, i.e. every stored object looks orphaned.

### Reproducibility
Include:
- WIN 11
- Python 3.12
- DataJoint Version ~(master branch)~ 2.3.1

The test introduced in https://github.com/datajoint/datajoint-python/pull/1521/changes/03436ff3d9f26bcbaec16ea34e025210c26907dc fails with 
> FAILED tests/unit/test_storage_adapter.py::TestStorageBackendPluginDelegation::test_file_protocol_full_path_uses_forward_slashes - AssertionError: assert 'data\\blobs\...\\cd\\hash123' == 'data/blobs/s...ab/cd/hash123'

lochhh:
## Correction to root cause
Apologies, turns out I was on v2.3.1, where gc.py computed relative paths via:
`file_path.replace(full_root, "")`
That is confirmed broken on Windows as described: backslash `full_root` vs forward-slash `file_path` → `.replace()` no-ops → `_is_covered()` always `False` → mass false-positive orphans.

This issue **no longer exists** on 2.3.2, given that `gc.py`'s `list_hash_paths`/`list_schema_paths`/`objectref.walk()` now use index-based slicing:
`relative_path = file_path[len(full_root):].lstrip("/")`

PR #1521, opened 2026-07-23
Use POSIX separators in file-protocol storage URLs and paths

lochhh:
`StorageBackend._full_path()` returned OS-native separators for the **file** protocol (via bare `str(Path(...))`), so paths came out with backslashes on Windows. This silently broke `delete_schema_path`'s empty parent dir pruning, which silently no-ops on backslash paths:
https://github.com/datajoint/datajoint-python/blob/afc350f2eb75f0cf7fad94d0756d6cbdecb56d46/src/datajoint/gc.py#L293-L294

This PR 
- fixes `_full_path()` such that it now returns `/`-separated paths for the **file** protocol
- additionally cleans up `normalize_to_url()` and `get_url()` which already handled Windows
paths correctly, but did so via a duplicated ad-hoc `.replace(chr(92), '/')`, by extracting a shared `_path_to_file_url()` helper (`.as_posix()` + lstrip) to match the `.as_posix()` idiom in `_full_path()`
- adds the following tests:
  - Platform-independent coverage via `PureWindowsPath`/`PurePosixPath` for `_full_path` and `_path_to_file_url`
  - Real `tmp_path` wiring test for `get_url` under the `file` protocol
  - Regression test for `delete_schema_path` parent-dir pruning

### DIFF
--- a/src/datajoint/storage.py
+++ b/src/datajoint/storage.py
@@ -418,7 +418,7 @@ class StorageBackend:
         elif self.protocol == "file":
             location = self.spec.get("location", "")
             if location:
-                return str(Path(location) / path)
+                return (Path(location) / path).as_posix()
             return path
         else:
             return self._require_adapter().full_path(self.spec, path)
--- a/src/datajoint/storage.py
+++ b/src/datajoint/storage.py
@@ -45,6 +45,27 @@ def is_url(path: str) -> bool:
     return path.lower().startswith(URL_PROTOCOLS)
 
 
+def _path_to_file_url(resolved_path: Path | PurePosixPath) -> str:
+    """
+    Convert an already-resolved absolute path to a ``file://`` URL.
+
+    Uses ``as_posix()`` so the same logic handles both POSIX paths (which
+    already start with ``/``) and Windows paths (``C:/...``, no leading
+    slash) without OS-specific branching.
+
+    Parameters
+    ----------
+    resolved_path : Path or PurePosixPath
+        Absolute, already-resolved path.
+
+    Returns
+    -------
+    str
+        ``file://`` URL.
+    """
+    return f"file:///{resolved_path.as_posix().lstrip('/')}"
+
+
 def normalize_to_url(path: str) -> str:
     """
     Normalize a path to URL form.
@@ -72,15 +93,7 @@ def normalize_to_url(path: str) -> str:
     """
     if is_url(path):
         return path
-    # Convert local path to file:// URL
-    # Ensure absolute path and proper format
-    abs_path = str(Path(path).resolve())
-    # Handle Windows paths (C:\...) vs Unix paths (/...)
-    if abs_path.startswith("/"):
-        return f"file://{abs_path}"
-    else:
-        # Windows: file:///C:/path
-        return f"file:///{abs_path.replace(chr(92), '/')}"
+    return _path_to_file_url(Path(path).resolve())
 
 
 def parse_url(url: str) -> tuple[str, str]:
@@ -453,13 +466,7 @@ class StorageBackend:
         full_path = self._full_path(path)
 
         if self.protocol == "file":
-            # Ensure absolute path for file:// URL
-            abs_path = str(Path(full_path).resolve())
-            if abs_path.startswith("/"):
-                return f"file://{abs_path}"
-            else:
-                # Windows path
-                return f"file:///{abs_path.replace(chr(92), '/')}"
+            return _path_to_file_url(Path(full_path).resolve())
         elif self.protocol == "s3":
             return f"s3://{full_path}"
         elif self.protocol == "gcs":
--- a/tests/unit/test_gc.py
+++ b/tests/unit/test_gc.py
@@ -1,0 +1,35 @@
+"""Tests for garbage collection module."""
+
+from pathlib import PurePosixPath, PureWindowsPath
+from unittest.mock import MagicMock
+
+import pytest
+
+from datajoint import storage
+from datajoint.gc import GarbageCollector
+from datajoint.storage import StorageBackend
+
+
+@pytest.mark.parametrize(
+    ("path_cls", "location"),
+    [
+        pytest.param(PureWindowsPath, "data\\blobs", id="windows"),
+        pytest.param(PurePosixPath, "data/blobs", id="posix"),
+    ],
+)
+def test_delete_schema_path_prunes_parent_dir(monkeypatch, path_cls, location):
+    """Pruning's `rsplit("/", 1)` needs forward slashes; `windows` param catches
+    a regression to `str(Path(...))`, `posix` pins the already-correct case."""
+    monkeypatch.setattr(storage, "Path", path_cls)
+    backend = StorageBackend.__new__(StorageBackend)
+    backend.spec = {"protocol": "file", "location": location}
+    backend.protocol = "file"
+    backend._fs = MagicMock()
+    backend.fs.exists.return_value = True
+    # Ensure rmdir called only once: non-empty dir stops the walk one level up
+    backend.fs.ls.side_effect = lambda p: [] if p == "data/blobs/schema/ab/cd" else ["sibling"]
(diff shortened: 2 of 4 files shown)


dimitri-yatsenko (commented):
Nice catch, and the right fix. `.as_posix()` on the `file` branch of `_full_path` also completes a convention already used elsewhere in this file — `get_url` and `normalize_to_url` both force forward slashes for Windows via `.replace(chr(92), '/')`, and `_full_path`'s `file` branch was the one spot still emitting native separators. Every file-protocol operation routes through `_full_path` (`put_file`/`get_file`/`put`/`get_buffer`/`exists`/`isdir`/`remove`/`size`/`open`/`put_folder`/`get_url`), so this single change covers all consumers, and fsspec's `LocalFileSystem` accepts forward-slash (incl. drive-letter) paths on Windows — no risk to the read/write paths. The `PureWindowsPath` monkeypatch makes the test platform-independent (runs on the Linux CI) and it fails pre-fix — nice.

LGTM to merge. One thing worth confirming, and one FYI:

**Confirm the end-to-end GC path (not blocking, but worth a line).** The new test asserts the `_full_path` separator contract, but it doesn't exercise the actual misclassification (`list_hash_paths`/`list_schema_paths` → `_is_covered`). That matters because the CI matrix has no Windows runner, and on Linux `.as_posix()` == `str(...)`, so nothing automated exercises the failing path. Also, #1520 describes the mechanism as `file_path.replace(full_root, "")`, but the code on master uses a length slice — `file_path[len(full_root):].lstrip("/")` (gc.py:212, 261) — which is largely insensitive to separator *style*. So it'd be good to either add a small integration test (walk a `file` store with a Windows-style `_full_path` and assert the relative paths match a reference set) or confirm the manual Win11 `collect()` repro now passes end-to-end — just so we're sure separators are the true root cause rather than, say, a relative-vs-absolute prefix mismatch.

**FYI:** added the `bug` label and the **v2.3.3** milestone so this data-loss fix lands in the release-drafter notes.

Nit (non-blocking): this file now has two idioms for the same thing — `.as_posix()` (new) vs `str(...).replace(chr(92), '/')` (lines 83, 462). `.as_posix()` is cleaner; worth unifying in a follow-up.

dimitri-yatsenko:
Opened #1522 to add a **Windows unit-test leg** to CI (`windows-latest`, container-free `tests/unit` on Python 3.10 and 3.14). The matrix was Linux-only, so this class of OS-specific bug — path separators here — had no automated guard; #1522 gives us the runner so this regression stays caught. It uses pip (`.[test]`) rather than pixi, since pixi targets linux/osx only.

That addresses the CI half of the coverage note above. The end-to-end GC integration test (walking a `file` store and asserting relative paths match references) can now land on top of that runner as a follow-up.

lochhh:
I've closed https://github.com/datajoint/datajoint-python/issues/1520#issuecomment-5061245224.
But switching to `as_posix` fixes a separate minor issue: `delete_schema_path`'s empty parent dir pruning silently no-ops on backslash paths, skipping cleanup.
https://github.com/datajoint/datajoint-python/blob/afc350f2eb75f0cf7fad94d0756d6cbdecb56d46/src/datajoint/gc.py#L293-L294
I've updated the PR title and description to reflect this.

dimitri-yatsenko:
Re-reviewed after the reframe (the #1520 GC-misclassification rationale withdrawn, PR now justified by `delete_schema_path`'s empty-parent pruning). Verdict unchanged — **approve**. I traced the code and ran the test.

**The reframed bug is real.** `gc.py`'s prune loop (293-302) walks up empty parent dirs after deleting an orphan. Pre-fix on a Windows-style path:
- `full_path = _full_path(path)` → `data\blobs\…\field_token` (backslashes)
- `parent = full_path.rsplit("/", 1)[0]` → no `/` to split on, so `parent` is the whole *file* path, not its directory
- `fs.ls(parent)` on a file returns that file (non-empty) → `break` → pruning silently no-ops; empty dirs leak

`.as_posix()` gives `data/blobs/…`, so `rsplit("/")` yields the real parent and the walk-up works. Confirmed.

This also retroactively fits the #1520 withdrawal: `list_*_paths` strips the root with a **length slice** (`file_path[len(full_root):]`), which is insensitive to separator *style* — so separators were never the #1520 root cause.

**No regression.** Every file-protocol consumer of `_full_path` (`objectref`, `npy`, `staged_insert`, all `storage.py` ops) routes through fsspec's `LocalFileSystem`, which accepts forward-slash paths on Windows; on Linux `.as_posix() == str(...)`, byte-identical. `get_url` re-resolves via `Path(full_path).resolve()` independently, so it doesn't double-handle.

**Test.** Passes on the branch; I confirmed it fails pre-fix (`str(...)` → `data\blobs\…`), so it's a valid regression guard, and the `PureWindowsPath` monkeypatch makes it run on the Linux CI.

**One open item (non-blocking):** the test asserts only the `_full_path` unit contract — it doesn't exercise `delete_schema_path`'s pruning, which is now the PR's sole justification, and the matrix has no Windows runner yet. Covered by the plan (#1522 Windows leg + GC integration test as follow-up), so fine to merge — but a cheap interim guard would be a unit test that calls `delete_schema_path` against a fake `fs` with a monkeypatched Windows-style `_full_path` and asserts `rmdir` is invoked on the parent dir.

**Minor:** two forward-slash idioms now coexist — `.as_posix()` (new) vs `.replace(chr(92), '/')` at `storage.py:83,462`. Those operate on `abspath` strings, not `Path` objects, so they can't trivially adopt `.as_posix()`; worth unifying in a follow-up.

lochhh:
Thanks. I've updated the PR title and description to additionally cover the below.

> **One open item (non-blocking):** the test asserts only the `_full_path` unit contract — it doesn't exercise `delete_schema_path`'s pruning, which is now the PR's sole justification, and the matrix has no Windows runner yet. Covered by the plan (#1522 Windows leg + GC integration test as follow-up), so fine to merge — but a cheap interim guard would be a unit test that calls `delete_schema_path` against a fake `fs` with a monkeypatched Windows-style `_full_path` and asserts `rmdir` is invoked on the parent dir.

[Test delete_schema_path parent dir pruning](https://github.com/datajoint/datajoint-python/pull/1521/commits/67df7815e0c1265117c50fcccc2f33888c5d980f)

> **Minor:** two forward-slash idioms now coexist — `.as_posix()` (new) vs `.replace(chr(92), '/')` at `storage.py:83,462`. Those operate on `abspath` strings, not `Path` objects, so they can't trivially adopt `.as_posix()`; worth unifying in a follow-up.

[Unify file:// URL slash handling for Windows and POSIX paths](https://github.com/datajoint/datajoint-python/pull/1521/commits/1852c1a3f48e8253a0a0879aaf0f66aecf91ecc0)

dimitri-yatsenko:
The added `test_delete_schema_path_prunes_parent_dir` closes the open item exactly as hoped — thanks for the quick turnaround. Reviewed the diff; it's the right shape:

- It drives `delete_schema_path` against a mocked `fs` with `storage.Path` monkeypatched to `PureWindowsPath`, so the Windows path behavior is exercised on any OS — no runner dependency.
- The `windows` param would fail on a regression back to `str(Path(...))` (backslash path → `rsplit("/", 1)` finds nothing to split → parent stays the file path → no `rmdir`), and the `posix` param pins the already-correct case.
- Asserting `rmdir.assert_called_once_with("data/blobs/schema/ab/cd")` together with the `ls` side-effect that returns a sibling one level up also confirms the walk *stops* at the first non-empty parent — not merely that pruning fires.

Because these guards are platform-independent, they already catch this class on the current Linux CI; the Windows leg in #1522 is defense-in-depth on top. Nothing further needed from my side — good to merge once green.

dimitri-yatsenko (approved):
Approving. CI is green across the full matrix (py3.10/3.14 × MySQL 8.0/8.4, unit-tests, lint), the reframed `delete_schema_path` pruning bug is real, and the added regression test guards it platform-independently. Good to merge.

Issue #1501, opened 2026-07-16
Diagram.restrict: apply Part conditions through the Master to preserve master-part compositional integrity


dimitri-yatsenko:
## Summary

`dj.Diagram.restrict` is logically **conjunctive** (AND — it removes rows that fail the criteria).
Unlike the **disjunctive** `trace`/`cascade` (which include everything touched and keep/pull whole
compositional units), a conjunctive filter can cut *into* a master-part group, leaving **partial
groups** — a Master exported with only a subset of its Parts (or none). This preserves *referential*
integrity but breaks *compositional* integrity: the exported entity misrepresents the original.

## Reproduction (structural)

```python
dj.Diagram(schema).restrict(Master & 'm=1').restrict(Master.Part & 'p>3')
```

- `Master` → all of `m=1`.
- `Master.Part` → (Parts of `m=1`, from the master-keyed downstream restriction) **AND** `p>3` → only
  the `p>3` Parts.
- Result: `Master m=1` whole, but only a **subset** of its Parts. A Master can even survive with **zero**
  Parts, since `restrict` does not walk up from Part to Master.

## Why it matters

- **Referential integrity — fine.** A kept Part's FK to its Master is satisfied; a Master needs no Parts
  to be valid. The export re-imports without dangling foreign keys.
- **Compositional integrity — broken.** DataJoint's master-part contract treats a Master's Parts as one
  atomic entity. A recording exported with only some of its channels loads cleanly but is a
  corrupted/partial entity.

## Root cause / model

In DataJoint's model the **Master is the entity**; Parts have no independent existence. The unit of
export is therefore the Master-with-all-its-Parts. A Part-level filter is either meaningless for export,
or should mean "Masters that have a matching Part."

## Decision (v2.4)

**In Diagram restriction, conditions on Part tables are applied through their Master:**

- A restriction that selects Part rows lifts to an **existential** restriction on the Master — Masters
  that have a matching Part are selected.
- A selected Master brings in **all** of its Parts.
- Net: compositional units are never split; Parts are never filtered within a unit.

This is the compositional-integrity analogue, for read-only restriction/export, of what `cascade`'s
`part_integrity="cascade"` does for delete.

## Scope / relationships

- **Not for 2.3.1** — behavior change deferred to **v2.4**.
- Distinct from #1500 (which established that `restrict` does not perform `cascade`'s *deletion*
  part→master walk-up). This issue is about **filtering granularity** — keeping units whole under a
[...]

Issue #1496, opened 2026-07-15
Cascade delete: a restriction referencing an earlier-deleted (downstream) table is silently invalidated by reverse-topological delete order — materialize (independent of MySQL 1093)


dimitri-yatsenko:
## Summary

A cascade delete whose restriction references a table that will be deleted **earlier** in the cascade
is silently invalidated by DataJoint's **reverse-topological delete order** — and this is **independent
of MySQL error 1093**. `Table.delete` deletes descendants (leaves) first and the seed last; if a
restriction references a table deleted before its own, that reference evaluates to **empty** by the
time its DELETE runs, so the row is stranded or mis-deleted. The fix is to **materialize** such
restrictions to literal keys *before* any deletion. MySQL 1093 is only an incidental, partial symptom
(see "Backend symptoms") — not the cause, and it must not shape the fix.

## Root cause (backend-independent)

`Table.delete` executes one DELETE per table in reverse-topological order (`table.py:1089`), leaves
first, seed last. Restrictions are evaluated **live** at each table's delete time. Two kinds of
restriction reference a table deleted *earlier*:

1. **User seed restriction referencing a descendant.** `(A & (X & cond)).delete()` with `X` downstream
   of `A`: `X` (descendant) is deleted first; then `A`'s DELETE re-evaluates `A & (X & cond)`, but the
   matching `X` rows are gone → `A` matches nothing → **`A` stranded** while its `X` children were
   deleted.
2. **Engine Part→Master upward reference.** The Master's restriction is derived from its Part; the Part
   (descendant) is deleted first → the Master would strand. Already materialized today in
   `_propagate_part_to_master`.

Both are the same bug — *a restriction referencing a table deleted before its own* — on **both** backends.

## Backend symptoms (secondary — NOT the framing)

In the *sub-case* where forward propagation makes a table's restriction reference **itself**, the
generated DELETE contains a self-referential subquery. MySQL rejects it with **error 1093** and aborts
(an incidental, loud, partial backstop); PostgreSQL permits it and fails **silently**. But 1093 only
covers the self-referential subset — the seed-stranding case (#1: `A`'s DELETE references `X`, not `A`)
does **not** trigger 1093 on either backend and is silent wherever it isn't materialized. So 1093 is
neither necessary nor sufficient to describe the problem.

## Recommended approach

**Materialize any cascade restriction that references a table deleted before it** (a descendant in
[...]

PR #1524, opened 2026-07-30
Design: expand/restrict diagram-traversal algebra (2.4)

dimitri-yatsenko:
## What

Proposed **2.4** design (docs-only) deriving a small, composable algebra for diagram traversal that collapses `Diagram.cascade` / `Diagram.trace` / `Diagram.restrict` into two operations. Adds `DESIGN-expand-restrict.md`; no code changes.

## The design

- **One primitive** — propagate a restriction across a foreign key by **semijoin**, either direction.
- **R1 (edge rule)** — the semijoin, with an identity fast-path when the FK is the whole primary key and unrenamed. Absorbs the former forward F1–3 / upward U1–3 rules into one.
- **R2 (group rule)** — a restriction touching any part of a master's group brings the whole group (existential lift part→master, re-expand to all parts).
- **`Diagram.expand(seed, direction="down"|"up"|"both")`** — additive **constructor**; grows a related sub-diagram from a single seed. `cascade` = `expand("down")` (delete blast radius) and `trace` = `expand("up")` (valid `make()` sources) are **retained as aliases**; `self.upstream` = `expand(self & key, "up")`.
- **`diagram.restrict(*conditions, direction=...)`** — subtractive **progressive carving** on any diagram (including one from `expand`): each table ends restricted by the conjunction of conditions reaching it. Chainable, order-independent, monotone.
- **Composition** — both are transforms over one representation (union to grow, intersection to carve), so they chain freely; the current "cascade and restrict are mutually exclusive" restriction goes away.

Includes the "all data for `mouse_id=5` **and** `method_id=5`" counterexample showing `restrict` is **not** reducible to combinations of `expand`.

## Scope / relationship to 2.3.3

This is a **2.4 API evolution**, not part of the 2.3.3 bug-fix line. The compositional-integrity issues #1496 / #1501 / #1481 fold into this redesign rather than shipping as standalone 2.3.3 patches. (The multi-FK-path walk + alias-docstring scrub already landed separately on the 2.3.3 branch.)

## Open for review

- `restrict` direction default and whether a descendant condition also trims ancestors (`both`) for a fully-consistent export slice.
- **#1481** — `expand("up")` applies R2, so `trace` descends from an ancestor master into its parts (reproducibility-contract grounds); this flips the currently-pinned `test_trace_stops_at_master_no_part_down_collection`.
- Platform behavior (Pipeline Navigator progressive-vs-batch carving) is a product decision; the library guarantees only the algebra.

Draft — opening for design review before any code or spec refactor.

### DIFF
--- a/DESIGN-expand-restrict.md
+++ b/DESIGN-expand-restrict.md
@@ -1,0 +1,176 @@
+# Design: `expand` and `restrict` — the diagram-traversal algebra
+
+Status: proposed (targets the 2.4 line). Supersedes the separate
+`Diagram.cascade` / `Diagram.trace` / `Diagram.restrict` operators with a small,
+composable algebra. `cascade` and `trace` are retained as aliases.
+
+This document derives the design from first principles so the rules — not just
+the API — are the thing agreed upon.
+
+---
+
+## 1. The one primitive: propagating a restriction across a foreign key
+
+A **restriction** on a table is a subset of its rows, written as a condition.
+
+Every foreign key `child → parent` defines a function: each child row references
+exactly one parent row. Propagating a restriction across that edge is a
+**semijoin** (restrict-by-matching), and it works in either direction:
+
+- **downstream** (parent restricted `T'` → child): `S' = S ⋉ T'` — the children
+  whose parent is in `T'`.
+- **upstream** (child restricted `S'` → parent): `T' = T ⋉ S'` — the parents
+  referenced by `S'`.
+
+Downstream and upstream are the *same* operation pointed opposite ways along the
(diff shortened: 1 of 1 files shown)


dimitri-yatsenko:
Closing: the derivation has been relocated to the datajoint-docs specs (folded into diagram.md's 'Traversal algebra' section) — datajoint/datajoint-docs#249 — per the decision that the full derivation and spec live with the other specs, not as a DESIGN-*.md in the code repo. DESIGN-expand-restrict.md was only ever on this branch, so it does not land on master.

